PR #367, opened 2023-12-01
Clean clibboard value during backup flow

zerts:


src/ui/pages/BackupWallet/clipboardWarning.ts
```diff
@@ -6,5 +6,6 @@ export const clipboardWarning = {
       seedType === SeedType.privateKey ? 'private key' : 'recovery phrase';
     return `Did you save the ${secretName}?`;
   },
-  isWarningMessage: (value: string) => value.startsWith('Did you save'),
+  isWarningMessage: (value: string | string[]) =>
+    (Array.isArray(value) ? value.join(' ') : value).startsWith('Did you save'),
```
everdimension, line 10: why not convert to string on call site? Here: https://github.com/zeriontech/zerion-wallet-extension/pull/367/files#diff-60196fc68e991de1816378ec4a5200f343de198e8255663d94a926b3a0e6b8b9R26

Otherwise it creates unnecessary coupling

src/ui/Onboarding/shared/useMnemonicInput.tsx
```diff
@@ -8,11 +8,13 @@ export function useMnemonicInput({
   rows = 4,
   maxInputNumber = 24,
   setValue,
+  showCleanedClipboardMessage,
 }: {
   columns?: number;
   rows?: number;
   maxInputNumber?: number;
   setValue: Dispatch<SetStateAction<string[]>>;
+  showCleanedClipboardMessage?: boolean;
```
everdimension, line 17: maybe simply `showClipboardMessage`?

Or another way:
```
type?: 'text' | 'password'
```

So if type is passed, use it, otherwise follow existing logic for `type`.

everdimension (approved):

PR #368, opened 2023-12-02
Animated border button for ledger signing state

everdimension:


src/ui/ui-kit/Button/Button.tsx
```diff
@@ -14,6 +14,7 @@ type Kind =
   | 'neutral'
   | 'ghost'
   | 'danger'
+  | 'loading-border'
```
zerts, line 17: What do you think about making this as a separate property? So any kind of button will be able to have such loading state?

everdimension: Yeah I thought about this initially, but I don't think it'll work well.
Some buttons don't have a border, so modifying them to have one is as much work as creating a new one

Also some buttons have transparent background, but this one technically requires a background

Also you can always do
```
<Button kind={loading ? 'loading-border' : 'primary'} />
```

PR #370, opened 2023-12-04
Extract large images from the bundle

zerts:


src/ui/popup.html
```diff
@@ -37,6 +37,26 @@
     <link rel="preload" as="image" href="https://protocol-icons.s3.amazonaws.com/uniswap-exchangev3.png">
     <link rel="preload" as="image" href="https://protocol-icons.s3.amazonaws.com/paraswap.png">
 
+    <!-- DNA Mint Flow -->
```
everdimension, line 40: Another way would be to programmatically add these elements from the dna mint module, but as long as we don't forget to remove these later it's ok

everdimension (approved):

PR #372, opened 2023-12-05
Details for timeout errors

everdimension:


src/shared/core/BackgroundScriptUpdateHandler.ts
```diff
@@ -75,7 +75,7 @@ export class BackgroundScriptUpdateHandler {
     const number = getRandomInteger();
     return Promise.race([
       sendPortMessage<{ syn: number }, { ack: number }>(port, { syn: number }),
-      rejectAfterDelay(1000),
+      rejectAfterDelay(1000, 'sendPortMessage'),
```
zerts, line 78: I suggest adding user-readable errors everywhere. At least, error screen will look more friendly:))

everdimension: I think it's quite hard to come up with "user readable" messages for each request. What would you suggest?

zerts: Background port is unavailable

zerts (approved):

src/shared/core/BackgroundScriptUpdateHandler.ts
everdimension: I am not sure this is more user readable 😁

PR #379, opened 2023-12-07
Add analytics for wallet name flag change

everdimension:


src/shared/analytics/analytics.background.ts
```diff
@@ -149,6 +153,35 @@ function trackAppEvents({ account }: { account: Account }) {
     });
     sendToMetabase('add_custom_evm', params);
   });
+
+  emitter.on('globalPreferencesChange', (state, prevState) => {
+    onIdle(() => {
```
zerts, line 158: I would like to see this as a separate named function, so in this common file it is clear what do we want from globalPreferencesChange event. But not critical

PR #381, opened 2023-12-12
Show warning for possibly failing transactions

vyorkin:


src/modules/ethereum/transactions/types.ts
```diff
@@ -12,7 +12,7 @@ export interface TransactionObject {
 
 export type StoredTransactions = Array<TransactionObject>;
 
-interface Warning {
+export interface InterpretWarning {
```
everdimension, line 15: Where is this used?

src/ui/pages/SendTransaction/TransactionWarnings/TransactionMayFailWarning.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { HStack } from 'src/ui/ui-kit/HStack';
+import { VStack } from 'src/ui/ui-kit/VStack';
+import ValidationErrorIcon from 'jsx:src/ui/assets/validation-error.svg';
+import { Spacer } from 'src/ui/ui-kit/Spacer';
+import { UIText } from 'src/ui/ui-kit/UIText';
+
+export function TransactionMayFailWarning() {
+  return (
+    <>
```
everdimension, line 10: This layout needs to be abstracted. It's used in two separate components which are supposed to look the same

everdimension (changes requested):
Where is this used?

src/ui/pages/SendTransaction/TransactionWarnings/TransactionMayFailWarning.tsx
vyorkin: Agree, I requested review too soon, this should be a draft PR actually... I'll extract the common display logic into `TransactionWarning` component

src/modules/ethereum/transactions/types.ts
vyorkin: The idea was to display these warnings in `TransactionWarnings` component (before I realized there is no interpretation warnings yet), I'll revert these changes for now

src/ui/pages/SendTransaction/TransactionWarnings/TransactionWarnings.tsx
```diff
@@ -0,0 +1,40 @@
...
+  addressAction: AddressAction | IncomingAddressAction;
+  chain: Chain;
+  networkFeeConfiguration: NetworkFeeConfiguration;
+}) {
+  return (
+    <>
+      {addressAction.transaction.status === 'failed' ? (
+        <TransactionWarning
+          title="Transaction may fail"
+          message="This transaction can not be broadcasted or it may fail during
```
everdimension, line 28: Why are we not using "message" that comes from interpret response?

everdimension (approved):

PR #386, opened 2023-12-14
Support setting allowance in swap form

vyorkin:
TODO:
- [x] Add "close" button to allowance dialog

Fixes:
- [x] Fix the case when there is no `currentTransaction` available after changing (refetching) allowance
- [x] Fix missing transaction when resetting allowance to unlimited

Refactoring:
- [x] Get rid of extra `RenderArea`

### DIFF
--- a/src/ui/components/address-action/SingleAsset/SingleAsset.tsx
+++ b/src/ui/components/address-action/SingleAsset/SingleAsset.tsx
@@ -16,7 +16,7 @@ import type { Chain } from 'src/modules/networks/Chain';
 import { VStack } from 'src/ui/ui-kit/VStack';
 import { HStack } from 'src/ui/ui-kit/HStack';
 import { isUnlimitedApproval } from 'src/ui/pages/History/isUnlimitedApproval';
-import { TextLink } from 'src/ui/ui-kit/TextLink';
+import { RenderArea } from 'react-area';
 import { AssetQuantity } from '../../AssetQuantity';
 import { AssetLink } from '../../AssetLink';
 
@@ -40,16 +40,12 @@ function FungibleAsset({
   actionType,
   fungible,
   quantity,
-  allowanceViewHref,
-  disabled,
 }: {
   address: string;
   chain: Chain;
   actionType: ActionType;
   fungible: Asset;
   quantity: string;
-  allowanceViewHref?: string;
-  disabled: boolean;
 }) {
   const commonQuantity = useMemo(
     () =>
@@ -99,24 +95,7 @@ function FungibleAsset({
                 )}{' '}
                 <AssetLink asset={fungible} address={address} />
               </UIText>
-              {actionType === 'approve' &&
-              Boolean(quantity) &&
-              allowanceViewHref ? (
-                disabled ? (
-                  <UIText kind="small/accent" color="var(--neutral-500)">
-                    Edit
-                  </UIText>
-                ) : (
-                  <UIText
-                    as={TextLink}
-                    kind="small/accent"
-                    color="var(--primary)"
-                    to={allowanceViewHref}
-                  >
-                    Edit
-                  </UIText>
-                )
-              ) : null}
+              <RenderArea name="single-asset-quantity-right" />
             </HStack>
           }
           detailText={null}
@@ -163,8 +142,6 @@ export function SingleAsset({
   actionType,
   singleAsset,
   allowanceQuantityBase,
-  allowanceViewHref,
-  disabled,
 }: {
   address: string;
   chain: Chain;
@@ -173,8 +150,6 @@ export function SingleAsset({
     NonNullable<AddressAction['content']>['single_asset']
   >;
   allowanceQuantityBase?: string;
-  allowanceViewHref?: string;
-  disabled: boolean;
 }) {
   const fungibleAsset = getFungibleAsset(singleAsset.asset);
   const nftAsset = getNftAsset(singleAsset.asset);
@@ -192,8 +167,6 @@ export function SingleAsset({
         actionType={actionType}
         fungible={fungibleAsset}
         quantity={quantity}
-        allowanceViewHref={allowanceViewHref}
-        disabled={disabled}
       />
     );
   }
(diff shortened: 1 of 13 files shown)


src/ui/components/address-action/SingleAsset/SingleAsset.tsx
```diff
@@ -97,18 +95,7 @@ function FungibleAsset({
...
-                <UIText
-                  as={TextLink}
-                  kind="small/accent"
-                  style={{ color: 'var(--primary)' }}
-                  to={allowanceViewHref}
-                >
-                  Edit
-                </UIText>
-              ) : null}
+              <RenderArea name="single-asset-quantity-right" />
```
vyorkin, line 98: So we don't have to pass the `allowanceViewHref` down to the `SingleAsset` component anymore

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -519,11 +540,63 @@ export function SwapForm() {
...
+                  requestedAllowanceQuantityBase={UNLIMITED_APPROVAL_AMOUNT}
+                  value={value}
+                  onSubmit={(quantity) => {
+                    setAllowanceQuantityBase(quantity.toString());
+                    allowanceDialogRef.current?.close();
+                  }}
+                  footerRenderArea="custom-allowance-footer"
+                />
+                <Spacer height={20} />
+                <RenderArea name="custom-allowance-footer" />
```
vyorkin, line 593: I don't like that I had to use the `RenderArea` here, alternatives that that I can think of:

* Provide smth like `renderFooter` callback and do this:
```tsx
renderFooter={() => (
  <>
    <Spacer height={20} />
    ...
  </>
)}
```
* ...

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -501,6 +519,9 @@ export function SwapForm() {
         height={innerHeight >= 750 ? '70vh' : '90vh'}
         containerStyle={{ display: 'flex', flexDirection: 'column' }}
         renderWhenOpen={() => {
+          if (!currentTransaction || approveTransactionQuery.isLoading) {
```
vyorkin, line 522: Looks like we should attempt to render `TransactionConfirmationView` if there is no `currentTransaction`.
The issue here is that `currentTransaction` won't be available until `approveTransactioinQuery` is loaded

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: Just added an extra `Spacer` to the `CustomAllowanceForm`

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: Setting `keepPreviousData: true` kind of helps but there is still an issue with resetting allowance

src/ui/pages/SendTransaction/SendTransaction.tsx
```diff
@@ -277,8 +278,19 @@ function TransactionDefaultView({
           actionTransfers={actionTransfers}
           singleAsset={singleAsset}
           allowanceQuantityBase={allowanceQuantityBase}
-          allowanceViewHref={allowanceViewHref}
         />
+        {allowanceQuantityBase ? (
```
everdimension, line 282: I think we also used to check that `action.type` is `approve` before showing this

vyorkin: Good point, thanks! Added an extra check for `action.type`

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -519,11 +535,69 @@ export function SwapForm() {
...
               />
             </ViewLoadingSuspense>
           );
         }}
-      ></BottomSheetDialog>
+      />
+
+      <BottomSheetDialog
+        ref={allowanceDialogRef}
+        key="allowance"
```
everdimension, line 549: what for?

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -519,11 +535,69 @@ export function SwapForm() {
...
+        height="min-content"
+        containerStyle={{ display: 'flex', flexDirection: 'column' }}
```
everdimension, line 551: is this necessary for this view?

src/ui/pages/SwapForm/shared/useApproveHandler.ts
```diff
@@ -84,6 +86,7 @@ export function useApproveHandler({
       return { ...tx, gas: gasAsHex, gasLimit: gasAsHex };
     },
     staleTime: Infinity,
+    keepPreviousData: true,
```
everdimension, line 89: This is a bit controversial btw, I think it makes it possible to submit stale transaction

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: I don't think it is necessary here, added for consistency with the `BottomSheetDialog` above, I'll delete this key

src/ui/pages/SwapForm/shared/useApproveHandler.ts
vyorkin: I'm thinking about blocking the Send (Approve) button until the new/updated `approveTransaction` is constructed.
I mean literally: `<SignTransactionButton ... disabled={... || allowanceIsLoading}`

vyorkin: ✅

src/ui/pages/SendTransaction/SendTransaction.tsx
vyorkin: ✅

everdimension (approved):

src/ui/components/address-action/SingleAsset/SingleAsset.tsx
```diff
@@ -95,7 +96,7 @@ function FungibleAsset({
                 )}{' '}
                 <AssetLink asset={fungible} address={address} />
               </UIText>
-              <RenderArea name="single-asset-quantity-right" />
+              {elementEnd}
```
vyorkin, line 99: this looks better indeed

PR #388, opened 2023-12-17
Add SpeedUp transaction flow

everdimension:


src/background/transactions/TransactionService.ts
```diff
@@ -52,22 +53,38 @@ export class TransactionService {
...
+          hash: transaction.hash,
+          initiator,
+          timestamp: Date.now(),
+        };
+        if (
+          addressAction &&
+          isLocalAddressAction(addressAction) &&
+          addressAction.relatedTransaction
+        ) {
+          newItem.relatedTransactionHash = addressAction.relatedTransaction;
```
zerts, line 77: A bit confusing that transactionHash = transaction

everdimension: Since we're augmenting the backend entity `AddressAction`, I tried to come up with a name that they're more likely to introduce on the backend if they ever do it. But locally I wanted to have a less confusing name.

PR #395, opened 2023-12-19
Register chains for addresses in zpi

zerts:


src/modules/networks/getBackendNetwork.ts
```diff
@@ -0,0 +1,44 @@
+import { ethers } from 'ethers';
+import { rejectAfterDelay } from 'src/shared/rejectAfterDelay';
+import { getNetworksBySearch } from '../ethereum/chains/requests';
+import { networksStore } from './networks-store.background';
+import type { Chain } from './Chain';
+import type { NetworkConfig } from './NetworkConfig';
+
+function maybeLocalChainId(id: string) {
+  return id.length === 21; // nanoid() standart length
```
vyorkin, line 9: Not sure if it is safe to rely on the id length... Are there any other ways to check for local chain id/network?

zerts: Yes :(

src/background/initialize.ts
```diff
@@ -36,6 +37,7 @@ export async function initialize() {
     globalPreferences.initialize();
   });
   initializeAnalytics({ account });
+  initializeRegisterChain(account.getCurrentWallet());
```
everdimension, line 40: instance of wallet can change during app lifecycle

src/background/requests/registerChain.ts
```diff
@@ -0,0 +1,41 @@
...
+  getBackendNetworkByChainId,
+  getBackendNetworkByLocalChain,
+} from 'src/modules/networks/getBackendNetwork';
+import { normalizeAddress } from 'src/shared/normalizeAddress';
+import { emitter } from '../events';
+import type { Account } from '../account/Account';
+
+export function initialize(account: Account) {
+  function listenForChainChange() {
+    account.getCurrentWallet().emitter.on('chainChanged', async (chain) => {
```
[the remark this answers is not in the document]
zerts: Hm. But it doesn't have chain data. Anyway, I see your point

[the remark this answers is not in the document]
zerts: Will listen to it instead

src/modules/zerion-api/requests/register-wallets.ts
```diff
@@ -0,0 +1,8 @@
+export interface Payload {
+  addressses: string[];
```
everdimension, line 2: is it three `sss` in this one?)

src/background/requests/registerAddresses.ts
```diff
@@ -0,0 +1,13 @@
+import { ZerionAPI } from 'src/modules/zerion-api/zerion-api';
+import { WalletOrigin } from 'src/shared/WalletOrigin';
+import { emitter } from '../events';
+
+export function initialize() {
+  emitter.on('walletCreated', async ({ walletContainer, origin }) => {
+    if (origin === WalletOrigin.imported) {
+      ZerionAPI.registerAddresses({
+        addressses: walletContainer.wallets.map((wallet) => wallet.address),
```
everdimension, line 9: I assume it might be `addresses`

src/modules/networks/getBackendNetwork.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+async function getChainId(network?: NetworkConfig) {
```
everdimension, line 12: This returns a `NetworkConfig`, not `chainId`

src/modules/networks/getBackendNetwork.ts
everdimension, line 9: "standard"

src/background/initialize.ts
```diff
@@ -38,6 +40,8 @@ export async function initialize() {
     setUninstallURL();
   });
   initializeAnalytics({ account });
+  initializeRegisterChain(account);
+  initializeRegisterAddresses();
```
everdimension, line 44: Do these handlers need separate initializations?

Btw their functionality looks a lot like analytics. Can they be added to the current `analytics.background.ts` file?

src/background/requests/registerAddresses.ts
```diff
@@ -0,0 +1,13 @@
+import { ZerionAPI } from 'src/modules/zerion-api/zerion-api';
+import { WalletOrigin } from 'src/shared/WalletOrigin';
+import { emitter } from '../events';
+
+export function initialize() {
+  emitter.on('walletCreated', async ({ walletContainer, origin }) => {
```
everdimension, line 6: I'd add a comment here why backend needs this event

src/background/requests/registerChain.ts
everdimension, line 12: I'd add a comment here why backend needs this event

src/background/requests/registerChain.ts
```diff
@@ -0,0 +1,41 @@
...
+        });
+      }
+    });
+  }
+  listenForChainChange();
+  // Update subscription after Wallet's recreation inside Account
+  account.on('reset', () => {
+    listenForChainChange();
+  });
+
```
everdimension, line 28: Currently events aren't cleaned up

This re-application of listeners might not be needed if we use the global emitter

src/background/initialize.ts
zerts: I don't want to mix analytics and lite-chain support. However, it is ok to combine them together

src/background/requests/registerChain.ts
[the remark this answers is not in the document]
everdimension: I think we can pass the chain to it since it's already being broadcasted

everdimension (changes requested):
some comments

src/background/requests/registerChain.ts
zerts: Yes. I didn't clean them because this reset event means that emitter was destroyed anyway.
But I will reuse global emitter anyway

everdimension (approved):

PR #396, opened 2023-12-20
Fix incorrect decimals type in custom networks

zerts:


src/modules/networks/networks-store.ts
```diff
@@ -29,6 +30,24 @@ export class NetworksStore extends Store<State> {
     this.getEthereumChainSources = getEthereumChainSources ?? null;
   }
 
+  private ensureNetworkConfigTypes(
```
everdimension, line 33: Should come up with a better name, this is more like "prepare"

PR #398, opened 2023-12-27
Add connect wallet params to open in web nft link

zerts:
Won't fix.We will add send nft form instead


zerts:
No need in this update

PR #412, opened 2024-01-25
Fix session expire logic for onboarding and backup wallet flows

zerts:


src/background/index.ts
```diff
@@ -203,9 +204,6 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
       throw new Error('popupUrl not found');
     }
     const url = new URL(browser.runtime.getURL(popupUrl));
-    url.searchParams.append('templateType', 'tab');
-    browser.tabs.create({
-      url: url.toString(),
-    });
+    openOnboarding(url);
```
zerts, line 207: This was the cause of the main problem. We didn't add context param here, so session was erased on page reload on initial onboarding open

everdimension (approved):

PR #414, opened 2024-01-26
Add editable autolock timer

zerts:


src/background/Wallet/GlobalPreferences.ts
```diff
@@ -23,6 +25,7 @@ export interface State {
   recognizableConnectButtons?: boolean;
   providerInjection?: ProviderInjection;
   walletNameFlags?: Record<string, WalletNameFlag[]>;
+  autoLockTimer?: number;
```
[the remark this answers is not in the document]
zerts: Ok, agreed. Probably if it is equal to 0, then it is "None" or something else from the other future field

src/background/user-activity.ts
```diff
@@ -17,7 +18,7 @@ export function scheduleAlarms() {
     // To my understanding, if an alarm with an existing name is created, it's
     // not gonna create duplicate alarms. Only one will be active and that's what we need
     browser.alarms.create('lastActiveCheck', {
-      periodInMinutes: 5, // Is this too frequent or too infrequent?
+      periodInMinutes: 1,
```
everdimension, line 21: 🤔 Why are we changing this?

src/background/user-activity.ts
```diff
@@ -27,12 +28,16 @@ export function scheduleAlarms() {
...
+}
+
 export async function handleAlarm(alarm: browser.Alarms.Alarm) {
   if (alarm.name === 'lastActiveCheck') {
     const lastActive = await getLastActive();
-    // const ONE_DAY = 1000 * 60 * 60 * 24;
-    const HALF_A_DAY = 1000 * 60 * 60 * 12;
-    if (lastActive && Date.now() - lastActive > HALF_A_DAY) {
+    const idleTimeout = await getIdleTimeout();
+    if (idleTimeout && lastActive && Date.now() - lastActive > idleTimeout) {
```
everdimension, line 40: can `idleTimeout` be falsy? If not, I'd compare explicitly to zero
If yes, then I'd add a fallback to logout after some predefined time

src/background/user-activity.ts
```diff
@@ -27,12 +28,16 @@ export function scheduleAlarms() {
...
+  return preferences.autoLockTimeout;
+}
+
 export async function handleAlarm(alarm: browser.Alarms.Alarm) {
   if (alarm.name === 'lastActiveCheck') {
     const lastActive = await getLastActive();
-    // const ONE_DAY = 1000 * 60 * 60 * 24;
-    const HALF_A_DAY = 1000 * 60 * 60 * 12;
-    if (lastActive && Date.now() - lastActive > HALF_A_DAY) {
+    const idleTimeout = await getIdleTimeout();
```
everdimension, line 39: why two names? `autoLockTimeout` and `idleTimeout`? Maybe use `autoLockTimeout` here, too?

src/background/user-activity.ts
zerts: Because we have a 1minute option now. So alarm every 5 mins will not catch this case

src/background/user-activity.ts
zerts: I implemented 0 as 'none'.
We can change this and make this option more obvious

PR #417, opened 2024-02-08
Fix native position filter [WLT-4174]

vyorkin:


src/ui/shared/requests/useNativeBalance.ts
```diff
@@ -29,10 +31,18 @@ function useNativeAddressPosition({
   );
 
   return useMemo(() => {
+    const nativePositions = value?.positions?.filter(
+      (item) =>
+        item.chain === chain.toString() &&
+        item.type === 'asset' &&
+        !item.protocol
```
zerts, line 38: As I understand, if item.type === asset, it can't be inside protocol. So this filter can be simplified to 
` (item) => item.chain === chain.toString() && item.type === 'asset'`

zerts: Oh I see. Do we really get position with type: asset and non-empty protocol?

vyorkin: Yeah, this was exactly the reason why I had to add this condition

zerts: Wow. Just faced this. Approve then!

src/ui/shared/requests/useNativeBalance.ts
```diff
@@ -29,10 +31,18 @@ function useNativeAddressPosition({
...
+    );
+    invariant(
+      !nativePositions || nativePositions.length <= 1,
+      'multiple native positions'
+    );
```
everdimension, line 43: Do we really want to crash here?

everdimension: Also this condition is quite hard to read :)

everdimension (changes requested):
https://github.com/zeriontech/zerion-wallet-extension/pull/417/files#r1497582516

src/ui/shared/requests/useNativeBalance.ts
vyorkin: The idea is to make it easier to debug in case anything similar happens again. So we'll have a clear message indicating that we received multiple native positions from the backend api

vyorkin: I simplified the invariant condition ✅

src/ui/shared/requests/useNativeBalance.ts
```diff
@@ -29,10 +31,16 @@ function useNativeAddressPosition({
...
 
   return useMemo(() => {
+    const nativePositions =
+      value?.positions?.filter(
+        (item) =>
+          item.chain === chain.toString() &&
+          item.type === 'asset' &&
+          !item.protocol
+      ) ?? [];
+    invariant(nativePositions.length <= 1, 'multiple native positions');
```
everdimension, line 41: Again, I don't understand why we want to crash here. Let's assume backend sends an extra position by mistake. Does this prevent us from signing a transaction?

vyorkin: I see, you're right. Replaced invariant error with console.warn

everdimension (approved):

PR #419, opened 2024-02-19
Add new links to settings and better connection icon

vyorkin:


src/ui/pages/Settings/Settings.tsx
```diff
@@ -126,6 +129,63 @@ function SettingsMain() {
             },
             {
               key: 8,
+              pad: false,
+              style: { paddingInline: 0 },
+              component: (
+                <SurfaceItemAnchor
```
zerts, line 135: This line should be shown only to non-premium wallets. However, it is not obvious how to check premium status in global menu related to all addresses.

everdimension:
Let's make this one a draft until it's approved by product

PR #425, opened 2024-02-25
Fix: inject in-page script in iframe documents

everdimension:


src/modules/ethereum/provider.ts
```diff
@@ -35,7 +35,8 @@ async function fetchInitialState(connection: Connection) {
 
 class MetamaskExperimentalNamespace {
   isUnlocked() {
-    throw new MethodNotImplemented('_metamask.isUnlocked: Not implemented');
+    return true;
```
everdimension, line 38: What do you think? We don't want to crash when some dapps access this field, but I don't think it matters if we track the actual "unlocked" state, since we have a flow of forwarding to the destination route after login anyway.

zerts (approved):

PR #426, opened 2024-02-26
Pass platform and version to socket api

zerts:


src/shared/packageVersion.ts
```diff
@@ -3,3 +3,4 @@ import { version as packageVersion } from '../../package.json';
 const isProd = process.env.NODE_ENV === 'production';
 
 export const version = isProd ? packageVersion : `${packageVersion}-dev`;
+export const platform = 'web-extension';
```
zerts, line 6: Is the a good place for this constant?

everdimension: Good question. I would put it separately from current file, since it's more bundle-specific.
Maybe somewhere inside `/src/shared/analytics` is a good location.

zerts: Not sure about analytics, but I don't have a better option, so let's move it to `shared/analytics`

PR #430, opened 2024-02-29
Fix: account for maxFee in Estimated Fee calculation

everdimension:


src/ui/pages/SendTransaction/NetworkFee/NetworkFeeDialog.tsx
```diff
@@ -178,8 +178,8 @@ function CustomNetworkFeeForm({
     }
 
     return {
-      priorityFeeFiat: getFiatValue(
-        (priorityFee || 0) + (eip1559?.base_fee || 0)
+      expectedFeeFiat: getFiatValue(
+        Math.min((priorityFee || 0) + (eip1559?.base_fee || 0), maxFee)
```
zerts, line 182: This can be defined as variable. Cause as I can see, we will use the same formula below.

everdimension: But the values are in different scopes?

PR #431, opened 2024-03-01
Prefer overflowWrap: word-break for better word breaking

everdimension:


vyorkin:
Ah, I see:
> In contrast to [word-break](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), overflow-wrap will only create a break if an entire word cannot be placed on its own line without overflowing

Nice

vyorkin (approved):

PR #432, opened 2024-03-04
Update Settings UI and networks UI

zerts:
Connected pull requests (with type updates):
- https://github.com/zeriontech/defi-sdk-js/pull/58
- https://github.com/zeriontech/transactions-ui/pull/6


src/background/Wallet/GlobalPreferences.ts
```diff
@@ -25,6 +25,7 @@ export interface State {
   recognizableConnectButtons?: boolean;
   providerInjection?: ProviderInjection;
   autoLockTimeout?: number | 'none';
+  enableTestnets?: boolean;
```
everdimension, line 28: why not part of user preferences?

src/modules/networks/networks-store.ts
```diff
@@ -1,101 +1,171 @@
...
-  }, 15000);
+  async loadNetworksWithChainId(chainId: number) {
+    const key = `chainId-${chainId}`;
+    if (!this.loaderPromises[key]) {
+      this.loaderPromises[key] = this.fetchNetworkById(chainId);
+    }
+    return this.loaderPromises[key].then((networks) => {
+      delete this.loaderPromises[key];
+      return networks;
+    });
```
everdimension, line 169: look like we're creating a new promise on each call of this method, and doing the same thing inside

src/modules/networks/networks-store.ts
```diff
@@ -1,101 +1,171 @@
...
+
+    const savedChainConfigs = await this.getEthereumChainConfigs?.();
+    const savedIds = savedChainConfigs?.ethereumChainConfigs?.map(
+      (config) => config.id
+    );
+    const chainsToFetch = Array.from(
+      new Set([...(savedIds || []), ...(chains || [])])
+    );
+
+    const [extraNetworkConfigs, commonNetworkConfis] = await Promise.allSettled(
```
everdimension, line 90:
```suggestion
    const [extraNetworkConfigs, commonNetworkConfigs] = await Promise.allSettled(
```

src/ui/ui-kit/SurfaceList/VirtualizedSurfaceList/VirtualizedSurfaceList.tsx
```diff
@@ -1,34 +1,93 @@
...
   estimateSize: (index: number) => number;
+  style?: React.CSSProperties;
+  context?: 'window' | 'dialog';
+  stickyFirstElement?: boolean;
 }) {
-  const rowVirtualizer = useWindowVirtualizer({
+  const listRef = React.useRef<HTMLDivElement | null>(null);
+  const { uiScrollRootElement } = useContext(UIContext);
+
+  const windowVirtualizer = useWindowVirtualizer({
```
zerts, line 30: Should there be 2 different components?

src/modules/networks/networks-store.ts
zerts: Hm. Do we? I think, it is created only if there is no current promise. Otherwise, we are setting another .then

src/modules/networks/NetworkConfig.ts
```diff
@@ -1,32 +1,40 @@
...
+  standard: 'eip155';
+  specification: {
+    eip155: {
+      eip1559: boolean;
+      id: number;
+    };
+  };
+};
+
+export type NetworkConfig = NetworkSpecification & {
```
zerts, line 19: Should we just import this type from @transactions-ui?

src/modules/networks/networks-store.ts
everdimension: Yes.
Imagine 3 calls to `loadNetworksWithChainId()`. Each call will attach a new 
```
.then(() => /* delete this.loaderPromises[key]; */)
```

So when `this.fetchNetworkById(chainId);` resolves, `delete` will be called 3 times.

Also, if `loadNetworksWithChainId()` is called again after previous calls are resolved, it will call `this.fetchNetworkById(chainId);`. Maybe this part is by design.

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -1,18 +1,36 @@
...
+export function maybeLocalChainId(id: string) {
+  return id.length === 21; // nanoid() standard length
+}
 
 export interface EthereumChainConfig extends NetworkConfigMetaData {
+  value: AddEthereumChainParameter;
+  id: string;
+}
+
+interface EtherenumNetworkConfig extends NetworkConfigMetaData {
```
everdimension, line 25:
```suggestion
interface EthereumNetworkConfig extends NetworkConfigMetaData {
```

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -1,18 +1,36 @@
...
   value: NetworkConfig;
 }
 
 export interface ChainConfig {
-  ethereumChains: EthereumChainConfig[];
+  ethereumChainConfigs?: EthereumChainConfig[];
+  migratedToV1_4?: boolean;
```
everdimension, line 31: What does this name mean? Why do we need such a complex flag?

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
   private collection: { [key: string]: NetworkConfig | undefined };
   private collectionByEvmId: { [key: string]: NetworkConfig | undefined };
-  private originalCollection: { [key: string]: NetworkConfig | undefined };
-  private supportedByBackend: Set<string>;
+  private ethereumChainConfigs: EthereumChainConfig[];
   static purposeKeyMap = {
     sending: 'supports_sending',
     trading: 'supports_trading',
-    bridge: 'supports_bridge',
+    bridge: 'supports_bridging',
```
everdimension, line 54:
```suggestion
    bridging: 'supports_bridging',
```

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
+    actions: 'supports_actions',
+    nfts: 'supports_nft_positions',
```
everdimension, line 56:
```suggestion
    nftPositions: 'supports_nft_positions',
```

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
+  networkConfigs: NetworkConfig[],
+  chainConfigs: EthereumChainConfig[]
+): NetworkConfig[] {
+  const chainConfigById = Object.fromEntries(
+    chainConfigs.map((config) => [config.id, config.value])
+  );
+  return networkConfigs.map((network) =>
+    mergeChainAndNetworkConfigs(network, chainConfigById[network.id])
+  );
 }
```
everdimension, line 38: Judging by implementation, this function does not _insert_ new values to networkConfigs, so the name is misleading
Instead it seems to update some of the `networkConfigs` with values from `chainConfigs`. Values not found in `networkConfigs` are ignored.

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
       this.networks,
-      (network) => network.chain,
+      (network) => network.id,
       (x) => x
     );
-    const collectionByEvmId = toCollection(
+    this.collectionByEvmId = toCollection(
       this.networks,
-      (x) => x.external_id,
+      (x) => getChainId(x) || '',
```
everdimension, line 76: It's weird to have empty string as a key.
What do we do about such networks? I would suggest filtering them out.

In turn, `getChainId(x): string` should maybe even throw error when it cannot find a chainId?
And we can later introduce a `peekChainId(x): string | null` if we need to deliberately look for a network that has a chainId missing?

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
-      collection,
-      collectionByEvmId,
-      originalCollection,
-      sourcesNormalized,
-    } = this.prepare();
-    this.collection = collection;
-    this.originalCollection = originalCollection;
-    this.collectionByEvmId = collectionByEvmId;
-    this.sourcesNormalized = sourcesNormalized;
+    ) as { [key: number]: NetworkConfig | undefined };
```
everdimension, line 78: why is `key: number`? `getChainId` seems to return a string, why do we have this casting?

src/modules/networks/Networks.ts
```diff
@@ -25,152 +25,75 @@ function toCollection<T, K>(
...
   private toId(chain: Chain) {
     const item = this.collection[chain.toString()];
     if (!item) {
       throw new Error(`Chain not found: ${chain}`);
     }
-    return item.external_id as ChainId;
+    return getChainId(item);
+  }
+
+  isSavedChain(chain: Chain) {
```
everdimension, line 93: Let's discuss how to name this one better

src/modules/networks/Networks.ts
```diff
@@ -364,13 +237,13 @@ export class Networks {
       throw new UnsupportedNetwork(`Unsupported network id: ${chainId}`);
     }
     return {
-      chainId,
+      chainId: valueToHex(chainId),
```
everdimension, line 240: isn't `chainId` parameter already a string?

src/modules/networks/Networks.ts
```diff
@@ -379,7 +252,10 @@ export class Networks {
     };
   }
 
-  supports(purpose: TransactionPurpose, chain: Chain): boolean {
+  supports(
+    purpose: keyof (typeof Networks)['purposeKeyMap'],
```
everdimension, line 256: I'd extract (in same file)
```
type TransactionPurpose = keyof (typeof Networks)['purposeKeyMap']
```

src/modules/ethereum/chains/ChainConfigStore.ts
zerts: This is a dev name, I will change it. Just changed it for trying migration multiple times

src/modules/networks/networks-store.ts
zerts: The second part is by design - it will stop inside the fetch function if data is exists.
So here is the same logic like in ready() for stores.

zerts: I will think about this today. It definitely can be done better

src/modules/networks/Networks.ts
zerts: Hm, why not. It is inserting chainConfigs into networkConfigs when possible.

src/modules/networks/Networks.ts
zerts: Had another idea 😏

src/modules/networks/Networks.ts
zerts: Yes! I will do it more reliable

src/modules/networks/Networks.ts
```diff
@@ -354,7 +224,10 @@ export class Networks {
     return this.getExplorerTokenUrl(this.collection[chain.toString()], address);
   }
 
-  getExplorerNameById(chainId: string) {
+  getExplorerNameById(chainId: string | null) {
+    if (!chainId) {
+      return undefined;
+    }
```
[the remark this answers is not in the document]
zerts: I will fix this

src/modules/networks/Networks.ts
zerts: Sure!

src/modules/networks/Networks.ts
zerts: Good catch!

src/modules/networks/networks-store.ts
zerts: Oh, I remember the idea. This promises should help us to avoid parallel requests. However, it is important that all future requests will get the most up-to-date networks object. So I clear these promises after the successful call. What I should do is to clear them after errors as well

src/ui/pages/Networks/shared/useWalletAddresses.ts
```diff
@@ -0,0 +1,26 @@
+import { useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
+import { walletPort } from 'src/ui/shared/channels';
+
+export function useWalletAddresses() {
```
zerts, line 5: Looks like we can use this hook in the wallet list in the future

src/ui/pages/Networks/shared/NetworkList/NetworkList.tsx
```diff
@@ -104,97 +117,124 @@ export function NetworkList({
...
+              )}
+            </HStack>
+          </HStack>
+        </SurfaceItemLink>
+      ),
+    })),
+  ].filter(isTruthy);
+
+  return items.length > 50 ? (
+    <VirtualizedSurfaceList
```
zerts, line 221: Should we just show virtualized list?

src/ui/pages/Networks/NetworkSelect/NetworkSelect.tsx
```diff
@@ -17,19 +16,30 @@ import { NetworkIcon } from 'src/ui/components/NetworkIcon';
...
   onChange: (value: string) => void;
-  renderButton?(params: { value: string; openDialog(): void }): React.ReactNode;
+  renderButton?(params: {
+    value: string;
+    openDialog(): void;
+    networks: Networks | null;
+    networksAreLoading: boolean;
+  }): React.ReactNode;
   dialogRootNode?: HTMLElement;
-  groups?: NetworkGroups;
```
zerts, line 32: For now there is no case when predefined groups are being used

src/modules/networks/Networks.ts
zerts: Ok, I renamed this :)

src/modules/networks/Networks.ts
zerts: Updated it, but still not so sure

src/ui/pages/Networks/Networks.tsx
```diff
@@ -429,97 +371,131 @@ function NetworksView({
...
-    locationStore.setState({
-      itemCreateSuccess: createChain(network.chain),
-    });
+  const chains = useMemo(
+    () => Object.keys(portfolioDecomposition?.chains || {}),
+    [portfolioDecomposition]
+  );
+  const { networks, isLoading } = useNetworks(chains);
+
+  useEffect(() => {
```
zerts, line 477: Should we define this as helper?

src/modules/networks/networks-store.ts
everdimension: Let's discuss this

src/ui/ui-kit/SurfaceList/VirtualizedSurfaceList/VirtualizedSurfaceList.tsx
everdimension: wdym?

src/modules/networks/helpers.ts
```diff
@@ -1,20 +1,29 @@
...
 import type { AddEthereumChainParameter } from '../ethereum/types/AddEthereumChainParameter';
+import { getCustomNetworkId } from '../ethereum/chains/helpers';
 import type { NetworkConfig } from './NetworkConfig';
-import type { Chain } from './Chain';
+
+export function getChainId(network: NetworkConfig) {
+  return network.standard === 'eip155'
+    ? normalizeChainId(network.specification.eip155.id)
+    : null;
+}
```
everdimension, line 11: What do you think about making this a static method of `Networks`?

src/modules/ethereum/chains/helpers.ts
```diff
@@ -0,0 +1,9 @@
+import { normalizeChainId } from 'src/shared/normalizeChainId';
+
+export function getCustomNetworkId(chainId: string) {
```
everdimension, line 3: maybe `toCustomNetworkId`?

src/modules/ethereum/chains/helpers.ts
```diff
@@ -0,0 +1,9 @@
...
+  return `zerion-extension-${normalizeChainId(chainId)}`;
```
everdimension, line 4: maybe something like `client-id-`? This way it could be theoretically reusable between clients :)

src/modules/ethereum/chains/requests.ts
```diff
@@ -22,46 +21,22 @@ function fetchChains(
...
         payload,
       },
-      onData: ({ value }) => {
+      onData: ({ value, isStale }) => {
         if (value) {
           resolve(value);
-          unsubscribe();
+          if (!isStale) {
+            unsubscribe();
+          }
```
everdimension, line 29: Can you explain what's going on here? :)

src/modules/ethereum/chains/requests.ts
```diff
@@ -22,46 +21,22 @@ function fetchChains(
...
   });
 }
 
 export async function getNetworksBySearch({ query }: { query: string }) {
-  const networks = await fetchChains({
+  return fetchChains({
     include_testnets: true,
     supported_only: false,
     search_query: query.trim().toLowerCase(),
   });
```
everdimension, line 41: Let's add reject after delay here, too. It seems it's missing

src/modules/ethereum/chains/requests.ts
zerts: It is a way to avoid calling unsubscribe before initialisation.
As I understand, it can happen because of instant onData call with stale data. So I tried to avoid this case

src/modules/ethereum/chains/helpers.ts
zerts: This is valid. However, I would probably mention Zerion in the template to decrease probability of collisions with real network names

src/modules/networks/helpers.ts
zerts: Ok, there is one moment I can't solve for now. 
There will be 2 `getChainId` methods inside Networks.
One is static accepting NetworksConfig, and one is not static, accepting Chain.
Is it ok to keep 2 methods with the same name?

src/modules/ethereum/chains/requests.ts
zerts: Agree!

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -23,35 +39,40 @@ function remove<T>(arr: T[], predicate: (item: T) => boolean) {
 }
 
 export class ChainConfigStore extends PersistentStore<ChainConfig> {
```
everdimension, line 41: I think `ChainConfigStore` class isn't used anywhere outside? Let's remove this export

src/ui/pages/Networks/shared/NetworkList/NetworkList.tsx
```diff
@@ -31,37 +31,39 @@ function getOriginUrlFromMetaData(metadata: NetworkConfigMetaData) {
...
 }
 
 function NetworkDetail({
   metadataRecord,
   network,
-  networks,
 }: {
   metadataRecord: Record<string, NetworkConfigMetaData | undefined>;
   network: NetworkConfig;
   networks: Networks;
```
everdimension, line 51: not used

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -61,10 +82,117 @@ export class ChainConfigStore extends PersistentStore<ChainConfig> {
...
+      ethereumChainConfigs = [],
+      ethereumChains,
+    } = this.getState();
+    if (migratedChainConfigs || !ethereumChains) {
+      return;
+    }
+    const existingIdSet = new Set(ethereumChainConfigs?.map(({ id }) => id));
+    for (const { value, ...config } of ethereumChains) {
+      const chainConfig = toAddEthereumChainParamer(value);
+      let id = !maybeLocalChainId(value.chain) ? value.chain : null;
```
everdimension, line 108:
```suggestion
      let id = maybeLocalChainId(value.chain) ? null : value.chain;
```

src/background/Wallet/Wallet.ts
```diff
@@ -203,6 +204,8 @@ export class Wallet {
     this.record = await this.walletStore.read(this.id, this.userCredentials);
     if (this.record) {
       this.emitter.emit('recordUpdated');
+      chainConfigStore.cleanupDeprecatedCustomChainPermissions(this);
+      chainConfigStore.checkChainsForUpdates(this);
```
everdimension, line 208: This is too much of a coupling. Why call two methods if we only ever call them together?

src/modules/networks/networks-store.ts
```diff
@@ -1,101 +1,170 @@
...
-            chainConfig.value.native_asset.decimals
-          );
-        }
-      });
-    });
+  private getKnownIdsSet() {
+    return new Set([
+      ...this.networkConfigs.map((config) => config.id),
+      ...this.customNetworkConfigs.map((config) => config.id),
+    ]);
```
everdimension, line 51: Isn't `this.state.networks.collection` already a dictionary with all known ids?

src/modules/networks/networks-store.ts
```diff
@@ -1,101 +1,170 @@
...
-      this.loaderPromise = this.fetchAndUpdate();
+  async pushConfigs(...extraNetworkConfigs: NetworkConfig[]) {
+    this.networkConfigs = mergeNetworkConfigs(
+      this.networkConfigs,
+      extraNetworkConfigs
+    );
+    return this.updateNetworks();
+  }
+
+  async load(chains?: string[]) {
```
everdimension, line 138: I don't like that we create lots of promise objects even if all chains are known already

src/background/Wallet/Wallet.ts
```diff
@@ -810,8 +813,9 @@ export class Wallet {
     spender: string;
   }>) {
     this.verifyInternalOrigin(context);
-    const networks = await networksStore.load();
+    const networks = await networksStore.load([chain]);
```
everdimension, line 816: I don't understand why we pass `[chain]` in these method. Isn't the network store expected to have them?

src/ui/pages/Networks/Networks.tsx
```diff
@@ -38,119 +42,80 @@ import { UnstyledLink } from 'src/ui/ui-kit/UnstyledLink';
...
+  chainConfig: AddEthereumChainParameter;
+  prevChain?: string;
+}) {
+  const networks = await networksStore.load(
+    [chain, prevChain].filter(isTruthy)
   );
+  const chainsMetadata = networks.getNetworksMetaData();
+  const metadata = chainsMetadata[prevChain || chain];
+  if (prevChain && prevChain !== chain) {
+    await walletPort.request('updateChainForAffectedOrigins', {
```
everdimension, line 85: I don't think this method should be called from the ui side. It's an internal accidental cleanup

src/modules/networks/networks-store.ts
zerts: It doesn't have it on initial fetchNetworks. I need to think, is it possible to get rid of this helper.

zerts: Fair

src/background/Wallet/Wallet.ts
zerts: This is one of the key points of the new system. If you need to work with a chain, it is better to load networks with this chain as a param.
If this chain is there, getter will not load anything, but in other cases you just don't need to think if it is here or not. This is especially important because we have 2 separate networks stores for UI and background. That's why event is UI knows about the networks, this can be the first time background gets it.

src/modules/networks/networks-store.ts
zerts: Let me remove this for real and check it properly again then :)

src/background/Wallet/Wallet.ts
```diff
@@ -1133,59 +1140,82 @@ export class Wallet {
...
     this.verifyInternalOrigin(context);
     this.ensureRecord(this.record);
-    const chain = createChain(chainStr);
-    chainConfigStore.removeEthereumChain(chain);
-    // We need to update all permissions which point to the removed chain
+    chainConfigStore.removeEthereumChain(createChain(chainStr));
+    this.verifyOverviewChain();
+  }
+
+  async updateChainForAffectedOrigins({
```
everdimension: This doesn't need to be async
(But keep in mind that making it sync will make it unavailable to outside scripts, so see https://github.com/zeriontech/zerion-wallet-extension/pull/432/files#r1586302220

everdimension: bump

src/modules/networks/networks-store.ts
everdimension: See NetworksStore `fetchNetworkById`

src/background/Wallet/Wallet.ts
everdimension: We decided to not call `chainConfigStore` here at all

src/ui/pages/Networks/shared/useWalletAddresses.ts
```diff
@@ -0,0 +1,26 @@
...
+  const addresses = useMemo(() => {
+    if (!walletGroups) {
+      return null;
+    }
+    const result: string[] = [];
+    walletGroups.forEach((group) =>
+      group.walletContainer.wallets.forEach((wallet) =>
+        result.push(wallet.address)
+      )
+    );
```
everdimension, line 21: for of?

src/ui/ui-kit/SurfaceList/VirtualizedSurfaceList/VirtualizedSurfaceList.tsx
everdimension: Yes, let's make 2 components :)

everdimension:
- [ ] When backend network is edited, set `origin` to something like `backend` or `chains info`
    That way we will be able to always distinct configs created manually by seeing `origin: 'chrome://...`

src/background/Wallet/model/versions.ts
```diff
@@ -7,7 +8,7 @@ import type {
   WalletRecordVersion4,
 } from './types';
 
-type PossibleEntry =
+export type PossibleEntry =
```
everdimension, line 11: unnecessary export?

src/shared/type-utils/versions.ts
```diff
@@ -0,0 +1,66 @@
+type VertionableEntity = { version: number };
```
everdimension, line 1:
```suggestion
type VersionableEntity = { version: number };
```

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -22,36 +17,65 @@ function remove<T>(arr: T[], predicate: (item: T) => boolean) {
...
+  async restore() {
+    const saved = await PersistentStore.readSavedState<ChainConfig>(this.key);
+    if (saved) {
+      this.setState(upgradeRecord(saved, upgrades));
+    }
+    this.isReady = true;
+    this.checkChainsForUpdates();
+  }
+
+  readyPromise = this.restore();
```
everdimension, line 35: Why overwrite this? The same code is being called in `PersistentStore`'s constructor

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -61,10 +85,45 @@ export class ChainConfigStore extends PersistentStore<ChainConfig> {
...
+      this.setState((current) => ({
+        ...current,
+        ethereumChainConfigs: updatedEthereumChainConfigs,
+      }));
+    }
+  }
+
+  async ready(): Promise<void> {
+    return this.isReady ? Promise.resolve() : this.readyPromise;
+  }
```
everdimension, line 126: Why overwrite this? Persistent store has the exact same lines

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -38,27 +19,49 @@ function remove<T>(arr: T[], predicate: (item: T) => boolean) {
...
     const state = this.getState();
     const existingItems = new Map(
       state.ethereumChainConfigs?.map((config) => [config.id, config])
     );
-    const existingEntry = existingItems.get(id);
+    const existingEntry = existingItems.get(prevId);
+    const existingPreviousIds = existingEntry?.previousIds || [];
+    const previousIds =
+      prevId !== id && !existingPreviousIds.includes(prevId)
+        ? [...existingPreviousIds, prevId]
```
everdimension, line 55: Why do we need to save this?

zerts: Cause id can be changed for custom networks. And I want to save all history in this case.
However, these aliases will be less in priority, so should work almost fine

src/modules/ethereum/chains/ChainConfigStore.ts
zerts: I need to check how classes in js works. In some languages rewrite child class's method will not override parent's constructor. And I want new `restore` function to work properly

src/modules/ethereum/chains/ChainConfigStore.ts
zerts: Yes, there can be a case when we updating id to custom and then we fetched the info that backend supports this chain. So there should be 2 prev ids in this case.

src/modules/ethereum/chains/ChainConfigStore.ts
zerts: Fixed

src/modules/ethereum/chains/types.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { NetworkConfig } from '@zeriontech/transactions/lib/shared/Networks.type';
+import type { AddEthereumChainParameter } from '../types/AddEthereumChainParameter';
+
+type DeprecatedEthereumChainConfig = NetworkConfigMetaData & {
+  value: NetworkConfig;
+};
+
+export interface EthereumChainConfig extends NetworkConfigMetaData {
+  value: AddEthereumChainParameter;
+  previousIds: string[];
```
everdimension, line 11: maybe it can be null?
```suggestion
  previousIds: string[] | null;
```

src/modules/ethereum/chains/versions.ts
```diff
@@ -0,0 +1,33 @@
...
+    for (const { value, ...config } of entry.ethereumChains || []) {
+      const chainConfig = toAddEthereumChainParameter(value);
+      const id = maybeLocalChainId(value.chain)
+        ? toCustomNetworkId(value.external_id)
+        : value.chain;
+      ethereumChainConfigs?.push({
+        ...config,
+        id,
+        value: chainConfig,
+        previousIds: [value.chain],
```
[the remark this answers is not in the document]
zerts: Fair!
I will update this

package.json
```diff
@@ -93,19 +94,19 @@
...
     "@store-unit/react": "^1.0.4",
     "@tanstack/react-query": "^4.35.3",
-    "@tanstack/react-virtual": "^3.0.0-beta.54",
+    "@tanstack/react-virtual": "^3.2.0",
     "@walletconnect/jsonrpc-provider": "^1.0.13",
     "@walletconnect/jsonrpc-utils": "^1.0.8",
     "@zeriontech/components": "^1.0.1",
     "@zeriontech/hardware-wallet-connection": "^1.0.11",
-    "@zeriontech/transactions": "^0.0.12",
+    "@zeriontech/transactions": "0.0.13",
```
everdimension, line 102: why exact version?

zerts: Just a habit to mention exact version. Will fix

src/ui/pages/Networks/Networks.tsx
```diff
@@ -38,119 +42,79 @@ import { UnstyledLink } from 'src/ui/ui-kit/UnstyledLink';
...
+  prevChain,
+}: {
+  chain: string;
+  chainConfig: AddEthereumChainParameter;
+  prevChain?: string;
+}) {
+  const networks = await networksStore.load(
+    [chain, prevChain].filter(isTruthy)
   );
+  const chainsMetadata = networks.getNetworksMetaData();
```
everdimension, line 83: Why do we need to query metadata here to send it back to the store that already knows about it?

I think we need to send the minimal necessary amount of data. The receiving side can take care of this logic

src/modules/networks/helpers.ts
```diff
@@ -35,24 +40,39 @@ export function toNetworkConfig(
...
+      ? item.rpc_url_public
+      : [],
     nativeCurrency: {
       symbol: item.native_asset?.symbol || '<unknown>',
       decimals: (item.native_asset?.decimals || NaN) as 18,
       name: item.native_asset?.name || '<unknown>',
     },
-    chainId: item.external_id,
+    // deprecated field is being used for chainConfigStore's migration
+    chainId: Networks.getChainId(item) || item.external_id || '',
```
everdimension, line 70: How can `chainId` be an empty string?

src/modules/ethereum/chains/ChainConfigStore.ts
```diff
@@ -61,10 +98,41 @@ export class ChainConfigStore extends PersistentStore<ChainConfig> {
     const chainStr = chain.toString();
     this.setState((state) =>
       produce(state, (draft) => {
-        remove(draft.ethereumChains, (x) => x.value.chain === chainStr);
+        if (!draft.ethereumChainConfigs) {
+          draft.ethereumChainConfigs = [];
+        }
```
everdimension, line 103: I don't think this check is needed anymore, it's guaranteed to exist

zerts: Yes! I fixed this whole thing

src/modules/networks/helpers.ts
zerts: Fixed.

src/ui/pages/Networks/Networks.tsx
zerts: Fixed

package.json
zerts: Fixed

src/modules/persistent-store.ts
```diff
@@ -2,9 +2,9 @@ import { Store } from 'store-unit';
 import * as browserStorage from 'src/background/webapis/storage';
 
 export class PersistentStore<T> extends Store<T> {
-  private key: string;
-  private isReady: boolean;
-  private readyPromise: Promise<void>;
+  protected key: string;
+  protected isReady: boolean;
+  protected readyPromise: Promise<void>;
```
everdimension, line 7: this one can remain `private`

src/ui/ui-kit/SurfaceList/SurfaceList.tsx
```diff
@@ -104,6 +104,7 @@ export const ItemButton = React.forwardRef<
           className={s.decoration}
           style={{
             border: outlined ? '1px solid var(--primary)' : undefined,
+            paddingBlock: 7,
```
everdimension, line 107: 7? 🤨

zerts: To avoid border overflow in virtualized lists :(

zerts: Cause now this border renders out of item, makes its height 50 instead of 48

src/ui/pages/Networks/README.md
```diff
@@ -11,21 +11,19 @@
 
 ### Chain, chainId and external_id
 
-- `chain` is used as a slug-like unique `id`
-- `external_id` is a name for `chainId`
-- `evm_id` is an int representation of `external_id`, therefore is unnecessary
-- for chain configs coming from dApps, `chain` is generated using `nanoid()`
-- for manually added networks, `chain` is currently created equal to `external_id`. This is wrong, because `external_id` can be changed later by the user, but `chain` will remain untouched. If a user manually creates network A with external_id: `'a'`, then later updates external_id to `'b'`, then later creates network `AA`, sets its external_id to `'a'`, network `AA` will overwrite network `A`.
+- HEX chain id is stored in `specification` object inside NetworkConfig
+- for chain configs coming from dApps, `id` is get from backend for the chain with the same `chainId` or directly from `chainId`
```
everdimension, line 15:
```suggestion
- for chain configs coming from dApps, `id` is queried from backend for the chain with the same `chainId` or derived directly from `chainId`
```

src/ui/pages/Networks/README.md
```diff
@@ -11,21 +11,19 @@
...
+- for manually added networks, `id` is currently created equal to `chainId` and will be updated if `chainId` is updated.
```
everdimension, line 16: Uhm, I'm not sure I see this? I think we always derive `id` from `chainId` using `toCustomNetworkId` function. Are there exceptions?

src/ui/pages/Networks/README.md
```diff
@@ -11,21 +11,19 @@
...
 
 ### Creating/modifying a chain
 
-- When we modify existing "mainnet" item `A`, an `EthereumChainConfig` is created, which has the same `chain` and `external_id` as `A`
-- [x] When a new `AddEthereumChainParameter` comes from dapp, an `EthereumChainConfig` is (should be) created with a generated nanoid() for `chain`
-- [x] When an `AddEthereumChainParameter` comes which has a `chainId` matching any existing network, we should use the `chain` from existing network and call `ChainConfigStore.addEthereumChain()`.
+- When we modify existing "mainnet" item `A`, an `EthereumChainConfig` is created, which has the same `id` as `A`
+- [x] When a new `AddEthereumChainParameter` comes from dapp, an `EthereumChainConfig` is (should be) created with an `id` got from backend if we have information about the chain with the same `chainId` or generated from `chainId`.
```
everdimension, line 21:
```suggestion
- [x] When a new `AddEthereumChainParameter` comes from dapp, an `EthereumChainConfig` is (should be) created with an `id` received from backend if we have information about the chain with the same `chainId` or generated from `chainId`.
```

src/ui/pages/Networks/README.md
zerts: You are correct. There is a mistake in readme.

src/ui/shared/channels.mock.ts
```diff
@@ -149,6 +149,8 @@ class WalletPortMock {
       this.setPreference({
         preferences: args[0] as Partial<WalletRecord['publicPreferences']>,
       });
+    } else if (method === 'getWalletGroupByAddress') {
+      return mockRecord.walletManager.groups[0];
```
everdimension, line 153: maybe actually do a search here?

src/ui/pages/Networks/NetworkForm/NetworkForm.tsx
```diff
@@ -163,23 +170,30 @@ export function NetworkForm({
...
+    chainId: (element) => {
+      const value = parsers.chainId(element.value) as ChainId;
+      if (restrictedChainIds.has(value)) {
         return 'Network already exists';
       }
+      try {
+        normalizeChainId(value);
+        normalizeChainId(Number(value));
+      } catch (error) {
+        return `Unsupported chainId${apostrophe}s format`;
```
everdimension, line 195: apostrophe 🤌

zerts: I'm evolving!

src/ui/shared/channels.mock.ts
zerts: May be... I tried to fix this quickly. Will add a TODO for the future :)

PR #437, opened 2024-03-05
Move domain lookup requests to zerion api

zerts:


src/modules/name-service/index.ts
```diff
@@ -1,27 +1,32 @@
...
-    const fulfilled = results.filter(
-      (res) => res.status === 'fulfilled'
-    ) as PromiseFulfilledResult<string>[];
-    return fulfilled.map((res) => res.value);
-  });
-  return addresses.filter((address): address is string => address !== null);
+  try {
+    const response = await ZerionAPI.getWalletsMeta({ identifiers: [address] });
+    return (
+      response.data?.[0].identities
```
everdimension, line 23: should it be `response.data?.[0]?.identities`, like you have in `resolveDomain`?

zerts: Thank you! Good catch!

everdimension (approved):

PR #439, opened 2024-03-05
Update optimistic fee calculation

zerts:


src/ui/pages/History/AccelerateTransactionDialog/shared/accelerate-helpers.ts
```diff
@@ -74,10 +74,6 @@ export function increaseGasPrices(
       draft.info.eip1559.fast.max_fee = increase(max_fee, 1.1);
       draft.info.eip1559.fast.priority_fee = increase(priority_fee, 1.3);
     }
-    if (draft.info.optimistic?.l2) {
```
zerts, line 77: Looks like we will not show the difference in speed up for optimistic chains because of this. And this logic should be refactored for them

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
```diff
@@ -42,22 +47,41 @@ export async function createOptimisticFee({
...
+          nonce,
+          gas,
+          to,
+          value,
+          data,
+          maxPriorityFeePerGas,
+          maxFeePerGas,
+        ])
+      : gasPrice != null
+      ? rlpEncode([nonce, gasPrice, gas, to, value, data])
```
everdimension, line 80: is this order of params for real a part of the optimistic algorithm? :D

src/ui/pages/History/AccelerateTransactionDialog/shared/accelerate-helpers.ts
everdimension: I'm not sure I understand

From the updated calculation code, it looks as if either classic gasPrice or eip1559 is required for the optimistic fee info. Is this correct? @sobolev-igor 
If so, then the increased fee should be displayed. What's the problem with removing this condition?

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
zerts: Yes :((

src/ui/pages/History/AccelerateTransactionDialog/shared/accelerate-helpers.ts
zerts: But as I understand it, the displaying algorithm will try to show optimistic fees here, which are constant. The good thing is that applying configuration will work as well

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
```diff
@@ -69,7 +93,11 @@ export async function createOptimisticFee({
 
   const l1_fee_est = Math.round(l1 * l1GasEstimation);
   return {
-    maxFee: l2 * Number(gas) + Math.round(l1_fee_est * 1.25), // Optimism guarantees that there should not be one-time jumps of l1 component of gas estimation more than 25%
-    estimatedFee: l2 * Number(gas) + l1_fee_est,
+    maxFee:
+      (maxFeePerGas ? Number(maxFeePerGas) : Number(gasPrice)) * Number(gas) +
+      Math.round(l1_fee_est * 1.25), // Optimism guarantees that there should not be one-time jumps of l1 component of gas estimation more than 25%
+    estimatedFee:
+      (maxFeePerGas ? Number(maxFeePerGas) : Number(gasPrice)) * Number(gas) +
```
zerts, line 100: I really didn't want to import base fee here from eip1559 just for this case.

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
sobolev-igor: I've doublechecked and seems like in the final version of [EIP1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification) they decided to use more adequate order of perameters.

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
sobolev-igor: In this case estimated fee will be too off. maxFeePerGas is usually x2 compared to baseFee

src/ui/pages/History/AccelerateTransactionDialog/shared/accelerate-helpers.ts
sobolev-igor: > either classic gasPrice or eip1559 is required for the optimistic fee info

correct! That's how speeding up will be reflected in new estimation even for optimistic fees

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
zerts: So this one is better? `The [EIP-2718](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md) TransactionPayload for this transaction is rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, amount, data, access_list, signature_y_parity, signature_r, signature_s]).`

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
zerts: Tryed to fix

src/modules/ethereum/transactions/gasPrices/optimistic/fee.ts
```diff
@@ -69,7 +92,20 @@ export async function createOptimisticFee({
...
+    estimatedFee:
+      (eip1559GasPrice
+        ? Math.min(
+            Number(eip1559GasPrice.max_fee),
+            Number(base_fee) + Number(eip1559GasPrice.priority_fee)
+          )
+        : Number(classicGasPrice)) *
+        Number(gas) +
+      l1_fee_est,
   };
```
everdimension, line 110: I think this might be more readable:
```ts
  const maxFee = eip1559GasPrice
    ? Number(eip1559GasPrice.max_fee)
    : Number(classicGasPrice);
  const expectedFee = eip1559GasPrice
    ? Math.min(
        Number(eip1559GasPrice.max_fee),
        Number(base_fee) + Number(eip1559GasPrice.priority_fee)
      )
    : Number(classicGasPrice);
  return {
    maxFee: maxFee * Number(gas) + Math.round(l1_fee_est * 1.25), // Optimism guarantees that there should not be one-time jumps of l1 component of gas estimation more than 25%
    estimatedFee: expectedFee * Number(gas) + l1_fee_est,
  };
```

everdimension (approved):

PR #443, opened 2024-03-13
Detect if a Dapp supports EIP-6963

everdimension:


src/content-script/eip6963.ts
```diff
@@ -15,19 +15,33 @@ const info = {
...
 
-type Options = { onRequestProvider: () => void };
+type Options = {
+  onRequestProvider?: () => void;
+  onAccessProvider?: () => void;
+};
 
 export function initializeEIP6963(
   provider: EthereumProvider,
   options?: Options
```
zerts, line 25: Do we need these options to be optional?

src/content-script/in-page.ts
```diff
@@ -145,8 +145,12 @@ if (dappsWithoutCorrectEIP1193Support.has(window.location.origin)) {
 
 initializeEIP6963(provider, {
   onRequestProvider: () => {
-    // DApp supports EIP-6963
     pageObserver.stop();
+    dappDetection.registerEip6963SupportOnce(provider);
+  },
+  onAccessProvider: () => {
```
zerts, line 151: Do we need to keep 2 different callbacks?

src/content-script/eip6963.ts
everdimension: I think it's a good practice, why not?

src/content-script/in-page.ts
everdimension: They have different semantic, I think it's better to be truthful with intentions. Calling "requestProvider" one more time when it's not actually requested can be both unexpected and undesired

PR #447, opened 2024-03-25
Feature/support watchlist wlt 4183

everdimension:
TODO:

- [x] Deduplicate new address during address import
	- [x] Prioritize signer or hardware address over watch address
- [x] Warning dialog for Sign Message views
- [x] Check analytics when readonly wallet is added
- [x] Check DNA registering service for readonly wallets


src/ui/components/Background/Background.tsx
```diff
@@ -50,11 +50,11 @@ const bgClassNames = {
...
 } as const;
 
-export function useBackgroundKind({
-  kind,
-}: {
-  kind: 'neutral' | 'white' | 'transparent';
-}) {
+type Props = { kind: 'neutral' | 'white' | 'transparent' };
+
+export const whiteBackgroundKind: Props = { kind: 'white' };
```
zerts, line 55: why not exporting just Props as a type?

src/background/Wallet/WalletRecord.ts
```diff
@@ -194,21 +199,32 @@ export class WalletRecordModel {
...
   }
 
-  static getWalletByAddress(record: WalletRecord, address: string) {
-    for (const group of record.walletManager.groups) {
-      const wallet = group.walletContainer.getWalletByAddress(address);
-      if (wallet) {
-        return wallet;
+  static getWalletByAddress(
+    record: WalletRecord,
+    { address, groupId }: { address: string; groupId: string | null }
```
zerts, line 204: Can we be in the situation when there are 2 same addresses in different groups? I thought, we remove readonly address when import it.

src/background/Wallet/WalletRecord.ts
```diff
@@ -260,6 +276,44 @@ export class WalletRecordModel {
...
+      /**
+       * NOTE:
+       * (De)duplication logic:
+       * When adding signer/hw wallet, existing readonly wallet should be removed
+       * When adding readonly address, it should be ignored if signer/hw wallet exists
+       * Duplication of addresses between mnemonic/privateKey/HW wallets is allowed
+       */
+
+      // Reuse existing wallet names for new addresses
+      for (const wallet of walletContainer.wallets) {
```
zerts, line 288: Should we be able to add watch address that is already existed as imported?

src/background/Wallet/WalletRecord.ts
```diff
@@ -580,15 +651,19 @@ export class WalletRecordModel {
...
     const normalizedAddress = normalizeAddress(address);
     return produce(record, (draft) => {
+      let didRename = false;
       for (const group of draft.walletManager.groups) {
         for (const wallet of group.walletContainer.wallets) {
           if (normalizeAddress(wallet.address) === normalizedAddress) {
             wallet.name = name || null;
-            return;
+            didRename = true;
+            // NOTE: don't break loop as we expect multiple groups to hold same address
```
zerts, line 660: Oh, I see. Looks a bit confusing, but ok!

src/ui/pages/HardwareWalletConnection/HardwareWalletConnection.tsx
```diff
@@ -92,7 +92,7 @@ export function HardwareWalletConnectionStart({
   const [searchParams] = useSearchParams();
 
   const navigate = useNavigate();
-  const existingAddresses = useAllExistingAddresses();
+  const existingAddresses = useAllExistingMnemonicAddresses();
```
zerts, line 95: What about private addresses here?

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
```diff
@@ -33,13 +33,29 @@ import { WalletAvatar } from 'src/ui/components/WalletAvatar';
...
+  removeWalletSubtitle: (containerType: ContainerType) =>
+    containerType === ContainerType.mnemonic
+      ? 'You can always import it again using your recovery phrase'
+      : containerType === ContainerType.privateKey
+      ? 'You can always import it again using your private key'
+      : containerType === ContainerType.hardware
+      ? 'You can import it again by connecting your hardware wallet'
+      : containerType === ContainerType.readonly
+      ? 'You can always add it back to your watch list'
+      : 'You can add it again on the Manage Wallets page',
```
zerts, line 56: Do we need this fallback text?

zerts:
I don't see watch address icons in select wallet list (the second address is watched)
Zerts Screenshot 2024-03-28 at 10 43@2x

zerts:
When address added as watch address, it is shown as already added in import mnemonic flow
Zerts Screenshot 2024-03-28 at 10 50@2x

zerts:
Does backend support lens?
Zerts Screenshot 2024-03-28 at 10 51@2x

zerts:
Do we need to warn user somehow when they are trying to add already imported address as watched?

zerts:
As I remember, we've discussed with design that every watch address should be in its own group
Zerts Screenshot 2024-03-28 at 10 54@2x

This is not a blocker at all, just interesting just noticed.

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
```diff
@@ -33,13 +33,29 @@ import { WalletAvatar } from 'src/ui/components/WalletAvatar';
...
   isMnemonicContainer,
-  isPrivateKeyContainer,
   isSignerContainer,
 } from 'src/shared/types/validators';
 import { NeutralDecimals } from 'src/ui/ui-kit/NeutralDecimals';
 
+const strings = {
+  recoveryPhraseTitle: 'Recovery Phrase',
+  privateKeyTitle: 'Private Key',
+  removeWalletSubtitle: (containerType: ContainerType) =>
```
zerts, line 47: Meybe we can rename this property somehow. Looks like a method to remove something :))
What about `getRemoveSubtitle`?

src/ui/pages/GetStarted/GetStarted.tsx
```diff
@@ -95,8 +120,69 @@ function Options() {
     '/connect-hardware-wallet',
     beforeCreate
   );
+  const addReadonlyHref = createNextHref('/get-started/readonly', beforeCreate);
   return (
     <Background backgroundKind="white">
```
zerts, line 125: Let's change <Background> to useBackgroundKind here?

src/ui/pages/GetStarted/GetStarted.tsx
```diff
@@ -73,6 +79,21 @@ function NewWalletOption({
   );
 }
 
+function SpaceZone({ children }: React.PropsWithChildren) {
```
zerts, line 82: I can see duplications with my updates for settings, but not something critical

src/ui/pages/ManageWallets/WalletAccount/WalletAccount.tsx
```diff
@@ -159,13 +177,18 @@ export function WalletAccount() {
...
-    queryKey: ['wallet/uiGetWalletByAddress', address],
-    queryFn: () => walletPort.request('uiGetWalletByAddress', { address }),
+    queryKey: ['wallet/uiGetWalletByAddress', address, groupId],
+    queryFn: () =>
+      walletPort.request('uiGetWalletByAddress', { address, groupId }),
     useErrorBoundary: true,
   });
+  const { data: walletGroup, isLoading: walletGroupIsLoading } = useQuery({
+    queryKey: ['getWalletGroupByAddress', address],
+    queryFn: () => getWalletGroupByAddress(address),
```
zerts, line 187: Should `useErrorBoundary: true` be added here?

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  const isDomainResolving = query && isInitialLoading && !isEthereumAddress;
+  const resolveError = isError && !isEthereumAddress;
+
+  useCustomValidity({
+    ref: inputRef,
+    customValidity: isEthereumAddress
+      ? '' // Form MUST be valid for valid eth address regardless of the getWalletsMeta request state
+      : isDomainResolving
+      ? 'Wait until address is resolved'
```
zerts, line 134: I'm not sure that validity here is the best option. I'm getting this "bad format-like" errors after just quick pressing enter and it looks like there is something wrong with the input.

src/ui/pages/Overview/Overview.tsx
```diff
@@ -314,7 +314,10 @@ function OverviewComponent() {
               icon={
                 <WalletSourceIcon
                   address={singleAddress}
+                  groupId={null}
```
zerts, line 317: Not sure is it a design requirement, but imho it would be nice to see wallet type here too.

src/ui/components/WalletSourceIcon/WalletSourceIcon.tsx
```diff
@@ -25,5 +36,19 @@ export function WalletSourceIcon({
     <LedgerIcon
```
zerts, line 36: Let's add `cutoutStroke` to other icons

src/ui/components/SignTransactionButton/ReadonlyWarningDialog/ReadonlyWarningDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+    // with "max update depth exceeded" ¯\_(ツ)_/¯. Not a good sign, but afaik it should be fixed
+    // in the next version of '@tanstack/react-query'
+    suspense: false,
+  });
+  const isReadonlyGroup = walletGroup
+    ? isReadonlyContainer(walletGroup.walletContainer)
+    : null;
+  return (
+    <>
+      <ReadonlyWarningDialog ref={dialogRef} />
```
zerts, line 85: Do we need to render this dialog if isReadOnly === false?

src/ui/components/Background/Background.tsx
everdimension: The idea is to export a stable object so that you don't need to wrap in useMemo for popular cases:

```
useBackgroundKind(useMemo(() => ({ kind: 'white' }), []))
```
vs
```
useBackgroundKind(whiteBackgroundKind)
```

src/background/Wallet/WalletRecord.ts
everdimension: No, that's the only group for which duplication isn't allowed :) When you add existing address as a watch address, the record is not updated except for making the new address a current one.

src/ui/pages/HardwareWalletConnection/HardwareWalletConnection.tsx
everdimension: That's one of the things fixed in the updated logic. Say you imported a single address by private key, but later you're ready to import the whole group. Currently we would show the address as "already added" and block adding it. I think this can be frustrating

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
everdimension: Oh, good point!!

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
everdimension: Good question. Do you think we should throw?

everdimension:
> Does backend support lens? 

I reported to them, I think it temporarily doesn't 🤷‍♀️

src/ui/pages/GetStarted/GetStarted.tsx
everdimension: Yeah ok!

src/ui/pages/ManageWallets/WalletAccount/WalletAccount.tsx
everdimension: 🤔  Makes sense, but currently there's a "NotFound" page being used. I have to remember whether this flow was on purpose

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
```diff
@@ -0,0 +1,249 @@
...
+    };
+  } else {
+    return {
+      address: data.data[0]?.address ?? null,
+      domains: data.data[0]?.identities.map((value) => value.handle) || null,
+    };
+  }
+}
+
+async function lookup(value: string) {
```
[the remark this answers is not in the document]
everdimension: Hm maybe, I thought of this as a local helper so that the code inside the component is more readable.

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
everdimension: It's a question of how to display validity error. What would you suggest should happen when submit button is clicked during loading? Except for disabling button :)

src/ui/pages/Overview/Overview.tsx
everdimension: But you do! :) we just don't know which group the "current address" belongs to, so we would just get the first one and display the icon for it

src/ui/components/WalletSourceIcon/WalletSourceIcon.tsx
everdimension: Maybe I misunderstood the design but it seemed to me that only "watch" one needs it 🤔

src/ui/components/SignTransactionButton/ReadonlyWarningDialog/ReadonlyWarningDialog.tsx
everdimension: Hmmm, also good point 🤔  I think we do not

src/ui/pages/HardwareWalletConnection/HardwareWalletConnection.tsx
everdimension: Ah, wait. I see, this is ledger's import flow. Right, I guess we should use all transacting wallets here

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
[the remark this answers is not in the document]
everdimension: Oh, I misread your question. Nope, this logic is new: it's about automatically adding `.eth` and `.lens` domains to a query like `"hello"`

src/ui/components/WalletSourceIcon/WalletSourceIcon.tsx
zerts: What about Ledger?

src/ui/pages/Overview/Overview.tsx
zerts: Oh, sorry, forgot to delete comment after checking old version

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
zerts: Yes:)

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
zerts: I think, the best way is to apply submit with delay in this case.
However, I understand, that this can be a tricky thing to do. 
So as an easy solution, I would probably show some yellow text like "try again after resolving the address"

src/ui/pages/GetStarted/AddReadonlyAddress/AddReadonlyAddress.tsx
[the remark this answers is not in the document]
zerts: Hm, ok

zerts (approved):

src/ui/components/WalletSourceIcon/WalletSourceIcon.tsx
everdimension: It follows the design currently, as I understand

PR #456, opened 2024-04-04
Fix Back to Home button in global error

zerts:


src/ui/components/ViewError/ViewError.tsx
```diff
@@ -131,7 +133,7 @@ export function ViewError({
               if (pageTemplateType !== 'dialog') {
                 navigate('/');
               }
-              window.location.reload();
```
everdimension, line 134: Can something break because we're no longer reloading the page?

zerts: Good question. I'll check

zerts: I don't know, tbh.
My logic was: we are updating a prop for ErrorBoundary, so it should be updated after going to home screen. However, I think window.reload can be helpful

everdimension: If we're reloading, what does `onRetry` accomplish?

zerts: This looked like GlobalBoundary wasn't updated during this reload properly. I will check again

zerts:
Approach didn't work properly

PR #457, opened 2024-04-04
Fix filtering in swap form

zerts:


zerts:
> I'm hesitant about this API. Why would `getGroupName` return something for some values and nothing for others? Does that mean values with no group will visually get attached to the previous group?
> 
> Turning off grouping is available already: pass `getGroupName={undefined}` at the call site. Let's update this?

Hm, Ok.
The idea actually was the same. Like is there is search, we don't need to divide popular and other tokens even if popular are still there. But ok, pass `undefined` looks good!

everdimension (approved):

PR #459, opened 2024-04-10
Chore/update json rpc tools

everdimension:
* Updates from unmaintained `@json-rpc-tools` to a more up-to-date `@walletconnect` tools
* Updates `payloadId()` generator with a custom one that should guarantee uniqueness


zerts (approved):

vyorkin (approved):

src/shared/custom-rpc/getPayloadId.ts
```diff
@@ -0,0 +1,12 @@
+let base = 1;
+const typedArray = new Uint16Array(1);
+
+/**
+ * returns a unique random value always larger than the previous one
+ */
+export function getPayloadId(): number {
+  if (base >= 2 ** 15 - 1) {
```
vyorkin, line 8: maybe it would be better to make it a named constant, like "max int16" or smth like that

src/shared/custom-rpc/getPayloadId.ts
```diff
@@ -0,0 +1,9 @@
+let i = 1;
+const initialValue = crypto.getRandomValues(new Uint32Array(1))[0];
+
+/**
+ * returns a unique random value always larger than the previous one
```
zerts, line 5: May be this can be rephrased, cause numbers are not random any more, only the base is random. However, not critical

PR #460, opened 2024-04-16
Set up currency

zerts:


src/background/Wallet/GlobalPreferences.ts
```diff
@@ -26,6 +26,7 @@ export interface State {
   providerInjection?: ProviderInjection;
   autoLockTimeout?: number | 'none';
   walletNameFlags?: Record<string, WalletNameFlag[] | undefined>;
+  currency?: string;
```
zerts, line 29: I decided to put currency into global settings, so we are able to reach it anywhere in background without passing the Wallet into props

bkvvldmr (approved):
Great work, approved, please merge!

zerts:
The alternative version of https://github.com/zeriontech/zerion-wallet-extension/pull/478

PR #462, opened 2024-04-18
Implement Hold to Sign button

zerts:


src/ui/pages/History/AccelerateTransactionDialog/CancelTx/CancelTx.tsx
```diff
@@ -56,6 +57,7 @@ function CancelTxContent({
   onDismiss: () => void;
   onSuccess: () => void;
 }) {
+  const { preferences } = usePreferences();
```
zerts, line 60: Unfortunately, useQuery doesn't work inside <Content/> right now, so I placed this preferences check outside of SignButtons. Will be glad to discuss better way if needed.

src/ui/pages/SendTransaction/SendTransaction.tsx
```diff
@@ -230,7 +232,11 @@ function TransactionDefaultView({
       });
     },
     onMutate: () => 'sendTransaction',
-    onSuccess: (tx) => handleSentTransaction(tx),
+    onSuccess: (tx) =>
+      setTimeout(
+        () => handleSentTransaction(tx),
+        preferences?.showHoldToSignButton ? 500 : 0
```
zerts, line 238: As I understand, we want to show success state for a second to the user before closing the window

src/ui/pages/History/AccelerateTransactionDialog/CancelTx/CancelTx.tsx
```diff
@@ -205,11 +207,14 @@ function CancelTxContent({
             >
               Back
             </Button>
-            <SignTransactionButton
-              wallet={wallet}
-              ref={signTxBtnRef}
-              onClick={() => sendTransaction()}
-            />
+            {preferences ? (
```
zerts, line 210: I didn't see any jumps in UI because of not rendering the button. But I'm not 100% sure this is the base option here.

src/background/Wallet/WalletRecord.ts
```diff
@@ -769,6 +769,7 @@ export class WalletRecordModel {
       recentAddresses: [],
       mintDnaBannerDismissed: false,
       upgradeDnaBannerDismissed: false,
+      showHoldToSignButton: true,
```
everdimension, line 772: I don't "show" is the right name here :)

src/ui/pages/SignTypedData/SignTypedData.tsx
```diff
@@ -337,34 +339,42 @@ function TypedDataDefaultView({
...
-
-              <SignMessageButton
-                wallet={wallet}
-                ref={signMsgBtnRef}
-                onClick={() => {
-                  if (!seenSigningData) {
-                    scrollSigningData();
-                  } else {
-                    signTypedData_v4();
+              {preferences ? (
```
everdimension, line 342: I would maybe bail out of rendering the whole component until we receive preferences

src/ui/pages/SignInWithEthereum/SignInWithEthereum.tsx
```diff
@@ -274,18 +276,21 @@ export function SignInWithEthereum() {
...
+                wallet={wallet}
+                onClick={() => personalSign()}
+                buttonTitle={
+                  personalSignMutation.isLoading
+                    ? `Signing In${ellipsis}`
+                    : !siweMessage?.isValid() && isDeviceAccount(wallet)
+                    ? 'Proceed anyway'
+                    : 'Sign In'
+                }
+                holdToSign={preferences.enableHoldToSignButton}
```
everdimension, line 291: doesn't `SignMessageButton` currently read preferences from inside? 🤔  confusing

src/ui/pages/SignTypedData/SignTypedData.tsx
```diff
@@ -337,34 +339,42 @@ function TypedDataDefaultView({
...
+                        alignItems="center"
+                        justifyContent="center"
+                      >
+                        <span>Scroll</span>
+                        <ArrowDownIcon style={{ width: 24, height: 24 }} />
+                      </HStack>
+                    )
+                  }
+                  holdToSign={
+                    preferences.enableHoldToSignButton && seenSigningData
```
everdimension, line 374: do we need to check `seenSigningData` here? I think we're not rendering children anyway if it's `false`

src/ui/ui-kit/Button/Button.tsx
```diff
@@ -17,9 +17,9 @@ export type Kind =
   | 'warning'
   | 'loading-border'
   | 'text-primary';
-type Size = 60 | 56 | 48 | 44 | 40 | 36 | 32 | 28;
+export type Size = 60 | 56 | 48 | 44 | 40 | 36 | 32 | 28;
 
-const borderRaduis: Record<Size, number> = {
+export const borderRaduis: Record<Size, number> = {
```
everdimension, line 22:
```suggestion
export const borderRadius: Record<Size, number> = {
```

src/ui/ui-kit/Button/HoldableButton.tsx
```diff
@@ -0,0 +1,229 @@
...
+          }
+        }}
+        onMouseUp={handleMouseUp}
+        onKeyDown={(e) => {
+          e.preventDefault();
+          if (e.key === 'Enter' && innerState === 'idle') {
+            handleMouseDown();
+          }
+        }}
+        onKeyUp={handleMouseUp}
```
everdimension, line 159: 1. (harmless) If I press any keyboard button and hold it, then hold click, then release keyboard button, the action will cancel
2. (dangerous, probably a bug) If I start holding "Enter", then do a quick click (click and release), then release enter, the animation seemingly cancels, but then fills up and completes! And transaction submits

everdimension (changes requested):

src/ui/pages/History/AccelerateTransactionDialog/CancelTx/CancelTx.tsx
```diff
@@ -209,6 +209,7 @@ function CancelTxContent({
               wallet={wallet}
               ref={signTxBtnRef}
               onClick={() => sendTransaction()}
+              holdToSign={true}
```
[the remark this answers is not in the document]
zerts: This is just a flag that shows inside component is there allowed to use holdToSign or not. For example, we don't need hold to sign in internal forms, cause we have another confirmation screen after the click. But I will rename the prop here

src/ui/components/SignTransactionButton/SignTransactionButton.tsx
```diff
@@ -33,41 +41,57 @@ export const SignTransactionButton = React.forwardRef(
...
-            ...params,
-          });
-        } else {
-          return await walletPort.request('signAndSendTransaction', [
-            transaction,
-            params,
-          ]);
+      mutationFn: async (params: SendTxParams) => {
+        const result = await sendTransactionInner(params);
+        if (!isDeviceAccount(wallet) && holdToSign) {
```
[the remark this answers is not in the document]
zerts: Yes, there is a bug

src/ui/pages/History/AccelerateTransactionDialog/CancelTx/CancelTx.tsx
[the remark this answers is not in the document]
zerts: I also avoided to read preferences at this level, cause we will need to read them in 3 different places then.

src/ui/pages/SignInWithEthereum/SignInWithEthereum.tsx
zerts: Yes. Because after the last refactor it was extracted from <Content> and now it is able to use useQuery hooks inside it. So I took this opportunity :))

src/ui/ui-kit/Button/HoldableButton.tsx
zerts: Hm, this is interesting. I didn't want to overcomplicate this component. The second one is a bug, however, I am not sure that it is really a dangerous. We tried to prevent misclicks, not beat the case when user is playing with this button and challenging is. I can think about a fix here or we can postpone it for later.

src/ui/pages/SignTypedData/SignTypedData.tsx
zerts: This is a possible solution. Ideally, this preference fetch should be inside the component to avoid same code in multiple places.

src/ui/pages/SignTypedData/SignTypedData.tsx
zerts: I need to double check, but this looks like a fair condition anyway. If we show `Scroll` button, we don't need to use holdable state

zerts: It looks like we are rendering them when it is false, btw. Anyway, let me check :)

zerts: Yes, this condition is important

src/ui/components/SignTransactionButton/SignTransactionButton.tsx
```diff
@@ -33,41 +41,57 @@ export const SignTransactionButton = React.forwardRef(
...
       wallet: ExternallyOwnedAccount;
       buttonTitle?: React.ReactNode;
       buttonKind?: ButtonKind;
+      holdToSign: boolean;
     },
     ref: React.Ref<SendTxBtnHandle>
   ) {
+    const { preferences } = usePreferences();
     const hardwareSignRef = useRef<SignTransactionHandle | null>(null);
+    const { mutateAsync: sendTransactionInner, ...sendTxMutationInner } =
```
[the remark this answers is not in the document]
zerts: What I tried here is to get the loading/completing/error state directly from different mutations. May be this can be refactored this way. Do we need to do it in this stage? I would probably do it separately, cause this will probably cause more potential bugs and we will need to test it carefully again before initial merge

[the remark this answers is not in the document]
zerts: I also always try to avoid async functions inside onSucess callbacks if possible. Not sure is it a right approach.

src/background/Wallet/WalletRecord.ts
```diff
@@ -786,6 +788,11 @@ export class WalletRecordModel {
   ) {
     return produce(record, (draft) => {
       Object.assign(draft.publicPreferences, preferences);
+      emitter.emit(
+        'preferencesChange',
+        draft.publicPreferences,
+        record.publicPreferences
+      );
```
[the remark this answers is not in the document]
zerts: Oh, yes, I see the problem. However, ideally I need a way to pass previousState and currentState together. I need to think about the best approach here. May be I will narrowing the event to `holdToSignPreferenceChange`

src/ui/components/SignTransactionButton/SignTransactionButton.tsx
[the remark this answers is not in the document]
everdimension: should be fixed?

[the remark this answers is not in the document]
zerts: Oh, didn't I uploaded the updates

src/ui/ui-kit/Button/HoldableButton.tsx
everdimension: The point of this button is to prevent accidental submissions. If it can create accidental submission, we should fix that

src/ui/components/SignTransactionButton/SignTransactionButton.tsx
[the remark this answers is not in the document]
everdimension: I think it's much more important to avoid mixing concerns. The UI elements of choice should not affect signing logic.

> cause this will probably cause more potential bugs

we should avoid merging something that "will cause more bugs" :)

src/ui/pages/SendTransaction/SendTransaction.tsx
```diff
@@ -642,7 +644,12 @@ function SendTransactionContent({
   }, []);
 
   const next = params.get('next');
-  function handleSentTransaction(tx: ethers.providers.TransactionResponse) {
+  async function handleSentTransaction(
+    tx: ethers.providers.TransactionResponse
+  ) {
+    if (preferences?.enableHoldToSignButton) {
+      await wait(500);
```
everdimension, line 651: comment what this delay is for?

src/ui/pages/SignInWithEthereum/SignInWithEthereum.tsx
```diff
@@ -100,7 +103,12 @@ export function SignInWithEthereum() {
     // a global onError handler (src/ui/shared/requests/queryClient.ts)
     // TODO: refactor to just emit error directly from the mutationFn
     onMutate: () => 'signMessage',
-    onSuccess: handleSignSuccess,
+    onSuccess: async (signature) => {
+      if (preferences?.enableHoldToSignButton) {
+        await wait(500);
```
everdimension, line 108: comment what this delay is for?

src/ui/components/SignMessageButton/SignMessageButton.tsx
```diff
@@ -91,32 +99,68 @@ export const SignMessageButton = React.forwardRef(function SignMessageButton(
...
   const isLoading =
     personalSignMutation.isLoading || signTypedData_v4Mutation.isLoading;
+  const isSuccess =
+    personalSignMutation.isSuccess || signTypedData_v4Mutation.isSuccess;
+  const isError =
+    personalSignMutation.isError || signTypedData_v4Mutation.isError;
+
+  // there is a small delay after using a holdable button
+  // button should be disabled after successful sign to prevent a duplicating call
+  const disabled = isLoading || isSuccess;
```
everdimension, line 109: should this be instead
```suggestion
  const disabled = isLoading || (holdToSign && isSuccess);
```

src/ui/components/SignTransactionButton/SignTransactionButton.tsx
```diff
@@ -69,38 +77,76 @@ export const SignTransactionButton = React.forwardRef(
...
+
     useImperativeHandle(ref, () => ({ sendTransaction }));
 
     const isLoading = isLoadingProp || sendTxMutation.isLoading;
     const isSending = sendTxMutation.isLoading;
-    const disabled = isLoading || disabledAttr;
+
+    // there is a small delay after using a holdable button
+    // button should be disabled after successful sign to prevent a duplicating call
+    const disabled = isLoading || sendTxMutation.isSuccess || disabledAttr;
```
everdimension, line 88: maybe like this?
```suggestion
    const disabled = isLoading || (holdToSign && sendTxMutation.isSuccess) || disabledAttr;
```

src/ui/pages/SignMessage/SignMessage.tsx
```diff
@@ -79,7 +82,12 @@ function SignMessageContent({
     // a global onError handler (src/ui/shared/requests/queryClient.ts)
     // TODO: refactor to just emit error directly from the mutationFn
     onMutate: () => 'signMessage',
-    onSuccess: handleSignSuccess,
+    onSuccess: async (signature) => {
+      if (preferences?.enableHoldToSignButton) {
+        await wait(500);
```
everdimension, line 87: comment for this delay?

src/ui/ui-kit/Button/HoldableButton.tsx
```diff
@@ -0,0 +1,233 @@
...
+        )}
+        onMouseDown={(e) => {
+          if (e.button === 0) {
+            e.preventDefault();
+            handleMouseDown();
+          }
+        }}
+        onMouseUp={handleMouseUp}
+        onKeyDown={(e) => {
+          e.preventDefault();
```
everdimension, line 158: A bug: once you focus this button, you cannot tab away from it

everdimension (approved):
See latest comments please

src/ui/components/SignMessageButton/SignMessageButton.tsx
zerts: This is fair!

PR #463, opened 2024-04-18
Wrap define ethereum object in try catch

zerts:


src/content-script/in-page.ts
```diff
@@ -121,23 +121,28 @@ dappDetection.onChange(({ dappIsZerionAware }) => {
...
+      dappDetection.handleForeignProvider(value);
+      competingProviders.handleForeignProvider({
+        foreignProvider: value,
+        ourProvider: provider,
+      });
+    },
+  });
+} catch {
+  // eslint-disable-next-line no-console
+  console.warn('Global ethereum object was not defined');
```
everdimension, line 144: maybe a bit more explicit:
```suggestion
  console.warn('Failed to set window.ethereum');
```

PR #464, opened 2024-04-23
Update gas price api

zerts:
Depends on:
- https://github.com/zeriontech/transactions-ui/pull/8

### DIFF
--- a/src/modules/ethereum/transactions/gasPrices/GasPriceObject.ts
+++ b/src/modules/ethereum/transactions/gasPrices/GasPriceObject.ts
@@ -2,8 +2,11 @@ import type { EIP1559 } from '@zeriontech/transactions';
 import type { EIP1559Base } from './EIP1559';
 
 export interface GasPriceObject {
-  classic?: number | string | null;
-  eip1559?: EIP1559 | null;
+  classic: number | string | null;
+  eip1559: EIP1559 | null;
+  optimistic: {
+    underlying: { classic: number | string | null; eip1559: EIP1559 | null };
+  } | null;
 }
 
 export interface GasPriceWithBaseObject {
--- a/src/modules/ethereum/transactions/gasPrices/requests.ts
+++ b/src/modules/ethereum/transactions/gasPrices/requests.ts
@@ -1,48 +1,9 @@
 import { ethers } from 'ethers';
-import { client, mergeSingleEntity } from 'defi-sdk';
-import type { EIP1559 } from '@zeriontech/transactions';
-import { rejectAfterDelay } from 'src/shared/rejectAfterDelay';
 import type { Chain } from 'src/modules/networks/Chain';
 import { sendRpcRequest } from 'src/shared/custom-rpc/rpc-request';
 import type { Networks } from 'src/modules/networks/Networks';
-import { SLOW_MODE } from 'src/env/config';
-import { wait } from 'src/shared/wait';
-
-export interface OptimisticGasPriceInfo {
-  l1?: number;
-  fixed_overhead?: number;
-  dynamic_overhead?: number;
-}
-
-export interface EIP1559GasPrices {
-  base_fee: number;
-  fast: EIP1559 | null;
-  rapid: EIP1559 | null;
-  slow: EIP1559 | null;
-  standard: EIP1559 | null;
-}
-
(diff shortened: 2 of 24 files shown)


src/modules/ethereum/transactions/gasPrices/GasPriceObject.ts
```diff
@@ -2,8 +2,11 @@ import type { EIP1559 } from '@zeriontech/transactions';
 import type { EIP1559Base } from './EIP1559';
 
 export interface GasPriceObject {
-  classic?: number | string | null;
-  eip1559?: EIP1559 | null;
+  classic: number | string | null;
+  eip1559: EIP1559 | null;
+  optimistic: {
+    underlying: { classic: number | string | null; eip1559: EIP1559 | null };
+  } | null;
```
everdimension, line 9: what's the point of this property if it only contains `classic` and `eip1559` data?

src/modules/ethereum/transactions/gasPrices/requests.ts
```diff
@@ -53,93 +14,35 @@ export async function fetchGasPriceFromNode(
...
   try {
-    if (networks.supports('sending', chain)) {
-      const gasPrices = await gasChainPricesSubscription.get();
-      const chainGasPrices = gasPrices[chain.toString()];
-      if (chainGasPrices) {
-        return chainGasPrices;
-      } else {
-        throw new Error('unable to get gas prices from api');
-      }
+    const response = await ZerionAPI.getGasPrices({ chain });
```
everdimension, line 40: We used to update the value no more often than the socket api returned new data.
Now we're making new request each time, right?
Let's maybe use `queryClient` to fetch no more often than 5-10 seconds?

src/ui/pages/SendTransaction/TransactionConfiguration/useTransactionFee.ts
```diff
@@ -195,30 +187,26 @@ function calculateTransactionCosts({
...
+      ({ eip1559: eip1559Info }) => eip1559Info && isMatch(eip1559Info, eip1559)
+    );
 };
 
 function getFeeTime(
   gasPrice: GasPriceObject | null,
   chainGasPrices: ChainGasPrice | null
 ) {
-  if (gasPrice?.eip1559 && chainGasPrices?.info.eip1559) {
+  if (gasPrice?.eip1559 && chainGasPrices?.fast.eip1559) {
```
everdimension, line 206: we're not checking "average" here, not sure if critical

everdimension (approved):
Let's do this one? https://github.com/zeriontech/zerion-wallet-extension/pull/464/files#r1613418127

src/modules/ethereum/transactions/gasPrices/GasPriceObject.ts
zerts: This is just a data object from backend. Not sure how it is built

src/modules/ethereum/transactions/gasPrices/requests.ts
zerts: Hmmm
Let me think about it.

src/ui/pages/SendTransaction/TransactionConfiguration/useTransactionFee.ts
zerts: Should not be critical, cause fast option is a default one. Good point anyway

src/modules/ethereum/transactions/gasPrices/GasPriceObject.ts
zerts: As I understand, this should help us to build optimistic fees, and at the same time gas price object doesn't have its own classic or eip info to avoid confusion

src/modules/ethereum/transactions/gasPrices/requests.ts
zerts: This is true. Let me add cache here. 
This will make a difference only for the case when we add gas price info for transactions without gas price info. Cause there we are doing 2 requests in a row.

zerts: Or may be not. Let's discuss it

PR #470, opened 2024-04-30
Show addresses in wallet select

everdimension:


src/ui/components/CopyButton/CopyButton.tsx
```diff
@@ -48,12 +73,17 @@ export function CopyButton({ address, title = 'Copy Address' }: Props) {
             color: 'var(--white)',
             boxShadow: 'var(--elevation-200)',
             position: 'absolute',
-            top: 2,
-            right: 0,
-            transform: 'translateX(100%)',
             padding: '4px 8px',
             borderRadius: 4,
             whiteSpace: 'nowrap',
+            top: tooltipPosition === 'center-bottom' ? undefined : 2,
```
zerts, line 79: I think, this can be one separate object, which can be spread with condition here

everdimension: Thought of that but spreading loses CSSProperties types for some reason, and since these are still very much hardcoded values and not a truly adaptive tooltip I don't think it's worth to bother. Once css anchors make it to stable versions we can finally implement them in the ui-kit :)

src/ui/pages/WalletSelect/WalletSelect.tsx
```diff
@@ -100,6 +100,7 @@ export function WalletSelect() {
               setCurrentAddressMutation.mutate(wallet.address);
             }}
             selectedAddress={singleAddress}
+            showAddressValues={true}
```
zerts, line 103: Maybe just `showAddresses`?

zerts (approved):

src/ui/pages/WalletSelect/WalletSelect.tsx
everdimension: I think this is more descriptive, my reasoning:
* `showAddress` is a single form, but this is a list
* `showAddresses` is less clear because "address" and "wallet" are often used interchangible, and this is already a wallet list. "Address value" seems to better convey that it means the raw `0x123...` value

zerts (approved):

PR #471, opened 2024-05-03
Refactor Wallet and PersistentStore [WLT-4849]

everdimension:
Remove redundant awaits and possible race conditions

### DIFF
--- a/src/modules/persistent-store.ts
+++ b/src/modules/persistent-store.ts
@@ -16,13 +16,11 @@ export class PersistentStore<T> extends Store<T> {
     this.isReady = false;
     this.readyPromise = this.restore();
     this.on('change', (state) => {
-      // TODO: FIX:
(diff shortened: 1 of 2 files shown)


src/modules/persistent-store.ts
```diff
@@ -33,10 +31,21 @@ export class PersistentStore<T> extends Store<T> {
...
 
+  setState(...args: Parameters<Store<T>['setState']>) {
+    if (!this.isReady) {
+      if (process.env.NODE_ENV === 'development') {
+        throw new Error(
+          'You are trying to write to a PersistentStore before {ready}'
+        );
+      }
+    }
+    return super.setState(...args);
```
zerts, line 42: Are we ok with writing if this.isReady is false? Cause in production it looks like we are going to write anyway

zerts (approved):

PR #478, opened 2024-05-07
Set up Currency 2

zerts:


src/modules/currency/currencies.ts
```diff
@@ -0,0 +1,173 @@
...
+    locale: 'zh-cn',
+    type: CurrencyType.Fiat,
+  },
+  rub: {
+    name: 'Russian Ruble',
+    code: 'rub',
+    symbol: '₽',
+    locale: 'ru',
+    type: CurrencyType.Fiat,
+    modifyParts: setCustomSymbol('₽'),
```
zerts, line 112: For some reason extension decided not to use symbol for some currencies, so I had to modify them manually

src/modules/currency/currencies.ts
```diff
@@ -0,0 +1,170 @@
...
+export enum CurrencyType {
+  Fiat,
+  Crypto,
+}
+
+export interface CurrencyConfig {
+  name: string;
+  code: string;
+  symbol: string;
+  locale: string;
```
everdimension, line 26: Do we need locale? We're not using it

src/modules/currency/currencies.ts
everdimension: The reason in "en" locale :)

src/modules/currency/useCurrency.ts
```diff
@@ -0,0 +1,16 @@
+import { useStore } from '@store-unit/react';
+import { useCallback } from 'react';
+import { preferenceStore } from 'src/ui/features/appearance';
+
+export function useCurrency() {
+  const { currency } = useStore(preferenceStore);
+  return {
+    currency,
+    setCurrency: useCallback((value: string) => {
```
everdimension, line 9: I don't think it's worth to create this function for every hook call. It's used only once: on the config page. We can access `preferenceStore` directly there

src/modules/currency/currencies.ts
```diff
@@ -0,0 +1,170 @@
...
+  },
+  eth: {
+    name: 'Ether',
+    code: 'eth',
+    symbol: 'Ξ',
+    locale: 'en-US',
+    modifyParts: setCustomSymbol('Ξ'),
+    getFormatterConfig: memoize((value) => {
+      return {
+        currencyDisplay: 'symbol',
```
everdimension, line 61: What does this do?

src/modules/currency/currencies.ts
```diff
@@ -0,0 +1,170 @@
...
+    type: CurrencyType.Fiat,
+  },
+  eth: {
+    name: 'Ether',
+    code: 'eth',
+    symbol: 'Ξ',
+    locale: 'en-US',
+    modifyParts: setCustomSymbol('Ξ'),
+    getFormatterConfig: memoize((value) => {
+      return {
```
everdimension, line 60: let's create this object statically so that we're returning the same object for different values, too

src/ui/pages/Settings/Preferences.tsx
```diff
@@ -0,0 +1,157 @@
...
+  return (
+    <PageColumn>
+      <NavigationTitle title="Currency" />
+      <PageTop />
+
+      <form
+        onSubmit={(e) => {
+          e.preventDefault();
+          const { currency: currencyValue } = collectData(e.currentTarget, {
+            currency: (x) => x as string,
```
everdimension, line 41: Do we need this "parser"?

src/ui/pages/Settings/Preferences.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+export function PreferencesPage() {
+  return (
+    <Routes>
+      <Route
+        path="/"
+        element={
+          <ViewSuspense>
+            <PreferencesMain />
```
everdimension, line 143: Why do we need an additional screen? We can display "currency" option in Settings

zerts: This is definitely need a discussion :))
I will refactor save/reset button for now

src/ui/pages/Settings/Preferences.tsx
zerts: Probably not. Just for types

src/modules/currency/currencies.ts
zerts: No, left this for the future, but let's remove it

src/modules/currency/currencies.ts
zerts: Removed!

src/modules/currency/currencies.ts
zerts: Ok!

src/modules/currency/currencies.ts
everdimension: Locale shouldn't be really tied to currency I believe)

src/ui/ui-kit/NeutralDecimals/NeutralDecimals.tsx
```diff
@@ -7,13 +9,17 @@ export function NeutralDecimals({
   parts: Intl.NumberFormatPart[];
   neutralColor?: string;
 }) {
+  const { currency } = useCurrency();
```
everdimension, line 12: The value being formatted can be unrelated to current currency

everdimension (changes requested):
https://github.com/zeriontech/zerion-wallet-extension/pull/478/files?diff=unified&w=0#r1631023475

src/modules/currency/currencies.ts
```diff
@@ -0,0 +1,135 @@
...
+  numberingSystem: string;
+  style: string;
+  currency?: string;
+  currencyDisplay?: string;
+  minimumIntegerDigits: number;
+  minimumFractionDigits: number;
+  maximumFractionDigits: number;
+  minimumSignificantDigits?: number;
+  maximumSignificantDigits?: number;
+  useGrouping: boolean;
```
everdimension, line 14: I think we shouldn't mention values that we're not relying upon. Or we should just reuse `NumberFormatOptions` type

everdimension (approved):

abalyasnikov:
qa verified

PR #483, opened 2024-05-13
Open backup flow in tab view

everdimension:
This fixes issues with losing state when
the extension popup closes (for any reason) and you need to go through flow again from the start to finish it


src/ui/pages/BackupWallet/BackupSettingsItem/BackupSettingsItem.tsx
```diff
@@ -75,6 +76,7 @@ export function BackupFlowSettingsSection() {
                       to={`/backup-wallet?groupId=${group.id}&backupKind=verify`}
                       size={32}
                       style={{ paddingLeft: 16, paddingRight: 16 }}
+                      onClick={openInTabView}
```
zerts, line 79: Right now this method doesn't work well in Arc :(
It opens new empty tab that required manual reload.
It can be fixed the same way as `openInFullScreen` works

zerts:
Let's open new tab after clicking this button as well
image

zerts (approved):

src/ui/pages/BackupWallet/BackupSettingsItem/BackupSettingsItem.tsx
everdimension: Thanks!

zerts (approved):

PR #486, opened 2024-05-15
Allow longer network symbol names

vyorkin:
Also successful case dialog looks good after this change
image
image
image


zerts:
Not a blocker. But can be fixed in the future 
Zerts Screenshot 2024-05-15 at 14 38@2x

zerts (approved):

everdimension (approved):

PR #487, opened 2024-05-15
Fix position filtering and don't count not displayable assets

zerts:


src/ui/pages/Overview/Positions/Positions.tsx
```diff
@@ -645,11 +637,15 @@ function MultiChainPositions({
...
   const items = useMemo(
     () =>
-      chainValue === NetworkSelectValue.All || !positions
-        ? positions
-        : positions.filter((position) => position.chain === chainValue),
+      positions?.filter(
+        (position) =>
+          (chainValue === NetworkSelectValue.All ||
+            position.chain === chainValue) &&
+          (position.type === 'asset' ? position.is_displayable : true)
```
everdimension, line 644: move this condition first so that we compare booleans before comparing strings :)

everdimension (approved):

PR #490, opened 2024-05-16
Remove chain from permissions when deleting network

vyorkin:
Previously we reset `Permission.chain` to Ethereum. 
This PR adds a new function `removeChainWithoutAddressesFromPermissions` which deletes the chain only if there is no associated addresses. For those permissions that still have associated addresses the chain is reset to Ethereum.

### DIFF
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -1204,13 +1204,15 @@ export class Wallet {
   }: WalletMethodParams<{ chain: string }>) {
     this.ensureRecord(this.record);
 
-    this.record = Model.removeChainFromPermissions(this.record, {
-      chain: createChain(chainStr),
+    const chain = createChain(chainStr);
+    const affectedPermissions = Model.getPermissionsByChain(this.record, {
+      chain,
     });
     affectedPermissions.forEach(({ origin }) => {
-      // TODO: remove chain for origin in case new chain is not set
       this.setChainForOrigin(createChain(NetworkId.Ethereum), origin);
     });
+    this.record = Model.removeChainFromPermissions(this.record, { chain });
+
     this.verifyOverviewChain();
     this.resetEthereumChain({ context, params: { chain: chainStr } });
   }
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -1205,13 +1205,16 @@ export class Wallet {
     this.ensureRecord(this.record);
 
     const chain = createChain(chainStr);
+    this.record = Model.removeChainWithoutAddressesFromPermissions(
+      this.record,
+      { chain }
+    );
     const affectedPermissions = Model.getPermissionsByChain(this.record, {
       chain,
     });
     affectedPermissions.forEach(({ origin }) => {
       this.setChainForOrigin(createChain(NetworkId.Ethereum), origin);
     });
-    this.record = Model.removeChainFromPermissions(this.record, { chain });
 
     this.verifyOverviewChain();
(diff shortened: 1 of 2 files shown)


src/background/Wallet/Wallet.ts
```diff
@@ -1143,13 +1143,11 @@ export class Wallet {
     params: { chain: chainStr },
   }: WalletMethodParams<{ chain: string }>) {
     this.ensureRecord(this.record);
-    const affectedPermissions = Model.getPermissionsByChain(this.record, {
```
zerts, line 1146: `getPermissionsByChain` will not be used anymore

src/background/Wallet/WalletRecord.ts
```diff
@@ -727,6 +727,19 @@ export class WalletRecordModel {
...
   }
 
+  static removeChainFromPermissions(
+    record: WalletRecord,
+    { chain }: { chain: Chain }
+  ) {
+    return produce(record, (draft) => {
+      draft.permissions = Object.fromEntries(
+        Object.entries(record.permissions)
+          .filter(([, permission]) => permission.chain === chain.toString())
```
zerts, line 737: As I understand, this method will remove all permissions with other chains completely.
Cause we need to keep all permissions for other chains + change filtered permissions and remove chain for them

vyorkin: thanks for catching this, what a shame %)

zerts (approved):

src/background/Wallet/WalletRecord.ts
```diff
@@ -718,13 +718,21 @@ export class WalletRecordModel {
...
+      draft.permissions = Object.fromEntries(
+        Object.entries(record.permissions).map(([origin, permission]) => {
+          if (permission.chain === chain.toString()) {
+            return [origin, { addresses: permission.addresses }];
+          } else {
+            return [origin, permission];
+          }
+        })
+      );
+    });
```
everdimension, line 735: The idea of `immer` is that you can mutate where it's convenient:
```suggestion
    return produce(record, (draft) => {
      for (const permission of Object.values(draft.permissions)) {
        if (permission.chain === chain.toString()) {
          delete permission.chain;
        }
      }
    });
```

vyorkin: Nice, I need to get used to that, thanks!

src/background/Wallet/Wallet.ts
```diff
@@ -1143,13 +1143,11 @@ export class Wallet {
...
+
+    this.record = Model.removeChainFromPermissions(this.record, {
       chain: createChain(chainStr),
     });
-    affectedPermissions.forEach(({ origin }) => {
-      // TODO: remove chain for origin in case new chain is not set
-      this.setChainForOrigin(createChain(NetworkId.Ethereum), origin);
```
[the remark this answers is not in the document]
vyorkin: Ah, you're right. Makes sense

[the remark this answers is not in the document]
vyorkin: ✅ I have updated this logic to only delete chain from permissions if there is no associated addresses

src/background/Wallet/Wallet.ts
```diff
@@ -1143,13 +1143,19 @@ export class Wallet {
     params: { chain: chainStr },
   }: WalletMethodParams<{ chain: string }>) {
     this.ensureRecord(this.record);
+
+    const chain = createChain(chainStr);
+    this.record = Model.removeChainWithoutAddressesFromPermissions(
```
everdimension, line 1148: Why "without addresses?"

PR #491, opened 2024-05-16
Refactoring: remove extra call to verifyOverviewChain

vyorkin:
- Removes extra call to `verifyOverviewChain`, since `resetEthereumChain` calls at as well
- Renames `resetEthereumChain` to `removeEthereumChainConfig`

### DIFF
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -1150,10 +1150,7 @@ export class Wallet {
       // TODO: remove chain for origin in case new chain is not set
       this.setChainForOrigin(createChain(NetworkId.Ethereum), origin);
     });
-    this.removeEthereumChainConfig({
-      context,
-      params: { chain: chainStr },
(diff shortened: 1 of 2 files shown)


src/background/Wallet/Wallet.ts
```diff
@@ -1182,7 +1184,7 @@ export class Wallet {
     return result;
   }
 
-  async resetEthereumChain({
+  async removeEthereumChainConfig({
```
everdimension, line 1187: I agree that these method names are confusing, but not sure that this is the best name. We still call `chainConfigStore.removeEthereumChain` inside

vyorkin: Right, after taking another look, I can see that renaming didn't improve readability...

vyorkin: ✅  Reverted

everdimension (approved):

PR #492, opened 2024-05-16
Update the icon placeholder for custom network

vyorkin:
Reported issue:
> Currently, no consistency, sometimes letters, sometimes numbers. 

It also not consistent with mobile platforms.
Let’s display the first three letters of their name? Like on iOS (and Android).


zerts (approved):

src/ui/components/NetworkIcon/NetworkIcon.tsx
```diff
@@ -43,25 +35,22 @@ function TextFallback({
         ...style,
       }}
     >
-      {(!value || value.length > 4) && name
-        ? name.charAt(0).toUpperCase()
-        : value || '?'}
+      {name ? name.slice(0, 3).toUpperCase() : '???'}
```
everdimension, line 38: isn't chainId a better fallback than `???`?

everdimension (changes requested):
https://github.com/zeriontech/zerion-wallet-extension/pull/492#discussion_r1605109916

src/ui/components/NetworkIcon/NetworkIcon.tsx
vyorkin: Both fallbacks are pretty much useless in my opinion

everdimension: `chainId` is a better fallback because two chains with missing names will at least be displayed differently
Adding `title` if the name is missing is a great idea, let's do it

zerts: But it will require to move all code back :(

zerts: Also it looks like this fallback can work only if there is no network at all

zerts: So in this case ??? is even better. 
`name` is not nullable field of `NetworkConfig`

vyorkin: Looks like it is almost impossible to have a network with an empty name. But even in that rarest (probably impossible?) case, showing the fallback in the form `???` won't be too bad

everdimension (approved):

everdimension:
Great!

PR #494, opened 2024-05-17
Clear local transactions that are known by backend

everdimension:


src/background/transactions/TransactionService.ts
```diff
@@ -52,9 +67,88 @@ export class TransactionService {
...
+    address,
+    fromNonce,
+    chainId,
+  }: {
+    address: string;
+    fromNonce: number;
+    chainId: ChainId;
+  }) {
+    const transactions = await this.transactionsStore.getSavedState();
+    const candidates = transactions.filter((item) => {
```
zerts, line 108: This is indeed an optimisation, but it looks like we can pass all transactions to `bulkDeleteTransactionsByHash` without significant difference in speed

src/background/transactions/TransactionService.ts
```diff
@@ -52,9 +67,88 @@ export class TransactionService {
...
+    for (const item of transactions) {
+      const chainId = normalizeChainId(item.transaction.chainId);
+      const key = `${item.transaction.from}:${chainId}` as const;
+      map.set(key, { hash: item.hash, timestamp: item.timestamp });
+    }
+
+    for (const [key, { hash, timestamp }] of map.entries()) {
+      const [address, chainIdStr] = key.split(':');
+      const chainId = chainIdStr as ChainId;
+      const network = await getNetworkByChainId(chainId);
```
zerts, line 137: Why not using networksStore here?

zerts (approved):

zerts:
Just some questions here. Nothing to block the merge

src/background/transactions/TransactionService.ts
```diff
@@ -52,9 +67,88 @@ export class TransactionService {
...
+    const transactions = this.transactionsStore.getState();
+    const pending = getPendingTransactions(transactions);
     this.transactionsPoller.add(pending.map(toPollingObj));
     this.addListeners();
+    if (transactions.length) {
+      this.startPurgeInterval({ leading: true });
+    }
+  }
+
+  private startPurgeInterval({ leading } = { leading: false }) {
```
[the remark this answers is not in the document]
everdimension: Thx, I moved private methods before public ones

src/background/transactions/TransactionService.ts
zerts: Incorrect comment

src/background/transactions/TransactionService.ts
zerts: Ok then. This will just remove an extra call for a chain info probably

everdimension: But it seems that we want fresh data here

src/background/transactions/TransactionService.ts
```diff
@@ -87,6 +181,8 @@ export class TransactionService {
             draft.push(newItem);
           })
         );
+        this.startPurgeInterval();
```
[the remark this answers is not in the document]
everdimension: Yeah, that will effectively behave the same. I guess it just feels nice not to run code and schedule an interval when it's known ahead of time that it's not needed

src/background/transactions/TransactionService.ts
```diff
@@ -41,20 +57,126 @@ export class TransactionService {
   private transactionsStore: TransactionsStore;
   private transactionsPoller: TransactionsPoller;
 
+  static ALARM_NAME = 'TransactionService:performPurgeCheck';
```
zerts, line 60: May be it is better to name this `ALARM_NAME` more explicit?

src/background/transactions/TransactionService.ts
```diff
@@ -41,20 +57,126 @@ export class TransactionService {
...
+      const key = `${item.transaction.from}:${chainId}` as const;
+      map.set(key, { hash: item.hash, timestamp: item.timestamp });
+    }
+
+    for (const [key, { hash, timestamp }] of map.entries()) {
+      const [address, chainIdStr] = key.split(':');
+      const chainId = chainIdStr as ChainId;
+      const network = await getNetworkByChainId(chainId);
+      if (network?.supports_actions) {
+        const knownNonce = await getLatestNonceKnownByBackend({
```
zerts, line 167: Will we make a call for all actions one by one here?
Should we somehow exclude removed hashes from the `map`?

zerts (approved):

src/background/transactions/TransactionService.ts
everdimension: More explicit than `'TransactionService:performPurgeCheck'`? :)

abalyasnikov:
qa verified

PR #495, opened 2024-05-17
Add properties to signed_transaction

myriamchah:
### Context 

Add properties to Metabase event signed_transaction

### Specs 

[WLT-4699](https://zerion-tech.atlassian.net/browse/WLT-4699) part 4

[WLT-4699]: https://zerion-tech.atlassian.net/browse/WLT-4699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/analytics/shared/addressActionToAnalytics.ts
+++ b/src/shared/analytics/shared/addressActionToAnalytics.ts
@@ -21,7 +21,6 @@ interface AnalyticsTransactionData {
   zerion_fee_percentage?: number;
   zerion_fee_usd_amount?: number;
   output_chain?: string;
-  transaction_zerion_fee_asset_id?: string;
 }
 
 interface AssetQuantity {
@@ -113,7 +112,6 @@ export function addressActionToAnalytics({
     const feeAmount = quote.protocol_fee_amount;
     const asset = incoming?.[0]?.asset;
     const output_chain = quote.output_chain;
-    const transaction_zerion_fee_asset_id = getFungibleAsset(asset)?.id;
     const zerion_fee_usd_amount =
       feeAmount && asset
         ? assetQuantityToValue({ quantity: feeAmount, asset }, chain)
@@ -124,7 +122,6 @@ export function addressActionToAnalytics({
       zerion_fee_percentage,
       zerion_fee_usd_amount,
       output_chain,
-      transaction_zerion_fee_asset_id,
     };
   } else {
     return value;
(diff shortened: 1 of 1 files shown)


zerts (approved):

src/shared/analytics/shared/addressActionToAnalytics.ts
```diff
@@ -68,6 +70,10 @@ function getAssetName({ asset }: { asset: ActionAsset }) {
   return getFungibleAsset(asset)?.name;
 }
 
+function getAssetAddresses({ asset }: { asset: ActionAsset }) {
```
everdimension, line 73: why "addresses" if it returns just one?

src/shared/analytics/shared/addressActionToAnalytics.ts
```diff
@@ -99,17 +105,27 @@ export function addressActionToAnalytics({
...
     asset_name_received: toMaybeArr(incoming?.map(getAssetName)),
+    asset_address_sent: toMaybeArr(outgoing?.map(getAssetAddresses)),
+    asset_address_received: toMaybeArr(incoming?.map(getAssetAddresses)),
   };
   if (quote) {
     const zerion_fee_percentage = quote.protocol_fee;
     const feeAmount = quote.protocol_fee_amount;
     const asset = incoming?.[0]?.asset;
+    const output_chain = quote.output_chain;
+    const transaction_zerion_fee_asset_id = getFungibleAsset(asset)?.id;
```
everdimension, line 116: I think the name implies that this should be id of asset that the fee is paid with, but we assign the id of the incoming asset instead?

everdimension (changes requested):
See comments

src/shared/analytics/shared/addressActionToAnalytics.ts
myriamchah: Oops right! Got misled 👀

src/shared/analytics/shared/addressActionToAnalytics.ts
myriamchah: In what case can incoming asset and zerion fee asset be different?

myriamchah: The field is not required in the specs anymore (I validated with Anna) so I'll remove it rn, thank you!

everdimension (approved):

PR #497, opened 2024-05-17
Improve displaying symbol names

vyorkin:
Get rid of `toLowerCase()` & `toUpperCase()` for symbols. Adjust CSS rules for long symbol names.


zerts:
Zerts Screenshot 2024-05-17 at 16 35@2x

zerts:
Everything looks fine except the bug with overflow state. May be this is not a regression, but would be nice to be fixed as well

vyorkin:
I was 100% sure I checked this view as well... Thanks!

everdimension (approved):

PR #500, opened 2024-05-17
Add options to PersistentStore to support migrations

everdimension:
Fix ChainConfigStore initialization

### DIFF
--- a/src/modules/persistent-store.ts
+++ b/src/modules/persistent-store.ts
@@ -1,25 +1,37 @@
 import { Store } from 'store-unit';
 import * as browserStorage from 'src/background/webapis/storage';
 
-export class PersistentStore<T> extends Store<T> {
-  protected key: string;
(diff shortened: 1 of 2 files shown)


src/modules/persistent-store.ts
```diff
@@ -1,25 +1,37 @@
...
+  save: (key: string, value: S) => Promise<void>;
+};
+
 export class PersistentStore<T> extends Store<T> {
+  static async readSavedState<T>(key: string) {
+    return browserStorage.get<T>(key);
+  }
+
   protected key: string;
   protected isReady: boolean;
```
zerts, line 15: It looks like we can make these fields private again

zerts (approved):

src/modules/persistent-store.ts
everdimension: Yep

PR #502, opened 2024-05-21
Fix chain config migration

zerts:


src/modules/ethereum/chains/versions.ts
```diff
@@ -6,7 +6,7 @@ import { toCustomNetworkId } from './helpers';
 type PossibleEntry = ChainConfigV1 | ChainConfig;
 
 function maybeLocalChainId(id?: string | null) {
-  return id?.length === 21; // nanoid() standard length
+  return !id || id.length === 21; // nanoid() standard length
```
everdimension, line 9: I suppose this fix will only help those who have not migrated yet?

everdimension (approved):

PR #503, opened 2024-05-21
Setup opt-in persistence for QueryClient

everdimension:


zerts (approved):

src/background/resource/sessionCacheService.ts
```diff
@@ -1,11 +1,17 @@
+type Entry = { timestamp: number; value: unknown };
+
 export class SessionCacheService {
-  private cache: Record<string, { timestamp: number; value: unknown }> = {};
+  private cache: Record<string, Entry | undefined> = {};
```
vyorkin, line 4: Do we really need to store `undefined` values? It type checks successfully without `| undefined` btw

everdimension: That's the problem :) typescript assumes the value will always be retrieved by any key

everdimension: And we're not allowing to pass undefined when setting a value btw

vyorkin: ah, alright, makes sense!

vyorkin (approved):

PR #504, opened 2024-05-21
Display native balance for custom networks in Network Select

everdimension:


src/ui/shared/requests/useNativeBalance.ts
```diff
@@ -75,13 +82,13 @@ export function useNativeBalance({
     const position =
       nativeAddressPosition.data || evmNativeAddressPosition.data;
     if (!position?.quantity) {
-      return { data: null, isLoading };
+      return { data: { value: null, position }, isLoading };
```
zerts, line 85: Should there be valueCommon?

zerts (approved):

src/ui/shared/requests/useNativeBalance.ts
everdimension: Oh! Good catch

zerts (approved):

vyorkin (approved):

PR #513, opened 2024-05-24
Refactor: minor changes for GlobalPreferences

vyorkin:
* move removing of expired provider injections into `GlobalPreferences`
* move `removeExpiredRecords()` inside `activate()`

It feels like the removing of expired provider injections belongs to `GlobalPreferences` class, wdyt?

### DIFF
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -123,7 +123,7 @@ initialize().then((values) => {
   // const httpConnection = new HttpConnection(() => account.getCurrentWallet());
   const memoryCacheRPC = new MemoryCacheRPC();
 
-  new ContentScriptManager().removeExpiredRecords().activate();
+  new ContentScriptManager().activate();
 
   portRegistry.addMessageHandler(
     createWalletMessageHandler(() => account.getCurrentWallet())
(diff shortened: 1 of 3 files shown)


src/background/index.ts
```diff
@@ -123,7 +123,7 @@ initialize().then((values) => {
   // const httpConnection = new HttpConnection(() => account.getCurrentWallet());
   const memoryCacheRPC = new MemoryCacheRPC();
 
-  new ContentScriptManager().removeExpiredRecords().activate();
```
vyorkin, line 126: But maybe the intent of keeping this call here was to bring an extra attention to it. To me it just looks natural to keep it in the `activate()`

everdimension:
I think this is mixing of concerns.
Currently, `GlobalPreferences` is unaware of the logic behind stored values.

This update seems to introduce coupling.

I use this rule of thumb: if you remove a feature, what code is affected?

Currently, if we remove `ContentScriptManager`, then only its file and its instantiation needs to be removed.

With this update, you will also need to remove a method from `GlobalPreferences` because it happened to be coupled to the "pause feature" logic

vyorkin:
Fair enough, thanks for clarification :)

PR #519, opened 2024-05-30
Improve switch chain error message

vyorkin:
It might be better to display a similar message for this error as we do for error 4100:
“Unrecognized chain ID. Try adding the chain using wallet_addEthereumChain first.”

This is consistent with MetaMask's error message: [MetaMask Docs](https://docs.metamask.io/wallet/reference/wallet_switchethereumchain/).

### DIFF
--- a/src/shared/errors/errors.ts
+++ b/src/shared/errors/errors.ts
@@ -86,7 +86,7 @@ export class SwitchChainError extends ErrorWithEnumerableMessage {
   code = 4902;
 
   constructor(
-    message = 'Unrecognized chain ID: Try adding the chain using wallet_addEthereumChain first.'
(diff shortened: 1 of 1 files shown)


src/shared/errors/errors.ts
```diff
@@ -85,7 +85,9 @@ export class OriginNotAllowed extends ErrorWithEnumerableMessage {
 export class SwitchChainError extends ErrorWithEnumerableMessage {
   code = 4902;
 
-  constructor(message = 'Chain not configured') {
+  constructor(
+    message = 'Unrecognized chain ID: Try adding the chain using wallet_addEthereumChain first.'
```
everdimension, line 89:
```suggestion
    message = 'Unrecognized chainId: Try adding the chain using wallet_addEthereumChain first.'
```

everdimension (approved):

PR #520, opened 2024-05-31
Unblock history render

zerts:


src/ui/pages/History/ActionItem/ActionItem.tsx
```diff
@@ -75,7 +81,9 @@ function ActionTitle({
 
   return (
     <UIText kind="body/accent">
-      {explorerUrl ? (
+      {!actionTitle ? (
+        <DelayedRender delay={1000}>[Missing network data]</DelayedRender>
```
everdimension, line 85: should there be a placeholder so that the element height doesn't change after `delay`?

zerts:
Too old pr

PR #523, opened 2024-06-03
Refactor: window and layout context params

everdimension:


vyorkin (approved):
Looks cleaner, didn't find any issues

PR #524, opened 2024-06-03
Set Esc shortcut globally for dialogs

everdimension:
Refactor some initialization code

PR #526, opened 2024-06-05
Remove Background in favour of useBackgroundKind hook

vyorkin:
Might be better to merge after #524 to reduce possible conflicts

- [ ] Test that updated components look the same as before


vyorkin:
Alright, now I think it wasn't a good idea to update all at once. That leads to endless rebase-hell, lets do it gradually instead

PR #532, opened 2024-06-08
Build 

OKEAMAH:


everdimension:
Hello!
We're using [svgo](https://github.com/zeriontech/zerion-wallet-extension/blob/408099abcfbcf3ff526b35134273c355c8d96b4c/svgo.config.js) to process `svg` assets
Besides, these newlines do not matter much in the extension context

PR #533, opened 2024-06-10
Fix issues with mixpanel events

zerts:


src/shared/analytics/mixpanel.ts
```diff
@@ -33,6 +34,19 @@ class DeviceIdStore extends PersistentStore<string | undefined> {
   }
 }
 
+function omitNullParams<T extends Record<string, unknown>>(
+  params: T
+): Partial<T> {
+  return produce(params, (draft) => {
+    const keys = Object.keys(draft);
+    for (const key of keys) {
+      if (draft[key] == null) {
```
everdimension, line 43: what about empty arrays?

zerts: We discussed only `name: null` cases as bad

zerts: This is because if Mixpanel's limitations

vyorkin (approved):

src/shared/analytics/mixpanel.ts
```diff
@@ -33,6 +34,19 @@ class DeviceIdStore extends PersistentStore<string | undefined> {
   }
 }
 
+function omitNullParams<T extends Record<string, unknown>>(
+  params: T
+): Partial<T> {
+  return produce(params, (draft) => {
+    const keys = Object.keys(draft);
+    for (const key of keys) {
```
everdimension, line 42: White not `for (const key in draft)`?

zerts: No reason, just forgot this syntax

src/shared/analytics/mixpanel.ts
```diff
@@ -155,9 +154,9 @@ class MixpanelApi {
       url.searchParams.append('verbose', '1');
     }
     url.searchParams.append('ip', '1');
-    const payload = omitNullParams({
+    const payload = {
       event,
-      properties: {
+      properties: omitNullParams({
```
everdimension, line 159: 👍

everdimension (approved):

PR #534, opened 2024-06-10
Fix user_id param for mixpanel

zerts:


src/shared/analytics/mixpanel.ts
```diff
@@ -244,7 +244,8 @@ export async function mixpanelTrack(
   const userId = account.getUser()?.id;
   mixpanelApi.track(event, {
     ...values,
-    ...(userId ? { user_id: userId } : null),
+    // pass userId params cause identify function can still be in progress
```
everdimension, line 247: ...or not yet invoked?

everdimension (approved):

PR #538, opened 2024-06-11
Remove extra dapp connection event

zerts:


src/shared/analytics/analytics.background.ts
```diff
@@ -47,7 +47,10 @@ function trackAppEvents({ account }: { account: Account }) {
     const getUserId = () => account.getUser()?.id;
     return createBaseParams({ ...params, userId: getUserId() });
   };
-  emitter.on('dappConnection', ({ origin, address }) => {
+  emitter.on('requestAccounts', ({ origin, address, explicitly }) => {
```
everdimension, line 50:
```suggestion
  /** {explicitly: true} means that user confirmed connection in a dialog. {false} means that we resolve a previously approved addess value */
  emitter.on('requestAccounts', ({ origin, address, explicitly }) => {
```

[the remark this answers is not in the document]
zerts: Well, this looks fine enough to use :)

vyorkin (approved):

src/shared/analytics/analytics.background.ts
```diff
@@ -47,7 +47,11 @@ function trackAppEvents({ account }: { account: Account }) {
     const getUserId = () => account.getUser()?.id;
     return createBaseParams({ ...params, userId: getUserId() });
   };
-  emitter.on('dappConnection', ({ origin, address }) => {
+  /** {explicitly: true} means that user confirmed connection in a dialog. {false} means that we resolve a previously approved addess value */
```
everdimension, line 50: This comment should be placed where the event is described: https://github.com/zeriontech/zerion-wallet-extension/pull/538/files#diff-f4612f4ef92560f56994174a746d1bdaaa3a0244832c843950cb7a3105844a50R48

everdimension (approved):

PR #542, opened 2024-06-11
Fix insufficient funds warning

everdimension:
Load gas prices in suspense phase
Correctly handle empty response

### DIFF
--- a/src/ui/shared/requests/useNativeBalance.ts
+++ b/src/ui/shared/requests/useNativeBalance.ts
@@ -1,9 +1,9 @@
 import { useMemo } from 'react';
+import { type AddressPosition, useAddressPositions } from 'defi-sdk';
 import type { Chain } from 'src/modules/networks/Chain';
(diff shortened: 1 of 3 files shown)


src/ui/shared/requests/useNativeBalance.ts
```diff
@@ -38,11 +38,13 @@ function useNativeAddressPosition({
           !item.protocol
       ) ?? [];
     if (nativePositions.length > 1) {
+      // eslint-disable-next-line no-console
       console.warn('multiple native positions');
     }
     return {
       data: nativePositions[0],
       isLoading,
+      isSuccess: Boolean(value?.positions),
```
zerts, line 47: Hm, what about isDone flag? Can it be used here?

zerts (approved):

src/ui/shared/requests/useNativeBalance.ts
everdimension: I think `isSuccess` more familiar because of `react-query`

PR #556, opened 2024-06-17
Feature/testnet mode

everdimension:
Test plan:
- Testmode ON:
  - [ ] Pending transaction polls rpc-url from test config
  - [ ] Submitted tx send correct chain value to analytics
  - [ ] RPC requests from Dapps use correct rpc-url
  - [ ] Dapp initiates mainnet transaction 
  - [ ] ...?

TODO:
- [x] Update to new flag in settings where we show "missing testnets" message
- [x] Use testnet endpoint for gas prices api
- [ ] In testnet mode, when dapp is connected to a mainnet
  - [ ] Display connected network correctly
  - [ ] What to display for positions?
- [ ] NFT tab / Perks tab
- [x] SendForm: what if testnet mode is disabled during confirmation dialog?


.github/workflows/pr.yml
```diff
@@ -55,9 +63,11 @@ jobs:
       - name: Install dependencies and build
         env:
           DEFI_SDK_API_URL: ${{ github.event.inputs.DEFI_SDK_API_URL || 'wss://api-v4.zerion.io/' }}
+          DEFI_SDK_TESTNET_API_URL: ${{ github.event.inputs.DEFI_SDK_TESTNET_API_URL || 'wss://api-testnet.zerion.io/' }}
           ZERION_API_URL: ${{ github.event.inputs.ZERION_API_URL || 'https://zpi.zerion.io/' }}
+          ZERION_TESTNET_API_URL: ${{ github.event.inputs.ZERION_TESTNET_API_URL || 'https://zpi-testnet.zerion.io/' }}
           BACKEND_ENV: ${{ github.event.inputs.BACKEND_ENV || '' }}
-          FEATURE_PAYMASTER: ${{ github.event.inputs.FEATURE_PAYMASTER || '' }}
+          FEATURE_PAYMASTER: ${{ github.event.inputs.FEATURE_PAYMASTER || 'on' }}
```
zerts, line 70: Looks like there should be `FEATURE_PAYMASTER ` in the inputs section in the beginning of the config

.github/workflows/release.yml
```diff
@@ -22,14 +22,17 @@ jobs:
...
           ZERION_API_URL: https://zpi.zerion.io/
+          ZERION_API_URL: https://zpi-testnet.zerion.io/
           PROXY_URL: https://proxy.zerion.io/
           DEFI_SDK_TRANSACTIONS_API_URL: https://transactions.zerion.io
           DEFI_SDK_API_TOKEN: Zerion.0JOY6zZTTw6yl5Cvz9sdmXc7d5AhzVMG
           SOCIAL_API_URL: https://social.zerion.io/
           TEST_WALLET_ADDRESS: ${{ secrets.TEST_WALLET_ADDRESS }}
           FEATURE_SEND_FORM: on
           FEATURE_FOOTER_BUG_BUTTON: off
+          FEATURE_PAYMASTER: off
```
zerts, line 35: hm, should this feature be 'on' for testnet release?

.github/workflows/pr.yml
everdimension: I think it's there, right? https://github.com/zeriontech/zerion-wallet-extension/blob/main/.github/workflows/pr.yml#L12-L15

.github/workflows/release.yml
everdimension: I don't know yet, we'll see :)

src/modules/networks/useNetworks.ts
```diff
@@ -1,54 +1,90 @@
...
+
+export function useNetworks(chains?: string[]) {
+  const networksStore = useNetworksStore();
   const { data: networks = null, ...query } = useQuery({
-    queryKey: ['loadNetworks', chainIds],
-    queryFn: () => networksStore.load(chainIds),
-    refetchOnWindowFocus: false,
-    refetchOnMount: false,
+    queryKey: ['loadNetworks', chains, networksStore],
+    queryKeyHashFn: (queryKey) => {
```
zerts, line 29: Isn't this a very long operation for queryKey?

src/modules/networks/useNetworks.ts
```diff
@@ -1,54 +1,90 @@
...
+} from 'src/modules/networks/networks-store.client';
+import { usePreferences } from 'src/ui/features/preferences';
+import { invariant } from 'src/shared/invariant';
 import { getNetworksBySearch } from '../ethereum/chains/requests';
 import type { ChainId } from '../ethereum/transactions/ChainId';
+import { NetworksStore } from './networks-store';
 
-export function useNetworks(chainIds?: string[]) {
+function useNetworksStore() {
+  const { preferences } = usePreferences();
```
zerts, line 17: Not sure how critical this is, but looks like this hook can just return boolean that shows which store should be used. And in the useNetworks we will use needed store. This can optimise dependency compare logic

src/ui/pages/Networks/Networks.tsx
```diff
@@ -222,23 +227,23 @@ function NetworkPage() {
   const mutation = useMutation({
     mutationFn: saveChainConfig,
     onSuccess() {
-      networksStore.update();
+      updateNetworks();
```
zerts, line 230: should we add `await` here?

src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
```diff
@@ -48,14 +57,20 @@ export function TransactionSimulation({
...
 
+  const client = useDefiSdkClient();
   const { data: localAddressAction } = useQuery({
     queryKey: [
       'incomingTxToIncomingAddressAction',
       transaction,
       transactionAction,
       networks,
       address,
+      client,
```
zerts, line 68: How can client be serialised here?

src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
```diff
@@ -93,6 +110,37 @@ export function TransactionSimulation({
...
+  const paymasterTxInterpretQuery = useQuery({
+    enabled: paymasterEligible,
+    suspense: false,
+    queryKey: ['interpret/typedData', client, transaction],
+    queryKeyHashFn: (queryKey) => {
+      const key = queryKey.map((x) => (x instanceof Client ? x.url : x));
+      return hashQueryKey(key);
+    },
+    queryFn: async () => {
+      console.log('paymaster simulation query');
```
zerts, line 122: Do not forget to remove console logs here

src/background/Wallet/WalletRecord.ts
```diff
@@ -772,6 +771,7 @@ export class WalletRecordModel {
       upgradeDnaBannerDismissed: false,
       backupReminderDismissedTime: 0,
       enableTestnets: false,
+      testnetMode: null,
```
zerts, line 774: What about moving this to globalPreferences? This param can be easier to read from globally available class

src/ui/pages/History/History.tsx
```diff
@@ -120,6 +123,7 @@ function useMinedAndPendingAddressActions({
       actions_search_query: searchQuery,
     },
     {
+      client: useDefiSdkClient(),
```
zerts, line 126: Do we need this if we have global Provider?

src/background/Wallet/Wallet.ts
```diff
@@ -752,6 +753,15 @@ export class Wallet {
     return Model.getPreferences(this.record);
   }
 
+  private async updateChainConfigStore() {
+    const preferences = await this.getPreferences({
+      context: INTERNAL_SYMBOL_CONTEXT,
+    });
+    const on = Boolean(preferences.testnetMode?.on);
+    const client = getDefiSdkClient({ on });
```
zerts, line 761: May be it is better to use more explicit name here?

src/modules/networks/networks-store.background.ts
```diff
@@ -1,16 +1,99 @@
...
+}
+
+interface FetchNetworkOptions {
+  preferences: Pick<WalletRecord['publicPreferences'], 'testnetMode'>;
+  /**
+   * 'current' chooses networks store based on preferences.testnetMode value
+   * 'mainnet' uses mainnet networks store
+   * 'both' will try testnet networks store first IF {preferences.testnetMode.on}, if no network is found, it will try mainNetworksStore
+   */
+  apiEnv: 'mainnet' | 'both' | 'current';
```
zerts, line 47: Maybe both can be changed to preferTestnet

src/modules/networks/networks-store.background.ts
```diff
@@ -1,16 +1,99 @@
...
+    client,
+    testnetMode: false,
   }
 );
 
+export const testenvNetworksStore = new NetworksStore(
+  { networks: null },
+  {
+    getEthereumChainConfigs: async () => {
+      await chainConfigStore.ready();
```
zerts, line 26: Do we need to set client for chainConfigStore here to testnet?

src/background/Wallet/WalletRecord.ts
everdimension: Why? Unless the value should be read when you're logged out, it doesn't belong to globalPreferences store

src/ui/pages/History/History.tsx
everdimension: That's right, not anymore. I'll remove

src/modules/networks/networks-store.background.ts
everdimension: I think this name would have the same problem as "both": it may seem that testnet api will be always queried.
Other ideas? :)

src/modules/networks/networks-store.background.ts
everdimension: Not really since they share same config store, but thanks for pointing out still, I just realized I'm not setting it initially in `Wallet`

src/modules/networks/networks-store.background.ts
zerts: testnetFirst ?
Like cache-first in defi-sdk. Doesn't guarantee that cache exists

src/background/Wallet/WalletRecord.ts
zerts: Ok

src/background/messaging/HttpConnection.ts
```diff
@@ -9,31 +9,22 @@ import {
   formatJsonRpcError,
   isJsonRpcRequest,
 } from '@walletconnect/jsonrpc-utils';
-import { networksStore } from 'src/modules/networks/networks-store.background';
-import type { ChainId } from 'src/modules/ethereum/transactions/ChainId';
 
 export class HttpConnection extends EventEmitter {
-  private chainId: ChainId;
+  private url: string;
```
zerts, line 14: NICE!

src/background/Wallet/Wallet.ts
```diff
@@ -1335,6 +1377,106 @@ export class Wallet {
       id,
     }) as Promise<string>;
   }
+
+  private async checkTestnetMode(
+    chainId: ChainId
+  ): Promise<
+    | { violation: true; network: NetworkConfig; mode: 'testnet' | 'default' }
```
zerts, line 1384: Can we define `mode` in this type as just `testnet`?

src/ui/style/global.module.css
```diff
@@ -40,7 +40,8 @@ body {
   --body-width: 425px;
   --body-height: 600px;
   width: var(--body-width);
-  height: var(--body-height);
+  /** TODO: check that using min-height instead of height doesn't break popup dimensions in supported browsers */
```
vyorkin, line 43: In Brave & Arc it looks the same as in Chrome

src/background/Wallet/Wallet.ts
```diff
@@ -1335,6 +1377,106 @@ export class Wallet {
...
+    }
+    const network = await fetchNetworkByChainId({
+      preferences,
+      chainId,
+      apiEnv: 'both',
+    });
+    if (network && !network.is_testnet) {
+      return { violation: true, network, mode: 'testnet' };
+    } else {
+      return { violation: false, network: null, mode: 'testnet' };
```
zerts, line 1401: As I see we don't use `mode` field at all. Do we need it?

src/ui/index.tsx
```diff
@@ -102,7 +104,10 @@ new BackgroundScriptUpdateHandler({
   onFailedHandshake: () => handleFailedHandshake(),
 }).keepAlive();
 
-initializeUI();
+initializeUI().then(() => {
+  // TODO: remove this event?
+  emitter.emit('uiInitialized');
```
vyorkin, line 109: Looks like it's not being used yet... but it's just an event, it's not hurting anyone :)

src/background/messaging/HttpConnection.ts
```diff
@@ -9,31 +9,22 @@ import {
   formatJsonRpcError,
   isJsonRpcRequest,
 } from '@walletconnect/jsonrpc-utils';
-import { networksStore } from 'src/modules/networks/networks-store.background';
-import type { ChainId } from 'src/modules/ethereum/transactions/ChainId';
 
 export class HttpConnection extends EventEmitter {
```
vyorkin, line 13: Looks like HttpConnection is not really living up to its potential—it's short-lived and no one subscribes to its `'payload'` events. Given that, do we really need it to be a class? If we're not leveraging inheritance features or maintaining state, we might be better off with a more straightforward approach. How about refactoring it into a simple async function like `sendJsonRpcRequest(url: string, request: JsonRpcPayload)`?

src/background/messaging/HttpConnection.ts
```diff
@@ -9,31 +9,22 @@ import {
...
-    this.chainId = chainId;
+    this.url = url;
   }
 
-  async send(
-    request: JsonRpcPayload,
-    _context: unknown
-  ): Promise<JsonRpcResult | JsonRpcError> {
+  async send(request: JsonRpcPayload): Promise<JsonRpcResult | JsonRpcError> {
     if (!isJsonRpcRequest(request)) {
```
vyorkin, line 22: Looks like this check for `isJsonRpcRequest` is done twice (not really big performance impact though :))
The function returned by the `createHttpConnectionMessageHandler` already checks that message before sending it to `HttpConnection`

src/ui/pages/Networks/Networks.tsx
everdimension: wasn't here before so 🤷‍♀️ and unless it's needed, it's preferred no to delay

src/modules/networks/networks-store.background.ts
everdimension: Kinda works, thanks

src/ui/index.tsx
everdimension: thanks!

src/background/messaging/HttpConnection.ts
everdimension: That's true. It's an artifact of an earlier implementation. I wouldn't refactor it in this PR though

src/background/messaging/HttpConnection.ts
everdimension: But we don't know that here

src/modules/defi-sdk/background.ts
```diff
@@ -26,3 +26,24 @@ export function configureBackgroundClient() {
     },
   });
 }
+
+export const configureBackgroundTestClient = memoizeOne(() => {
+  invariant(DEFI_SDK_TESTNET_API_URL, 'DEFI_SDK_TESTNET_API_URL not defined');
+  invariant(DEFI_SDK_API_TOKEN, 'DEFI_SDK_API_TOKEN not defined');
+  const client = new Client({
+    url: DEFI_SDK_TESTNET_API_URL,
+    apiToken: DEFI_SDK_API_TOKEN,
```
vyorkin, line 35: Would we want to use a separate token for the testnet api?

src/ui/features/testnet-mode/TestModeDecoration/TestModeDecoration.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { useBodyStyle } from 'src/ui/components/Background/Background';
+import { templateData } from 'src/ui/shared/getPageTemplateName';
+import { KeyboardShortcut } from 'src/ui/components/KeyboardShortcut';
+import React from 'react';
+import { HStack } from 'src/ui/ui-kit/HStack/HStack';
+import { UIText } from 'src/ui/ui-kit/UIText';
+import { Toggle } from 'src/ui/ui-kit/Toggle';
+import { usePreferences } from '../../preferences';
+
+export function TestModeDecoration() {
```
vyorkin, line 13: Not sure how big of a deal this is, but it does seem a bit odd when the extension pops up in a separate tab
image

everdimension: I see! we don't have bug report button enabled now, though

vyorkin: ah, I see, thats my .env, nvm

src/modules/defi-sdk/background.ts
everdimension: 🤷‍♀️

PR #558, opened 2024-06-19
Fix network fetch for typedSignData

zerts:

PR #559, opened 2024-06-19
Ensure networksStore has network for chain

everdimension:


everdimension:
Fixed in https://github.com/zeriontech/zerion-wallet-extension/pull/558

PR #562, opened 2024-06-23
Fix missing buttons on edit custom allowance form

vyorkin:
Added missing `RenderArea` to `SendTransaction` to restore Cancel & Save buttons on `AllowanceForm`


everdimension:
Why use render area here?

vyorkin:
> Why use render area here?

`AllowanceForm` relies on presence of the `sign-transaction-footer` area. It needs it to show the "Save" / "Reset" buttons

zerts (approved):

PR #567, opened 2024-06-28
Do no display both application and recipient at the same time

vyorkin:


src/ui/components/address-action/AddressActionDetails/AddressActionDetails.tsx
```diff
@@ -32,7 +32,9 @@ export function AddressActionDetails({
   showApplicationLine: boolean;
   singleAssetElementEnd: React.ReactNode;
 }) {
-  const applicationLineVisible = showApplicationLine && addressAction?.label;
+  const applicationLineVisible =
+    showApplicationLine && addressAction?.label && !recipientAddress;
+
   return (
     <>
       {recipientAddress && addressAction?.type.value === 'send' ? (
```
everdimension, line 40: These conditions don't match, but they should, as I understand

everdimension: Or maybe this isn't incorrect. But another idea: what if we always show network icon on RecipientLine?

everdimension (approved):

everdimension (approved):

PR #568, opened 2024-06-28
Fix condition for regular tx interpretation

everdimension:


src/ui/pages/SendForm/SendForm.tsx
```diff
@@ -594,6 +594,7 @@ function SendFormComponent() {
                     chain={chain}
                     sendView={sendView}
                     paymasterEligible={paymasterEligible}
+                    eligibilityQuery={eligibilityQuery}
```
zerts, line 597: Can we lose eligibilityQuery.data here because of spread operator above?

src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
```diff
@@ -94,8 +99,13 @@ export function TransactionSimulation({
     useErrorBoundary: true,
   });
 
+  const network = chain ? networks?.getNetworkByName(chain) : null;
   const txInterpretQuery = useQuery({
-    enabled: FEATURE_PAYMASTER_ENABLED ? paymasterEligible === false : true,
+    enabled: FEATURE_PAYMASTER_ENABLED
+      ? network?.supports_sponsored_transactions
```
zerts, line 105: This definitely looks like something that can be moved to a helper. Because we use this sophisticated condition twice :)

src/ui/pages/SendForm/SendForm.tsx
zerts: It looks like we missed interpretations for send form because of this

vyorkin (approved):

src/ui/pages/SendForm/SendForm.tsx
everdimension: good catch, thank you!

zerts (approved):

PR #569, opened 2024-06-28
Update backup flow

vyorkin:
### Test plan

- [ ] [Check success background image](https://github.com/zeriontech/zerion-wallet-extension/pull/569#discussion_r1672198318) in production build
- [ ] Check backup from wallet group
  - [ ] Double-check the backup success screen since I removed the confetti from it
- [ ] Check reveal private key
- [ ] Check all onboarding scenarios

#### Summary

* Different success screens for onboarding VS regular backups (no confetti, no pin extension for regular backup success screen, "you can close this tab now" instead of smth like "you can now get started")
* [Different headlines](https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#diff-fd5894838223bafbea05658e23bd488269c08437307eb23b2bbe5b35abaca1c3R115-R129) for when doing onboarding backup VS regular backup
* TODO

#### Old backup flow

I renamed `BackupWallet` to `RevealPrivateKey` and removed everything except the "reveal" functionality. I updated the route as well, now it is `/reveal-private-key?groupId={groupId}&address={address}`. User still needs to be verified in order to see its private key (same as before).

#### Query parameters, contexts, modes

Some of the query string parameters renamed:
* `windowContext` -> `windowType`
* `layout` -> `windowLayout`: To keep it consistent with the `windowType` parameter.
* `context` -> `appMode`: We use the term "context" frequently, so let's pick a name that avoids it. Additionally, I've added a new `UrlContext` class, and too many "contexts" could be confusing. The `appMode: AppMode` indicates that the Zerion wallet extension can operate in different modes. Currently, these modes are "wallet" (the main/default mode) and "onboarding"

`UrlContext` is a helper (decorator) class which allows us to easily check states described by the following types:
```
export type AppMode = 'onboarding' | 'wallet';
export type WindowType = 'popup' | 'tab' | 'dialog';
export type WindowLayout = 'column' | 'page';
```

#### Refactoring

- Got rid of the `mode` parameter in `App` component, simplified `initializeUI` function
- Added `getPopupUrl` to remove ~15 lines of duplicate code, simplified `getPopupRoute` function
- [Renamed](https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#diff-81eb772327ae14d98a807c91981d882a4a051c5352bdcffa18d7276aa254d729R53) `isOnboardingContext` -> `isOnboardingMode` (now we have `AppMode = 'onboarding' | 'wallet'`)
- `openOnboarding` function doesn't take the `url` parameter anymore, it just calls `getPopupUrl` to get it
- Use `set` instead of `append` to set window type/layout & context related search parameters
- Renamed `useSizeStore` -> `useWindowSizeStore`
- Moved `onboarding` into `features` directory
- Renamed `Onboarding/Create/CreatePassword.tsx → onboarding/Create/Create.tsx`

### TODO

- [ ] Would be nice to refactor [onboarding/initialization.ts](https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#diff-c2c076a3cde067bdecf771bc345553701036736217da7c2f5a7150182044d238) as well


src/ui/shared/openInTabView.ts
```diff
@@ -1,7 +1,7 @@
 import browser from 'webextension-polyfill';
 
 export async function openTabView(url: URL) {
-  url.searchParams.append('windowContext', 'tab');
+  url.searchParams.set('windowContext', 'tab');
```
vyorkin, line 4: Otherwise we would sometimes get 2 `windowContext` parameters

src/ui/ui-kit/Stack/Stack.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import { HStack } from '../ui-kit/HStack';
-import { VStack } from '../ui-kit/VStack';
+import { HStack } from 'src/ui/ui-kit/HStack';
+import { VStack } from 'src/ui/ui-kit/VStack';
 
 export const Stack = React.forwardRef<
```
vyorkin, line 5: Shared, used in a few places and looks convenient, so I extracted it and moved to the ui-kit

src/ui/pages/RevealPrivateKey/index.ts
```diff
@@ -0,0 +1 @@
+export { RevealPrivateKey } from './RevealPrivateKey';
```
vyorkin, line 1: Currently, this page is dedicated solely to revealing the private key, hence the new name

src/ui/pages/RevealPrivateKey/RevealPrivateKey.tsx
```diff
@@ -426,7 +426,7 @@ function VerifySuccess({ seedType }: { seedType: SeedType }) {
   );
 }
 
-export function BackupWallet() {
+export function RevealPrivateKey() {
```
vyorkin, line 429: This page is now exclusively for revealing the private key, which is why it has a new name

src/ui/pages/Networks/NetworkForm/NetworkForm.tsx
```diff
@@ -8,7 +8,7 @@ import { Spacer } from 'src/ui/ui-kit/Spacer';
 import { UIText } from 'src/ui/ui-kit/UIText';
 import { UnstyledButton } from 'src/ui/ui-kit/UnstyledButton';
 import { VStack } from 'src/ui/ui-kit/VStack';
-import LockIcon from 'jsx:src/ui/assets/lock.svg';
+import LockIcon from 'jsx:src/ui/assets/lock-outline.svg';
```
vyorkin, line 11: I renamed it, since we have another shared svg icon that is not "outline"

src/ui/pages/Backup/useRecoveryPhrase.ts
```diff
@@ -0,0 +1,27 @@
...
+
+async function getRecoveryPhrase(context: BackupContext) {
+  if (context.name === 'wallet') {
+    const mnemonic = await walletPort.request('getRecoveryPhrase', {
+      groupId: context.groupId,
+    });
+    return mnemonic.phrase;
+  } else {
+    return walletPort.request('getPendingRecoveryPhrase');
+  }
```
vyorkin, line 13: Depending on the context, we either obtain a pending recovery phrase or retrieve the recovery phrase by the `groupId`

src/ui/App/App.tsx
```diff
@@ -172,6 +174,17 @@ function PageLayoutViews() {
...
+
+      <Route
+        path="/backup/*"
+        element={
+          <RequireAuth>
+            <PageLayout>
+              <Backup />
+            </PageLayout>
+          </RequireAuth>
+        }
```
vyorkin, line 186: I wanted to move the usage of <PageLayout> inside the <Backup> component, but in the <Onboarding> component, all these elements are explicitly wrapped (as if it is possible for these components to be used outside of <PageLayout> as well):
```
<Route                        
  path="/onboarding"          
  element={                   
    <PageLayout>              
      <Welcome />             
    </PageLayout>             
  }                           
/>                            
<Route                        
  path="/onboarding/create/*" 
  element={                   
    <PageLayout>              
      <Create />              
    </PageLayout>             
  }                           
/>                            
```
So I decided to leave it as is

src/ui/features/onboarding/Success/styles.module.css
```diff
@@ -10,7 +10,7 @@
 }
 
 .container {
-  background-image: url('../assets/success_background.png');
+  background-image: url('~src/ui/assets/success_background.png');
```
vyorkin, line 13: Now it is shared (since we have 2 `Success` components that use this image), so I moved it to `src/ui/assets`

src/ui/features/onboarding/shared/useOnboardingSession.ts
```diff
@@ -32,6 +32,5 @@ export function useOnboardingSession(
 
   return {
     sessionDataIsLoading: isLoading,
-    hasExistingUser: Boolean(existingUser),
```
vyorkin, line 35: Not used anywhere

src/ui/pages/Backup/RecoveryPhrase.tsx
```diff
@@ -141,9 +148,11 @@ export function Backup({
...
                   Verify Backup
                 </Button>
-                <Button kind="ghost" onClick={onSkip}>
-                  Do it Later
-                </Button>
+                {context.name === 'onboarding' ? (
+                  <Button kind="ghost" onClick={onSkip}>
+                    Do it Later
+                  </Button>
+                ) : null}
```
vyorkin, line 155: We want to show "Do it Later" only during onboarding. Otherwise, it would be odd to display this button right after the user clicked "Backup," as it contradicts their current intention

src/ui/pages/Backup/RecoveryPhrase.tsx
```diff
@@ -68,8 +74,9 @@ export function Backup({
           <UIText kind="headline/h2">Let’s Back Up Your Wallet!</UIText>
           <Spacer height={8} />
           <UIText kind="body/regular">
-            Save these 12 words in a password manager or write them down and
-            store in a secure location
+            Save these{' '}
+            {recoveryPhrase ? recoveryPhrase.split(/\s+/).length : ''} words in
```
vyorkin, line 78: Replaced hardcoded "12 words"

src/ui/pages/Backup/Success/Success.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <VStack gap={24}>
+            <div className={styles.title}>
+              Nicely Done
+              <br />
+              Self-Custodial Human!
+            </div>
+            <UIText kind="headline/h3" color="var(--always-white)">
+              Zerion makes exploring web3 feel better than ever.
+              {isNarrowView ? ' ' : <br />}
+              You can close this tab now.
```
vyorkin, line 116: "You can close this tab now." instead of smth like "you can now get started"

src/ui/pages/Backup/Success/Success.tsx
vyorkin, line 1: This one is very similar this the `onboarding/Success`, same appearance & behaviour, but with updated text & not "you can pin the extension" hint

src/ui/App/App.tsx
```diff
@@ -257,10 +269,10 @@ function Views({ initialRoute }: { initialRoute?: string }) {
             }
           />
           <Route
-            path="/backup-wallet/*"
+            path="/reveal-private-key/*"
             element={
               <RequireAuth>
-                <BackupWallet />
+                <RevealPrivateKey />
```
vyorkin, line 275: The `RevealPrivateKey` page/component is essentially the old BackupWallet, but stripped down to only include the "reveal private key" functionality. I also changed the route to better reflect its purpose

src/ui/App/App.tsx
```diff
@@ -433,28 +446,28 @@ export interface AppProps {
...
 
   const { connected } = useStore(runtimeStore);
 
   useBodyStyle(
     useMemo(() => ({ opacity: connected ? '' : '0.6' }), [connected])
   );
 
-  const isOnboardingTemplate =
+  const isOnboardingView =
     mode === 'onboarding' && initialView !== 'handshakeFailure';
```
vyorkin, line 470: I'm not sure what the "template" suffix is supposed to mean here, but I find it easier to think of it as "onboarding view

src/ui/features/onboarding/initialization.ts
```diff
@@ -0,0 +1,33 @@
...
+import { WindowContext, WindowParam } from 'src/ui/shared/WindowContext';
+import { OnboardingInterrupt } from './errors';
+
+async function getMode(windowContext: WindowContext) {
+  const hasOnboardingUrl = document.location.hash.startsWith('#/onboarding');
+  const hasExistingUser = Boolean(await getCurrentUser());
+  return hasOnboardingUrl || (!windowContext.isPopup() && !hasExistingUser)
+    ? 'onboarding'
+    : 'wallet';
+}
```
vyorkin, line 12: I extracted this function. Yes, we access the browser storage _twice_ in the `maybeOpenOnboarding` function, but it's a bit easier to read this way

src/ui/features/onboarding/initialization.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (mode === 'onboarding' && !windowContext.isPopup()) {
+    // TODO: setting "mode=onboarding" is duplicated here and in {openOnboarding(url)}
+    const searchParams = new URLSearchParams(window.location.search);
+    if (!searchParams.has(WindowParam.mode)) {
+      searchParams.append(WindowParam.mode, 'onboarding');
+      window.location.search = searchParams.toString();
+    }
+  }
+  return { mode } as const;
+}
```
vyorkin, line 33: I renamed `userHasWallets` to `hasExistingUser` since that’s what it represents.
Honestly, I find this logic overcomplicated, but I haven’t spent enough time to debug and think about the best way to refactor it. At least it should be possible to get rid of the `OnboardingInterrupt` exception

src/ui/shared/useSizeStore.ts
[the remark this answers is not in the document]
zerts: This is a legacy from web app. Kinda used to this name :))

zerts:
If you open backup flow, then click on the extension icon and move back, it will show empty screen because session is expired. It is better to
- move back to password screen
- show "session expired" error (already implemented)
- keep session for backup flow somehow
Zerts Screenshot 2024-07-10 at 11 18@2x
Probably moving to password screen is the easiest workaround for now, but the third one can be more convenient for the user

zerts:
We don't have "Skip backup" option anymore during onboarding. Is it expected?
Zerts Screenshot 2024-07-10 at 11 21@2x

zerts:
May be we can change this title for Backup flow (not a blocker)
Zerts Screenshot 2024-07-10 at 11 23@2x

zerts:
If you press Back button from Success screen, you will see empty state (previously it was a Session Expired error).
Screenshot from the production version
Zerts Screenshot 2024-07-10 at 11 26@2x

src/ui/shared/WindowContext.ts
zerts, line 1: This file is probably should be inside `src/shared` cause it is actively used in background's files

src/ui/shared/WindowContext.ts
```diff
@@ -0,0 +1,50 @@
...
+export class WindowContext {
+  readonly windowType: WindowType;
+  readonly windowLayout: WindowLayout;
+
+  private constructor(windowType: WindowType, windowLayout: WindowLayout) {
+    this.windowType = windowType;
+    this.windowLayout = windowLayout;
+  }
+
+  static get() {
```
zerts, line 24: Why not create a singleton entity here and make constructor public?

src/ui/shared/WindowContext.ts
```diff
@@ -0,0 +1,50 @@
+export type WindowType = 'popup' | 'tab' | 'dialog';
+export type WindowLayout = 'column' | 'page';
+
+export enum WindowParam {
+  windowType = 'windowType',
+  windowLayout = 'windowLayout',
+  mode = 'mode',
+}
+
+function getParam<T>(param: WindowParam) {
```
zerts, line 10: Maybe we can move setParam logic here too?

src/ui/shared/WindowContext.ts
```diff
@@ -0,0 +1,50 @@
+export type WindowType = 'popup' | 'tab' | 'dialog';
+export type WindowLayout = 'column' | 'page';
+
+export enum WindowParam {
+  windowType = 'windowType',
+  windowLayout = 'windowLayout',
+  mode = 'mode',
```
zerts, line 7: Should we move work with mode to this file too?

src/ui/shared/getPopupUrl.ts
```diff
@@ -0,0 +1,9 @@
+import browser from 'webextension-polyfill';
+
+export function getPopupUrl() {
```
zerts, line 3: This function is not used anywhere

src/ui/ui-kit/Stack/Stack.tsx
zerts: Also absolute imports are good for mocks inside ui-lab. So I agree with such changes :))

src/ui/shared/requests/useWalletGroups.ts
```diff
@@ -2,6 +2,14 @@ import { useQuery } from '@tanstack/react-query';
 import { walletPort } from 'src/ui/shared/channels';
 import { checkForTestAddress } from '../meta-app-state';
 
+export function useWalletGroup({ groupId }: { groupId: string }) {
+  return useQuery({
+    queryKey: [`wallet/uiGetWalletGroup/${groupId}`],
```
zerts, line 7: `['wallet/uiGetWalletGroup', groupId]`

src/ui/shared/WindowContext.ts
vyorkin: Sometimes we want to be able to just import the types from this file, without accessing `window` (for example, in background script). To me it is just safer to call the `get` function explicitly, when needed. This way we can be sure we're not accessing anything we didn't want to access. We could move the types somewhere, but they seem to belong here, so...

vyorkin: Also this way we can be sure it is always up to date (since we read query string parameters every time we're calling the `get` fn)

zerts:
Ledger onboarding layout is broken :(
Zerts Screenshot 2024-07-10 at 11 54@2x

src/ui/shared/WindowContext.ts
vyorkin: I thought about this too... I kept it in the `ui/` directory because it is describing _window_ parameters (its context)

src/ui/shared/WindowContext.ts
zerts: If I understand right, importing types should be safe with import type {} syntax

zerts: Good moment about the up-to-date state.
Let me think

zerts: I think, we can still use just a Constuctor everywhere.
So const windowContext = new WindowContext() should work fine everywhere in ui

src/ui/App/App.tsx
zerts: +
I hope it is a bit more flexible to use them this way tbh

src/ui/App/App.tsx
```diff
@@ -433,28 +448,28 @@ export interface AppProps {
 }
 
 export function App({ initialView, mode, inspect }: AppProps) {
+  const windowContext = WindowContext.get();
   const bodyClassList = useMemo(() => {
     const result = [];
-    if (templateData.windowContext === 'dialog') {
+    if (windowContext.isDialog()) {
```
zerts, line 454: can we move windowContext inside useMemo?

src/ui/shared/openUrl.ts
```diff
@@ -0,0 +1,37 @@
...
+    active: true,
+    lastFocusedWindow: true,
+  });
+  browser.tabs.create({
+    url: url.toString(),
+    index: activeTab?.index + 1,
+  });
+}
+
+export function openHref(event: React.MouseEvent, params?: WindowParams) {
```
zerts, line 30: May be it is better to call this `openExternalHref`? To avoid potential usage for links inside extension?

src/ui/shared/WindowContext.ts
zerts: May be I was wrong, but I used window inside src/shared if it is truly shared between ui and background. Not critical anyway

src/ui/shared/openUrl.ts
zerts: Ah, I see. Well, probably this is ok then

src/ui/pages/RevealPrivateKey/usePrivateKey.ts
```diff
@@ -0,0 +1,19 @@
+import { useQuery } from '@tanstack/react-query';
+import { invariant } from 'src/shared/invariant';
+import { walletPort } from 'src/ui/shared/channels';
+
+async function getPrivateKey(address: string) {
+  return walletPort.request('getPrivateKey', { address });
```
zerts, line 6: This can be an inline call inside usePrivateKey.
I think, we don't need to wrap port requests in general

src/ui/features/onboarding/Success/styles.module.css
zerts: We will need to check this in production build

src/ui/pages/Backup/Success/Success.tsx
zerts: May be we can remove confetti from this one? Not sure we need them here :)

src/ui/features/onboarding/Create/Create.tsx
```diff
@@ -10,20 +10,15 @@ import { PrivacyFooter } from 'src/ui/components/PrivacyFooter';
...
+import { isSessionExpiredError } from 'src/ui/shared/isSessionExpiredError';
+import { Stack } from 'src/ui/ui-kit/Stack';
 import { Password } from '../Password';
-import { Stack } from '../Stack';
 import { assertPasswordStep } from '../Password/passwordSearchParams';
 import * as helperStyles from '../shared/helperStyles.module.css';
-import { isSessionExpiredError } from '../shared/isSessionExpiredError';
+import { useOnboardingSession } from '../shared/useOnboardingSession';
 
-export function CreatePassword({
```
zerts, line 20: Let's keep CreatePassword name here

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
```diff
@@ -256,8 +250,16 @@ export function WalletGroup() {
...
+                  to: isMnemonicGroup
+                    ? `/backup?groupId=${walletGroup.id}`
+                    : undefined,
+                  onClick: isMnemonicGroup
+                    ? (event) =>
+                        openHref(event, {
+                          windowType: 'tab',
+                          windowLayout: 'page',
+                        })
+                    : undefined,
```
zerts, line 262: This condition looks a bit strange tbh. Can it be not a mnemonicGroup and what will happen in this case?

src/ui/features/onboarding/Create/Create.tsx
vyorkin: I renamed it because it creates a user (by creating a password and encryption key..., but these are internal details of the function we call). It literally calls the `createUser` function to create a user. So now, it matches the rest of the single-word component names like `Welcome`, `FAQ`, `Import`, etc. On the other hand, the `CreatePassword` name does match the page title

src/ui/shared/openUrl.ts
vyorkin: Yeah, it is for internal links :)

src/ui/shared/getPopupUrl.ts
vyorkin: Ah, alright, that's unfinished refactoring :) I noticed the same pattern to get the default popup URL being used over and over again, so I extracted a function called `getPopupUrl`. I intended to use it but forgot to. Thanks, I'll update other places to call `getPopupUrl` instead of repeating this:
```
  const popupUrl = browser.runtime.getManifest().action?.default_popup;
  if (!popupUrl) {
    throw new Error('popupUrl not found');
  }
  return new URL(browser.runtime.getURL(popupUrl));
```

src/ui/shared/WindowContext.ts
vyorkin: > importing types should be safe with import type {} syntax
Right, it's fine until we need to import something that's not a type, like enum WindowParam -- that's what led me to the current design. Moving WindowParam to another place just didn't feel right

> So const windowContext = new WindowContext() should work fine everywhere in UI
Yes, I think so. At least for now, it should work

src/ui/shared/WindowContext.ts
vyorkin: We don't really access the `mode` parameter anywhere except for `useBackupContext`, where it needs to be checked explicitly (using equality comparison) to help with type inference since the `BackupContext` is kind of a discriminated union

src/ui/shared/WindowContext.ts
vyorkin: Ah, I think I got the idea. It would be nice to have something well-typed that we can use instead of manually setting search parameters like this:
```
popupUrl.searchParams.append(WindowParam.windowType, 'tab');
popupUrl.searchParams.append(WindowParam.mode, 'onboarding');
```
Let me think...

src/ui/features/onboarding/Success/styles.module.css
vyorkin: Thanks, indeed! Added to the test plan in the PR description

src/ui/pages/Backup/Success/Success.tsx
vyorkin: Agree, I was doubtful about whether I should remove it. It does look fancy, though

src/ui/shared/WindowContext.ts
vyorkin: I came up with a better design, check it out

src/ui/shared/WindowContext.ts
vyorkin: ✅

src/ui/shared/WindowContext.ts
vyorkin: ✅ Added `appMode` (prev. `context`) to the `WindowContext` class + extra functions:
```
  isOnboardingMode() {
    return this.params.appMode === 'onboarding';
  }

  isWalletMode() {
    return this.params.appMode === 'wallet';
  }
```

src/ui/shared/getPopupUrl.ts
vyorkin: Refactored, now its being used in 3 functions, also `openOnboarding` doesn't take the url parameter anymore (it calls the `getPopupUrl` as well)

src/ui/shared/requests/useWalletGroups.ts
vyorkin: ✅

src/ui/App/App.tsx
vyorkin: I think it is not relevant anymore, since `windowContext` is now a global constant ✅

src/ui/pages/RevealPrivateKey/usePrivateKey.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function usePrivateKey(address: string | null) {
+  invariant(address, 'Address param is required for privateKey seedType');
```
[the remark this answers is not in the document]
vyorkin: Thats refactoring leftover, thanks!

src/ui/pages/RevealPrivateKey/usePrivateKey.ts
vyorkin: Post-refactoring cringe moments... thanks! ✅

src/ui/index.tsx
```diff
@@ -13,10 +13,10 @@ import type { AppProps } from './App/App';
 import { initialize as initializeChannels } from './shared/channels';
 import { queryClient } from './shared/requests/queryClient';
 import { emitter } from './shared/events';
-import { maybeOpenOboarding } from './Onboarding/initialization';
-import { OnboardingInterrupt } from './Onboarding/errors';
 import { persistQueryClient } from './shared/requests/queryClientPersistence';
 import { getPreferences } from './features/preferences/usePreferences';
+import { OnboardingInterrupt } from './features/onboarding/errors';
+import { maybeOpenOboarding } from './features/onboarding/initialization';
```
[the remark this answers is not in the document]
vyorkin: Missing letter effect :) ✅

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
vyorkin: Agree, let me try to explain how I understand this part of display logic.
Here we are already in a context where`isSignerGroup = true`, which means that it is a wallet created/imported using either mnemonic phrase or a private key.
And now instead of having
```
to: `/backup-wallet?groupId=${walletGroup.id}&backupKind=verify`,
onClick: needsBackup(walletGroup) ? openInTabView : undefined,
```
we have
```
to: isMnemonicGroup ? `/backup?groupId=${walletGroup.id}` : undefined,
onClick: isMnemonicGroup
  ? (event) =>
    openHref(event, {
      windowType: 'tab',
      windowLayout: 'page',
    }) : undefined,
```
So basically nothing have changed significantly, except that I fixed a couple of bugs here (user was unable to backup already backedup wallet and missing `needsBackup` condition for the "to" attribute)

vyorkin: I used `isMnemonicGroup` instead of `needsBackup` here because I believe users should have the option to back up their wallet, even if it was imported or already backed up. If you have any concerns or think otherwise, let's discuss

src/ui/DNA/shared/WideScreen.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useSizeStore } from 'src/ui/Onboarding/useSizeStore';
+import { useWindowSizeStore } from 'src/ui/shared/useWindowSizeStore';
```
zerts, line 2: I'm not sure that we need to rename this hook in this pr, but ok

vyorkin: Yeah, it's used everywhere in backup and onboarding, like on almost every page:
```
const { isNarrowView } = useWindowSizeStore();
```
So I decided to update it now (sorry if this change caused extra review work)

src/ui/shared/WindowParam.ts
```diff
@@ -0,0 +1,26 @@
...
+  windowLayout = 'windowLayout',
+  appMode = 'appMode',
+}
+
+export function getWindowParam<T>(param: WindowParam, defaultValue?: T) {
+  const url = new URL(window.location.href);
+  return (url.searchParams.get(param) as T) || defaultValue;
+}
+
+export function setAppMode(params: URLSearchParams, value: AppMode) {
```
zerts, line 16: Personally, I would probably  prefer one function to set all params. May be this would be easier for usage. However, not critical

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -8,21 +8,45 @@ import ArrowLeftIcon from 'jsx:src/ui/assets/arrow-left.svg';
...
+  context: BackupContext,
+  expected: string,
+  actual: string
+) {
+  if (context.appMode === 'onboarding') {
+    if (expected !== actual) {
+      throw new Error('Incorrect seed phrase');
+    }
+  } else {
+    const isCorrect = await walletPort.request('verifyRecoveryPhrase', {
```
zerts, line 36: If I understand, we have expected value here in both contexts, so this helper can be simplified to the first condition

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -31,23 +55,35 @@ export function VerifyBackup({ onSuccess }: { onSuccess(): void }) {
...
+  const {
+    data: recoveryPhrase,
+    isLoading,
+    error,
+  } = useRecoveryPhrase(backupContext);
 
   useEffect(() => {
     if (isSessionExpiredError(error)) {
-      navigate('/onboarding/session-expired', { replace: true });
+      if (backupContext.appMode === 'onboarding') {
```
zerts, line 67: Looks like this logic should be described inside external param onSessionExpired
In general, it would be nice to guard backup flow from describing different behaviour for different context as much as possible and move everything to the top level.

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -8,21 +8,45 @@ import ArrowLeftIcon from 'jsx:src/ui/assets/arrow-left.svg';
...
 const INPUT_NUMBER = 12;
 const ARRAY_OF_NUMBERS = Array.from({ length: INPUT_NUMBER }, (_, i) => i);
 
+async function verifyRecoveryPhrase(
+  context: BackupContext,
+  expected: string,
+  actual: string
+) {
+  if (context.appMode === 'onboarding') {
+    if (expected !== actual) {
```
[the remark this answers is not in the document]
vyorkin: Yes, I think it's better to force lowercase here. Previously, we didn't use toLowerCase [here](https://github.com/zeriontech/zerion-wallet-extension/blob/0a6f7715a877100bdd8fe4e088c82522a4a4a144/src/ui/pages/BackupWallet/BackupWallet.tsx#L279-L284), so I left it unchanged before the initial refactoring

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+            <form
+              onSubmit={(event) => {
+                event.preventDefault();
+                const password = new FormData(event.currentTarget).get(
+                  'password'
+                ) as string | undefined;
+                if (!password) {
+                  return;
+                }
+                if (!user) {
```
zerts, line 114: I think, we can write invariant for user right after `if (isLoading) return null`

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+              <UIText kind="body/regular">
+                This password is required to reveal your recovery phrase.
+              </UIText>
+            </VStack>
+            <form
+              onSubmit={(event) => {
+                event.preventDefault();
+                const password = new FormData(event.currentTarget).get(
+                  'password'
+                ) as string | undefined;
```
zerts, line 110: Can password be undefined when it is a required field?

src/ui/pages/Backup/useBackupContext.ts
```diff
@@ -0,0 +1,34 @@
...
+  return new URLSearchParams([...params.entries(), ...hashParams.entries()]);
+}
+
+export function useBackupContext(): BackupContext {
+  const params = useUrlParams();
+
+  if (windowContext.isOnboardingMode()) {
+    return { appMode: 'onboarding' };
+  } else {
+    const groupId = params.get('groupId');
```
zerts, line 30: Can groupId be inside global search params?
Do we really need to combine all params to get this value here?

zerts: If we pass group id as query param, may be it is better to pass it as hash param from the first place?

src/ui/pages/RevealPrivateKey/RevealPrivateKey.tsx
```diff
@@ -0,0 +1,171 @@
...
+      ) : null}
+      {params.get('step') === 'revealSecret' ? (
+        <Reveal
+          groupId={groupId}
+          address={address}
+          onSessionExpired={() => {
+            updateSearchParam('step', 'verifyUser', { replace: true });
+          }}
+          onSubmit={() => {
+            navigate('/');
```
zerts, line 165: I think, we should zeroize here, shouldn't we?

src/ui/pages/RevealPrivateKey/RevealPrivateKey.tsx
```diff
@@ -0,0 +1,171 @@
...
+export function RevealPrivateKey() {
+  useBackgroundKind(whiteBackgroundKind);
+
+  const [params, setSearchParams] = useSearchParams();
+  const groupId = params.get('groupId');
+  const address = params.get('address');
+  invariant(groupId, 'groupId param is required for RevealPrivateKey view');
+  invariant(address, 'address param is required for RevealPrivateKey view');
+
+  const updateSearchParam = (
```
zerts, line 124: We've been using this helper a lot. May be it worth extracting into a file?

src/ui/shared/UrlContext.ts
```diff
@@ -0,0 +1,97 @@
...
+      this.params.windowLayout = params.windowLayout;
+    }
+    if (params.windowType) {
+      searchParams.set(UrlContextParam.windowType, params.windowType);
+      this.params.windowType = params.windowType;
+    }
+    if (params.appMode) {
+      searchParams.set(UrlContextParam.appMode, params.appMode);
+      this.params.appMode = params.appMode;
+    }
```
vyorkin, line 41: I avoided using a for loop here, so there is some duplication instead of
```
for (const [key, value] of Object.entries(params)) {
  if (value) {
    ...
  }
```

src/ui/shared/UrlContext.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+class UrlContext {
+  protected params: UrlContextParams;
+
+  constructor(params: UrlContextParams) {
+    this.params = params;
+  }
+
+  set(searchParams: URLSearchParams, params: UrlContextParams) {
```
vyorkin, line 29: The only way to set the URL context parameters is by using this `set` function. It also keeps our global params object (see below) in sync with the window's `URLSearchParams`

src/ui/pages/RevealPrivateKey/RevealPrivateKey.tsx
vyorkin: Yep, I noticed it's duplicated in at least two places. I thought about fixing it, but I've already made so many changes in this PR. Since you mentioned it, I'll go ahead and extract it :)

src/ui/shared/WindowParam.ts
vyorkin: I refactored this part again, and now we have three contexts:

- `urlContext` (with a set method to set all params at once)
- `windowContext` (for querying info about the window type and layout)
- `appContext` (for global app mode info: `onboarding` or `wallet`)

I think it looks better now since `appMode` isn't really part of `windowContext` like it was before.

src/ui/pages/Backup/VerifyBackup.tsx
vyorkin: Actually you're right, thanks! ✅

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { UnstyledButton } from 'src/ui/ui-kit/UnstyledButton';
+import { useNavigate } from 'react-router-dom';
+import * as styles from './styles.module.css';
+
+export function VerifyUser({ onSuccess }: { onSuccess: () => void }) {
+  const { isNarrowView } = useWindowSizeStore();
+
+  const navigate = useNavigate();
+  const handleBackClick = useCallback(() => {
+    navigate(-1);
```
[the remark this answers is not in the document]
vyorkin: ah, didn't know we have `useGoBack` hook :) 👍 ✅

src/ui/pages/RevealPrivateKey/RevealPrivateKey.tsx
vyorkin: The `zeroizeAfterSubmission` function does the following:
```
export function zeroizeAfterSubmission() {
  /** Call this function to remove sensitive input values from runtime memory */
  const inputs = document.querySelectorAll('input[type=password]');
  inputs.forEach((input) => {
    if (input instanceof HTMLInputElement) {
      input.value = '';
    }
  });
  queryClient.getMutationCache().clear();
}
```
So, it specifically targets inputs with `type=password` and clears the mutation cache. But from what I see, we don't have any input `type="password"` on this screen, and there are no mutations either

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
vyorkin: Basically, this is copied from [here](https://github.com/zeriontech/zerion-wallet-extension/blob/0a6f7715a877100bdd8fe4e088c82522a4a4a144/src/ui/pages/Login/Login.tsx#L82-L91) (or [here](https://github.com/zeriontech/zerion-wallet-extension/blob/0a6f7715a877100bdd8fe4e088c82522a4a4a144/src/ui/components/VerifyUser/VerifyUser.tsx#L71-L81)). I'd like to keep this logic as close as possible to the original so it's easier to refactor identical parts in the future

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
vyorkin: I didn't want to change it for the same reason I mentioned in [this comment](https://github.com/zeriontech/zerion-wallet-extension/pull/569#discussion_r1678492732)

vyorkin:
> We don't have "Skip backup" option anymore during onboarding. Is it expected?

Fixed ✅

vyorkin:
> May be we can change this title for Backup flow (not a blocker)
✅

vyorkin:
> If you press Back button from Success screen, you will see empty state (previously it was a Session Expired error).

Looks like it is fixed ✅ 
I can't think of any reasons why should we expire the session when user goes back from the success screen... But now it works like this

Uploading Screen Recording 2024-07-16 at 13.11.15.mov…

src/ui/pages/Backup/VerifyBackup.tsx
vyorkin: Agree, we can improve this even more!
Let me try to extract this session-expiration-checking logic and move it to the parent component

src/ui/App/App.tsx
everdimension: Thanks for clarifying

src/ui/shared/useUpdateSearchParam.ts
```diff
@@ -0,0 +1,23 @@
...
+    value: string,
+    navigateOptions?: Parameters<typeof setSearchParams>[1]
+  ) => {
+    const newParams = new URLSearchParams(params);
+    newParams.set(key, value);
+    setSearchParams(newParams, navigateOptions);
+  };
+
+  return [params, updateSearchParam];
+}
```
everdimension, line 23: In my opinion name pattern of `use<verb>` is not a good one

Second, returning a tuple with `[state, setState]` isn't necessary for a custom hook; we can return regular object. The makes sense for `useState` because it's supposed to be used often. This hook isn't.

Third, I don't think this hook is helpful. The custom logic it contains is just a pure function. 
We should aim to reuse js code, not react-specific code.

It's better, in my opinion, to use the provided `useSearchParams` hook with a helper fn:

```
const [params, setSearchParams] = useSearchParams();
...
setSearchParams(setParam(params, key, value))
```

src/ui/shared/openUrl.ts
```diff
@@ -0,0 +1,27 @@
+import browser from 'webextension-polyfill';
+import { urlContext, type UrlContextParams } from './UrlContext';
+
+async function getNextToActiveTabIndex() {
+  const [activeTab] = await browser.tabs.query({
+    active: true,
+    lastFocusedWindow: true,
+  });
+  return activeTab?.index + 1;
```
everdimension, line 9:
```suggestion
  return activeTab.index + 1;
```

everdimension, line 9:
```suggestion
  return (activeTab?.index ?? 0) + 1;
```

everdimension: Although maybe it's better to return `null` and handle it in the call site

src/ui/shared/UrlContext.ts
```diff
@@ -0,0 +1,82 @@
+export type AppMode = 'onboarding' | 'wallet';
+export type WindowType = 'popup' | 'tab' | 'dialog';
+export type WindowLayout = 'column' | 'page';
```
everdimension, line 3: Two of these aren't used outside

src/ui/shared/UrlContext.ts
```diff
@@ -0,0 +1,82 @@
...
+  return (url.searchParams.get(param) as T) || defaultValue;
+}
+
+export interface UrlContextParams {
+  appMode?: AppMode;
+  windowType?: WindowType;
+  windowLayout?: WindowLayout;
+}
+
+class UrlContext {
```
everdimension, line 22: Why are these classes? They do not hold state

vyorkin: Ah, they did, just one commit before... Post-refactoring cringe moments, as always. I should come up with smth better for this

src/ui/shared/WindowContext.ts
vyorkin: I came up with even better design :) 
Now we have single function with the following signature:
```
export function setUrlContext(
  searchParams: URLSearchParams,
  params: Partial<UrlContext>
)
```

src/ui/shared/UrlContext.ts
vyorkin: Now all 3 are internal / not exported anymore ✅

src/ui/shared/UrlContext.ts
vyorkin: I simplified the `UrlContext.ts`, please check out the new version

src/ui/shared/openUrl.ts
vyorkin: I prefer your suggested change. The caller would always want to fall back to tab index = 0 anyway

src/ui/shared/openUrl.ts
```diff
@@ -0,0 +1,31 @@
+import browser from 'webextension-polyfill';
+import { setUrlContext } from './setUrlContext';
+import type { UrlContext } from './UrlContext';
+
+async function getNextToActiveTabIndex() {
+  const [activeTab] = await browser.tabs.query({
+    active: true,
+    lastFocusedWindow: true,
+  });
+  return (activeTab?.index ?? -1) + 1;
```
everdimension, line 10: I just realized that `0` index is a bad default. We shouldn't open the tab as the first one, but as a last one. So it's better to introduce null here I think

src/ui/DNA/components/MintBanner.tsx
```diff
@@ -32,7 +32,12 @@ export function MintBanner({
         <UnstyledAnchor
           href={`#/mint-dna?address=${address}`}
           target="_blank"
-          onClick={openInPageTabView}
+          onClick={(event) =>
+            openHref(event, {
+              windowType: 'tab',
+              windowLayout: 'page',
+            })
+          }
```
everdimension, line 40: can this fit in one line?

src/ui/DNA/components/UpgradeBanner.tsx
```diff
@@ -28,7 +28,12 @@ export function UpgradeBanner({
         <UnstyledAnchor
           href={`#/upgrade-dna?address=${address}`}
           target="_blank"
-          onClick={openInPageTabView}
+          onClick={(event) =>
+            openHref(event, {
+              windowType: 'tab',
+              windowLayout: 'page',
+            })
+          }
```
everdimension, line 36: https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#r1682651806

src/ui/DNA/pages/MintDnaFlow/MintDna.tsx
```diff
@@ -30,9 +30,9 @@ import { AddressDetails } from 'src/ui/pages/Receive/AddressDetails';
 import type { HTMLDialogElementInterface } from 'src/ui/ui-kit/ModalDialogs/HTMLDialogElementInterface';
 import { INTERNAL_ORIGIN } from 'src/background/constants';
 import { invariant } from 'src/shared/invariant';
-import { SidePanel } from 'src/ui/Onboarding/FAQ/SidePanel';
 import { useGasPrices } from 'src/ui/shared/requests/useGasPrices';
 import { useCurrency } from 'src/modules/currency/useCurrency';
+import { SidePanel } from 'src/ui/components/SidePanel';
```
everdimension, line 35: This is a bad generic name
Can we place it in `src/ui/Onboarding`? Or maybe in `src/ui/Onboarding/components`? It's onboarding-specific

src/ui/DNA/pages/MintDnaFlow/MintDnaFlow.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Route, Routes } from 'react-router-dom';
-import { PageLayout } from 'src/ui/Onboarding/shared/PageLayout/PageLayout';
+import { PageLayout } from 'src/ui/components/PageLayout/PageLayout';
```
everdimension, line 3: Same here, this "PageLayout" is onboarding specific. Let's either rename or put in `src/ui/Onboarding/components`

src/ui/features/onboarding/shared/useOnboardingSession.ts
everdimension: 🤔

src/ui/shared/UrlContext.ts
```diff
@@ -0,0 +1,26 @@
...
+
+function getSearchParam<T>(param: UrlContextParam, defaultValue: T) {
+  const url = new URL(window.location.href);
+  return (url.searchParams.get(param) as T) || defaultValue;
+}
+
+export const urlContext = {
+  appMode: getSearchParam(UrlContextParam.appMode, 'wallet'),
+  windowType: getSearchParam(UrlContextParam.windowType, 'popup'),
+  windowLayout: getSearchParam(UrlContextParam.windowLayout, 'column'),
```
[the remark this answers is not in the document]
vyorkin: Improved ✅

src/ui/DNA/components/MintBanner.tsx
vyorkin: ✅

src/ui/DNA/components/UpgradeBanner.tsx
vyorkin: ✅

src/ui/shared/openUrl.ts
vyorkin: I just can't imagine how should we handle that `null` value. If `null` represents the case when there is no active tab then we should create the first tab anyway and it will get `index = 0`, right?

src/background/NotificationWindow/createBrowserWindow.ts
```diff
@@ -1,25 +1,16 @@
 import type { Windows } from 'webextension-polyfill';
 import browser from 'webextension-polyfill';
 import { nanoid } from 'nanoid';
+import { getPopupUrl } from 'src/ui/shared/getPopupUrl';
+import { setUrlContext } from 'src/ui/shared/setUrlContext';
```
everdimension, line 5: I think both of these belong to `/src/shared/`

src/ui/features/onboarding/Create/Create.tsx
everdimension: Why not name it `CreateUser` then? Just `Create` is ambiguous

src/ui/components/EraseData/EraseDataListButton.tsx
```diff
@@ -21,10 +21,11 @@ export function EraseDataListButton({
   const dialogRef = useRef<HTMLDialogElementInterface | null>(null);
   const eraseAllData = useEraseDataMutation({
     onSuccess: () => {
-      if (templateData.windowContext === 'tab') {
+      const isTab = urlContext.windowType === 'tab';
+      if (isTab) {
         emitter.emit('reloadExtension');
       } else {
-        maybeOpenOboarding();
+        maybeOpenOnboarding();
```
everdimension, line 28: 👍

src/ui/components/SidePanel/SidePanel.tsx
everdimension, line 1: https://github.com/zeriontech/zerion-wallet-extension/pull/569#discussion_r1682653461

src/ui/features/onboarding/Hardware/Hardware.tsx
```diff
@@ -112,7 +114,7 @@ export function Hardware() {
             </div>
           ) : null}
           <UnstyledButton
-            onClick={() => navigate(-1)}
+            onClick={goBack}
```
everdimension, line 117: 👍

src/ui/features/onboarding/Onboarding.tsx
```diff
@@ -45,6 +46,14 @@ export function Onboarding() {
           </PageLayout>
         }
       />
+      <Route
+        path="/onboarding/backup/*"
+        element={
+          <PageLayout>
+            <Backup />
+          </PageLayout>
```
everdimension, line 54: https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#r1680870245

Though if you also wrap in `<PageLayout />` here, I don't see why it can't just be inside `<Backup />`?

src/ui/features/onboarding/Password/Password.tsx
```diff
@@ -70,10 +71,7 @@ function WeakPasswordWarning() {
           ref={focusNode}
           kind="primary"
           as={UnstyledLink}
-          onClick={(event) => {
-            event.preventDefault();
-            navigate(-1);
-          }}
+          onClick={goBack}
```
everdimension, line 74: Aren't we losing `preventDefault`?
Though I don't know what it was for, ideally we should comment

src/ui/features/onboarding/initialization.ts
```diff
@@ -0,0 +1,34 @@
+import { getCurrentUser } from 'src/shared/getCurrentUser';
+import { openOnboarding } from 'src/shared/openOnboarding';
+import { UrlContextParam, urlContext } from 'src/ui/shared/UrlContext';
+import { setUrlContext } from 'src/ui/shared/setUrlContext';
+import { OnboardingInterrupt } from './errors';
+
+async function getAppMode() {
+  const hasOnboardingUrl = document.location.hash.startsWith('#/onboarding');
+  const hasExistingUser = Boolean(await getCurrentUser());
```
everdimension, line 9: You're querying `await getCurrentUser()` twice. Let's pass `{ hasExistingUser: boolean }` as a param instead? This will also make this function synchronous.

src/ui/features/onboarding/Password/Password.tsx
everdimension: I see there a `to="..."` prop lower down. `preventDefault` was obviously used to handle (prevent) that transaction, so it should be preserved

src/ui/pages/Backup/useBackupContext.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Combines URL search parameters from the window's location and react-router's search params.
+ * @returns {URLSearchParams} Merged search parameters from both sources.
+ */
+function useUrlParams(): URLSearchParams {
+  const params = new URLSearchParams(window.location.search);
+  const [hashParams] = useSearchParams();
+  return new URLSearchParams([...params.entries(), ...hashParams.entries()]);
+}
+
+export function useBackupContext(): BackupContext {
```
everdimension, line 24: I don't see why we need this hook

Consumers can read `appMode` from `urlContext` and `groupId` from `useSearchParams`

src/ui/pages/Backup/Backup.tsx
```diff
@@ -0,0 +1,151 @@
...
+      <Route
+        path="/verify-user"
+        element={
+          <VerifyUser
+            onSuccess={() => navigate(`recovery-phrase${searchParams}`)}
+          />
+        }
+      />
+      <Route
+        path="/recovery-phrase"
```
everdimension, line 130: I do believe that at some point the notion of "backup" will exist for both private key and seed phrase. Therefore i'd give this route a more generic name, e.g. `/secret-value` or `/reveal` or something like that

src/ui/features/onboarding/Password/Password.tsx
vyorkin: ah, it has `as={UnstyledLink}`, I missed that somehow...

src/ui/pages/Backup/useBackupContext.ts
vyorkin: It allows not to do `invariant(groupId, "error msg")` every time

vyorkin: Ok, lets remove it

src/ui/features/onboarding/Onboarding.tsx
everdimension: I see, it's done here for consistency with sibling routes

src/ui/pages/Backup/Backup.tsx
```diff
@@ -0,0 +1,160 @@
...
+    <Routes>
+      <Route
+        path="/"
+        element={
+          <Navigate
+            to={{ pathname: 'info', search: searchParams }}
+            replace={true}
+          />
+        }
+      />
```
everdimension, line 107: Why do we need `/info` route? Can we display `<Info />` element on `/backup/`?

src/ui/DNA/pages/MintDnaFlow/MintDnaFlow.tsx
vyorkin: Moved to `src/ui/features/onboarding/shared`✅

src/ui/DNA/pages/MintDnaFlow/MintDna.tsx
vyorkin: Moved to `src/ui/features/onboarding/shared` ✅

src/ui/features/onboarding/Create/Create.tsx
```diff
@@ -51,6 +46,12 @@ export function CreatePassword({
     },
   });
 
+  const { sessionDataIsLoading } = useOnboardingSession('session-expired');
```
everdimension, line 49: This just doesn't make sense if you read it :)

src/ui/features/onboarding/initialization.ts
vyorkin: yep, thats better, thanks ✅

src/background/NotificationWindow/createBrowserWindow.ts
vyorkin: ✅ moved

src/ui/components/SidePanel/SidePanel.tsx
vyorkin: ✅ moved

src/ui/pages/Backup/Backup.tsx
vyorkin: Why not wait to make this change when it actually reflects the situation? Right now, it would look strange if the route is called `/reveal` or `/secret-value` but the `<RecoveryPhrase>` component is displayed. Or are you just avoiding changing the route in the future?

src/ui/pages/Backup/Backup.tsx
vyorkin: I see, I took bad example from some other set of routes.
yep, I think we can!

src/ui/features/onboarding/shared/SessionExpired.tsx
```diff
@@ -28,11 +28,11 @@ export function SessionExpired({ onSubmit }: { onSubmit(): void }) {
             Session expired
           </UIText>
           <UIText kind="body/regular" color="var(--neutral-600)">
-            Try create another wallet
+            Try creating another wallet
```
everdimension, line 31: 👍

src/ui/shared/WindowContext.ts
vyorkin: I see, now I understand the purpose of these 2 directories, moved `UrlContext` and others to `src/shared` ✅

src/ui/pages/Backup/Backup.tsx
everdimension: ok, extension routes are not public, so np

src/ui/pages/Backup/useRecoveryPhrase.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function useRecoveryPhrase(context: BackupContext) {
+  return useQuery({
+    queryKey: ['getRecoveryPhrase', context],
+    queryFn: () => getRecoveryPhrase(context),
+    suspense: false,
+    retry: 0,
+    cacheTime: 0,
+    refetchOnMount: false,
+    refetchOnWindowFocus: true,
```
everdimension, line 24: why?

src/ui/pages/Backup/useRecoveryPhrase.ts
```diff
@@ -0,0 +1,27 @@
+import { useQuery } from '@tanstack/react-query';
+import { walletPort } from 'src/ui/shared/channels';
+import type { BackupContext } from './useBackupContext';
```
everdimension, line 3: let's not use this

src/ui/pages/Backup/useRecoveryPhrase.ts
vyorkin: I only added `cacheTime: 0`, everything else is unchanged, I didn't refactor this part

vyorkin: but yes, it doesn't make sense :) I'll remove that

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -127,13 +143,13 @@ export function VerifyBackup({ onSuccess }: { onSuccess(): void }) {
               >
                 Verify
               </Button>
-              {validationError ? (
+              {verifyMutation.error ? (
```
everdimension, line 146:
```suggestion
              {verifyMutation.isError ? (
```

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -127,13 +143,13 @@ export function VerifyBackup({ onSuccess }: { onSuccess(): void }) {
...
                 <UIText
                   kind="caption/regular"
                   color="var(--negative-500)"
                   style={errorStyle}
                 >
-                  Incorrect seed phrase
+                  {(verifyMutation.error as Error).message || 'unknown error'}
```
everdimension, line 152: We have an awesome helper precisely for such cases 🙂:
```suggestion
                  {getError(verifyMutation.error).message}
```

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const faq = (
+    <VStack gap={24} style={{ alignContent: 'start' }}>
+      {isNarrowView ? null : (
+        <div className={styles.faqIcon}>
+          <div style={{ width: 20, height: 20 }}>☝️</div>
+        </div>
+      )}
+      <VStack gap={8}>
+        <UIText kind="small/regular" color="var(--neutral-600)">
+          For security, it's crucial to write down the recovery phrase and store
```
everdimension, line 62: apostrophe

src/ui/pages/Backup/useBackupContext.ts
vyorkin: ✅ removed

src/ui/pages/Backup/VerifyUser/VerifyUser.tsx
vyorkin: ✅ thanks

src/ui/pages/Backup/useRecoveryPhrase.ts
vyorkin: Refactored ✅

src/ui/features/onboarding/Create/Create.tsx
vyorkin: I updated it to use a named parameters object and renamed `navigateOnAccountCreated` to `navigateOnExistingUser`. I know this isn't a perfect solution, and I might come up with something better later (so, this isn't the final version)

src/ui/shared/openUrl.ts
everdimension: If no active tab is found, we should open tab as usual, making it last

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
everdimension: 1.  We do not show group view for PrivateKeyWalletContainer. How to see these conditionals in effect?
2. Let's do a helper for
```
const openInTabView = (event) => openHref(event, { windowType: 'tab', windowLayout: 'page' })
```
The diff will be smaller and easier to read

3. 
> I believe users should always have the option to back up their wallet

I agree 100%, but we went over this with product already and currently they disagree. Let's not introduce such changes unexpectedly

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
```diff
@@ -258,8 +252,16 @@ export function WalletGroup() {
...
```
everdimension, line 264:
```suggestion
                  to: `/backup?groupId=${walletGroup.id}`,
                  onClick: openInTabView,
```

src/ui/shared/openUrl.ts
everdimension: i.e. not pass `index` argument at all

src/ui/features/onboarding/Create/Create.tsx
everdimension: ping

src/ui/features/onboarding/Onboarding.tsx
```diff
@@ -66,12 +86,14 @@ export function Onboarding() {
         path="/onboarding/session-expired"
         element={
           <PageLayout>
-            <SessionExpired onSubmit={() => navigate('/onboarding')} />
+            <SessionExpired
+              onRestart={
+                existingUser ? () => navigate('/onboarding') : undefined
```
everdimension, line 91: Why `undefined`? Can't navigation restart is no user had been created?

src/ui/features/onboarding/Create/Create.tsx
vyorkin: updated ✅

src/ui/pages/Backup/Backup.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export function BackupComponent({
+  groupId,
+  onStart,
+  onExit,
+  onSkip,
+  onSuccess,
+  onSessionExpired,
+}: {
+  groupId?: string;
```
everdimension, line 23:
```suggestion
  groupId: string | null;
```

src/ui/pages/Backup/useRecoveryPhrase.ts
```diff
@@ -0,0 +1,26 @@
+import { useQuery } from '@tanstack/react-query';
+import { invariant } from 'src/shared/invariant';
+import { walletPort } from 'src/ui/shared/channels';
+
+export function useRecoveryPhrase({ groupId }: { groupId?: string }) {
+  return useQuery({
+    queryKey: ['getRecoveryPhrase', groupId],
+    queryFn: async () => {
+      if (groupId) {
+        invariant(groupId, 'groupId param is required for wallet appMode');
```
everdimension, line 10: why `invariant` if already checked for truthyness?

src/ui/pages/Backup/RecoveryPhrase.tsx
```diff
@@ -11,35 +10,44 @@ import ArrowLeftIcon from 'jsx:src/ui/assets/arrow-left.svg';
...
+  const { isNarrowView } = useWindowSizeStore();
+
+  const {
+    data: recoveryPhrase,
+    error,
+    isLoading,
+  } = useRecoveryPhrase({ groupId });
 
   useEffect(() => {
     if (isSessionExpiredError(error)) {
```
everdimension, line 44: let's not compute error when not necessary:
```suggestion
    if (isError && isSessionExpiredError(error)) {
```

src/ui/features/onboarding/Backup/Backup.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <BackupComponent
+      onStart={() => navigate('/onboarding/backup/recovery-phrase')}
+      onExit={() => navigate('/onboarding')}
+      onSkip={handleSkip}
+      onSuccess={handleSuccess}
+      onSessionExpired={() =>
+        navigate('/onboarding/session-expired', { replace: true })
+      }
```
everdimension, line 58: Since this value is used as a dependency of `useEffect`, we should pass a stable value, e.g. wrap in `useCallback`:

https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#diff-04ff63434a4618dc184dc118c89196327a3a3f3575863a2fc7a2ab11755a00ffR47

src/ui/pages/Backup/Backup.tsx
```diff
@@ -0,0 +1,104 @@
...
+    useErrorBoundary: true,
+  });
+
+  return (
+    <PageLayout>
+      <BackupComponent
+        groupId={groupId}
+        onStart={goToVerifyUser}
+        onSuccess={handleSuccess}
+        onSessionExpired={() => goToVerifyUser({ replace: true })}
```
everdimension, line 100: Since this value is used as a dependency of `useEffect`, we should pass a stable value, e.g. wrap in `useCallback`:

https://github.com/zeriontech/zerion-wallet-extension/pull/569/files#diff-04ff63434a4618dc184dc118c89196327a3a3f3575863a2fc7a2ab11755a00ffR47

src/ui/pages/Backup/useRecoveryPhrase.ts
vyorkin: thats my refactoring madness, thanks

everdimension (changes requested):
See comments

src/ui/shared/WindowContext.ts
vyorkin: Not relevant anymore

src/ui/shared/WindowContext.ts
vyorkin: Not relevant anymore

src/ui/pages/Backup/useBackupContext.ts
vyorkin: I removed `useBackupContext`

src/ui/shared/openUrl.ts
vyorkin: What if I return `Promise<number | undefined>`?
I mean:
```
getNextToActiveTabIndex(): Promise<number | undefined>
```
...
```
const index = await getNextToActiveTabIndex();
browser.tabs.create({ url: url.toString(), index });
```
^ thats how I updated this part

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
vyorkin: 1. Updated, not relevant anymore
2. Done ✅ 
3. I was like 100% sure it is a bug, will try to highlight changes like that next time so there're easier to spot & discuss

src/ui/shared/WindowParam.ts
vyorkin: Refactored, now we only have `UrlContext` + `setUrlContext`, so not relevant anymore

src/ui/pages/Backup/useBackupContext.ts
vyorkin: Updated, not relevant anymore

src/ui/pages/Backup/VerifyBackup.tsx
```diff
@@ -31,23 +40,30 @@ export function VerifyBackup({ onSuccess }: { onSuccess(): void }) {
...
-  const handleVerify = useCallback(() => {
-    setValidationError(false);
-    if (mnemonic === value.join(' ')) {
+  const verifyMutation = useMutation({
+    mutationFn: async (value: string) => {
+      invariant(recoveryPhrase, 'recoveryPhrase is missing');
+      if (recoveryPhrase !== value) {
+        throw new Error('Incorrect seed phrase');
+      }
+    },
```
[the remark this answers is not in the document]
vyorkin: Thanks, half-finished refactoring again, updated ✅

src/ui/pages/Backup/Backup.tsx
vyorkin: Ah, right, thank you ✅

src/ui/features/onboarding/Backup/Backup.tsx
vyorkin: Thanks, yeah, both `RecoveryPhrase` and `VerifyBackup` has it as a dependency of `useEffect`, updated ✅

src/ui/pages/ManageWallets/WalletGroup/WalletGroup.tsx
vyorkin: ✅

src/ui/features/onboarding/shared/SessionExpired.tsx
vyorkin: Updated to "Try creating or importing another wallet"

src/ui/features/onboarding/Onboarding.tsx
vyorkin: It should be the other way around. Fixed & moved this logic to the `SessionExpired` component to make it more readable ✅

src/ui/pages/Backup/Backup.tsx
vyorkin: ✅

src/ui/pages/Backup/useRecoveryPhrase.ts
```diff
@@ -0,0 +1,26 @@
+import { useQuery } from '@tanstack/react-query';
+import { invariant } from 'src/shared/invariant';
+import { walletPort } from 'src/ui/shared/channels';
+
+export function useRecoveryPhrase({ groupId }: { groupId?: string }) {
```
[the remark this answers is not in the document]
vyorkin: ✅ I created 2 separate hooks to make it more explicit & hopefully readable

src/ui/pages/Backup/RecoveryPhrase.tsx
vyorkin: I didn't get it, because `isSessionExpiredError` is defined like this:
```
export function isSessionExpiredError(e: unknown) {
  if (!e) {
    return false;
  }
  return getError(e).message === 'Session expired';
}
```
So it just returns `false` in case `!e` evaluates to `false`

src/ui/pages/Backup/Backup.tsx
```diff
@@ -0,0 +1,104 @@
...
+            onSuccess={() => navigate(`recovery-phrase${searchParams}`)}
+          />
+        }
+      />
+      <Route
+        path="/recovery-phrase"
+        element={
+          <RecoveryPhrase
+            groupId={groupId}
+            onNextStep={() => navigate(`verify-backup${searchParams}`)}
```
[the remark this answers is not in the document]
vyorkin: There are no anchors inside `VerifyUser` or `RecoveryPhrase` components, only buttons:
```
<Button kind="primary" disabled={loginMutation.isLoading}> 
  {loginMutation.isLoading                                 
    ? 'Checking password...'                               
    : 'Reveal Recovery Phrase'}                            
</Button>                                                  
```

```
<Button                                  
  disabled={isLoading || Boolean(error)} 
  onClick={() => {                
    emptyClipboard();                    
    onNextStep();                        
  }}                                     
>                                        
  Verify Backup                          
</Button>
```

src/ui/shared/openUrl.ts
vyorkin: This is how I updated this part ^

src/ui/pages/Backup/Backup.tsx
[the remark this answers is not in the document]
everdimension: Obviously this would mean changing button to anchor
But ok, let's not do this now

src/ui/pages/Backup/RecoveryPhrase.tsx
everdimension: Doesn't look right to rely on internal implementation. `isError` flag that comes from react-query is reliable
Also not sure it's good to handle falsy error there
Also one-letter param is not the best style

everdimension (approved):

src/ui/pages/Backup/RecoveryPhrase.tsx
vyorkin: Alright, I'll updated everywhere then, there are 10 places where we use `isSessionExpiredError`

PR #571, opened 2024-07-03
Fix SignTypeData

vyorkin:
Don't show scroll button until interpretation is fetched or if interpretation data available


vyorkin:
Closing in favour of #572

PR #578, opened 2024-07-25
Navigate directly to the backup flow if there is only one not backed up wallet group

vyorkin:


src/ui/shared/requests/useNotBackedUpWalletGroups.ts
```diff
@@ -0,0 +1,23 @@
...
+    useErrorBoundary: true,
+    select: options?.select,
+  });
+}
+
+export function useNotBackedUpWalletGroupsCount() {
+  const { data: count } = useNotBackedUpWalletGroups({
+    select: (walletGroups) => walletGroups.length,
+  });
+  return count ?? 0;
```
zerts, line 22: Can't we just return data.length?

src/ui/shared/channels.mock.ts
```diff
@@ -111,7 +111,7 @@ class WalletPortMock {
         mockRecord.walletManager.groups
       );
       return result;
-    } else if (method === 'getNoBackupCount') {
+    } else if (method === 'getNotBackedUpWalletGroups') {
```
zerts, line 114: The result of this case should not be a number any more

src/ui/shared/requests/useNotBackedUpWalletGroups.ts
zerts: Or we can just remove this hook and use the other one in Settings counter. Just add .length in that hook

src/ui/pages/Overview/BackupReminder.tsx
```diff
@@ -74,9 +82,18 @@ function BackupReminderComponent({ onDismiss }: { onDismiss: () => void }) {
...
+            onClick={(event) => {
+              handleHide();
+              if (notBackedUpWalletGroups.length === 1) {
+                openHrefInTabView(event);
+              }
+            }}
             as={UnstyledLink}
-            to="/settings"
+            to={
+              notBackedUpWalletGroups.length > 1
```
zerts, line 93: May be it is a bit safer to swap condition parts. So we will not fall into the second line if length === 0

vyorkin: yeah, I think its would be even better to check for `=== 1` ✅

zerts (approved):

src/ui/pages/Overview/BackupReminder.tsx
```diff
@@ -34,6 +37,13 @@ function BackupReminderComponent({ onDismiss }: { onDismiss: () => void }) {
     dialogRef.current?.close();
   });
 
+  const { data: walletGroups, isLoading } = useWalletGroups();
+  const notBackedUpGroups = walletGroups?.filter((group) => needsBackup(group));
```
everdimension, line 41: useMemo

everdimension (changes requested):

src/ui/shared/requests/useNotBackedUpWalletGroups.ts
vyorkin: ✅

src/ui/pages/Overview/BackupReminder.tsx
vyorkin: ✅

everdimension (approved):

PR #579, opened 2024-07-26
Register chain for lite chain support on every chain switch in the UI

zerts:


src/background/events.ts
```diff
@@ -30,6 +30,7 @@ export const emitter = createNanoEvents<{
   accountsChanged: () => void;
   chainsUpdated: () => void;
   chainChanged: (chain: Chain) => void;
+  registerChain: (chain: Chain) => void;
```
everdimension, line 33: Events names should describe something that has happened, not prescribe what should happen

PR #580, opened 2024-07-26
Handle mainnets in testmode

everdimension:


zerts:
Should we fix these titles too? 
Zerts Screenshot 2024-07-26 at 16 47@2x

zerts:
Oh, I see the behaviour. If we have extension, connected with some chain, and turn on the testmode, we will see the same chain as a connected one, but after reload the extension it changes to Ethereum

PR #583, opened 2024-08-01
Fix extension appear for external monitor

zerts:


vyorkin (approved):

src/background/NotificationWindow/createBrowserWindow.ts
```diff
@@ -66,19 +63,38 @@ export async function createBrowserWindow({
...
-    ...position,
-  });
+  let window: Windows.Window | undefined;
+  try {
+    window = await browser.windows.create({
+      focused: true,
+      url,
+      type: type === 'dialog' ? 'popup' : 'normal',
+      width,
+      height: heightValue,
```
everdimension, line 75: maybe let's extract this to something like `windowOptions` and reuse in `catch`?

everdimension (approved):
🔥

src/background/NotificationWindow/createBrowserWindow.ts
zerts: I thought about this. Let me try to do it

PR #585, opened 2024-08-06
Update network switcher layout on history tab

vyorkin:


zerts:
Need to be discussed with design first, cause this switcher was done this way on purpose previously

vyorkin:
The designers are cool with this change

zerts (approved):

everdimension (approved):

PR #586, opened 2024-08-06
Indicate gas token in asset selector

vyorkin:


zerts:
This task looks like it is a copy of what @everdimension did during the last glow up. But I don't know, why it is not in prod build anymore

zerts (approved):

everdimension (approved):

PR #587, opened 2024-08-06
Add search to connected sites

vyorkin:


src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +148,120 @@ function ConnectedSitesList({
   );
 }
 
+export function ConnectedSitesSearch({
```
everdimension, line 151: Why are we exporting?

src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +148,120 @@ function ConnectedSitesList({
...
+  onChange(value: string): void;
+}) {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const debouncedHandleChange = useDebouncedCallback(onChange, 300);
+
+  useLayoutEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.value = value || '';
+    }
+  }, [value]);
```
everdimension, line 165: What is this for?

zerts: This is probably for Reset button. To make this input both uncontrolled and controlled

src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +148,120 @@ function ConnectedSitesList({
...
+
+  return (
+    <SearchInput
+      ref={inputRef}
+      boxHeight={40}
+      type="search"
+      placeholder="Search"
+      defaultValue={value}
```
zerts, line 173: Can we make this input fully controllable?

src/ui/pages/ConnectedSites/ConnectedSites.tsx
vyorkin: Looks like I wanted to keep it in a separate file initially but then changed my mind, removed `export` ✅

src/ui/pages/ConnectedSites/ConnectedSites.tsx
vyorkin: The approach is directly inspired by the search on the history tab: `useLayoutEffect` ensures that `inputRef`'s value stays in sync with the `value` prop (particularly when it gets reset).

Now I've refactored `ConnectedSitesSearch` to be fully controlled and moved `inputRef` outside

src/ui/pages/ConnectedSites/ConnectedSites.tsx
vyorkin: Refactored, please check out the updated version ✅

src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +149,131 @@ function ConnectedSitesList({
...
     useErrorBoundary: true,
     suspense: true,
   });
+  const [searchQuery, setSearchQuery] = useState<string | undefined>();
+  const searchInputRef = useRef<InputHandle | null>(null);
 
-  if (connectedSitesQuery.isLoading) {
-    return <p>Loading qweasd...</p>;
-  }
+  const filteredConnectedSites = useMemo(() => {
```
everdimension, line 238: I think a better name is something like "itemsToDisplay" because it's not always a filtered array but also the original array

src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +149,131 @@ function ConnectedSitesList({
...
+    data: allConnectedSites,
+    isLoading,
+    ...connectedSitesQuery
+  } = useQuery({
     queryKey: ['getPermissionsWithWallets'],
     queryFn: getPermissionsWithWallets,
     useErrorBoundary: true,
     suspense: true,
   });
+  const [searchQuery, setSearchQuery] = useState<string | undefined>();
```
everdimension, line 235:
```suggestion
  const [searchQuery, setSearchQuery] = useState<string | null>(null);
```
But also why not an empty string?

src/ui/pages/ConnectedSites/ConnectedSites.tsx
```diff
@@ -143,38 +149,131 @@ function ConnectedSitesList({
...
-          </UIText>
-        </FillView>
+        <EmptyView
+          hasConnectedSites={Boolean(allConnectedSites?.length)}
+          hasFilters={Boolean(searchQuery)}
+          onReset={() => {
+            if (searchInputRef.current) {
+              searchInputRef.current.setValue('');
+            }
+            setSearchQuery(undefined);
```
everdimension, line 274: Trying to follow this practice and avoid persisting "undefined" values: https://x.com/sebmarkbage/status/1774082542372520441

everdimension (approved):

src/ui/pages/ConnectedSites/ConnectedSites.tsx
vyorkin: ✅

PR #590, opened 2024-08-14
Fix ethers v5 error when populating transaction for some chains

zerts:


src/background/Wallet/Wallet.ts
```diff
@@ -131,6 +131,36 @@ async function prepareNonce<T extends { nonce?: BigNumberish; from?: string }>(
...
+      type: 2,
+    };
+  }
+  if (transaction.gasPrice != null) {
+    return {
+      ...transaction,
+      type: 0,
+    };
+  }
+  // we weren't able to populate transaction's type
```
vyorkin, line 160: Should we throw a descriptive error in this case? Just thinking

vyorkin: Ah, nvm, it doesn't make sense to assume it would fail

vyorkin (approved):

src/background/Wallet/Wallet.ts
```diff
@@ -1065,7 +1095,11 @@ export class Wallet {
     const txWithFee = await prepareGasAndNetworkFee(prepared, networks, {
       source: mode === 'testnet' ? 'testnet' : 'mainnet',
     });
-    const transaction = await prepareNonce(txWithFee, networks, chain);
+    // todo: remove after update to ethers v6
```
everdimension, line 1098: Let's write more clearly which line exactly should be removed, otherwise it'll be hard to remember

src/background/Wallet/Wallet.ts
```diff
@@ -131,6 +131,36 @@ async function prepareNonce<T extends { nonce?: BigNumberish; from?: string }>(
   }
 }
 
+function prepareTransactionType<
```
everdimension, line 134: I'd comment here too that this is a temporary helper

everdimension (approved):
When merging, let's rephrase the commit message to deliver intent: this is a fix for ethers internals when signTransaction is called.

Populating type is not the intent but a hack that happens to solve the problem :)

src/background/Wallet/Wallet.ts
```diff
@@ -1065,7 +1097,11 @@ export class Wallet {
     const txWithFee = await prepareGasAndNetworkFee(prepared, networks, {
       source: mode === 'testnet' ? 'testnet' : 'mainnet',
     });
-    const transaction = await prepareNonce(txWithFee, networks, chain);
+    // TODO: remove `prepareTransactionType` helper after update to ethers v6
+    // ethers v5 throws error inside `getFeeData` for some chains with too big totalDifficulty param
+    // can be reproduced with https://chainlist.org/chain/30732
+    const txWithFeeAndType = await prepareTransactionType(txWithFee);
```
everdimension, line 1103: Why are we awaiting? It seems to be sync

zerts: Ah, really! Good catch!

PR #592, opened 2024-08-15
Use supports_simulations flag

everdimension:
Fix NetworkConfig type stored in storage

### DIFF
--- a/src/modules/networks/Networks.ts
+++ b/src/modules/networks/Networks.ts
@@ -76,10 +76,22 @@ export class Networks {
   private ethereumChainConfigs: EthereumChainConfig[];
   private visitedChains: Set<string>;
 
-  static getChainId(network: NetworkConfig) {
-    return network.standard === 'eip155'
-      ? normalizeChainId(network.specification.eip155.id)
-      : null;
+  static getChainId(
+    network: Partial<
+      Pick<NetworkConfig, 'standard' | 'specification' | 'external_id'>
+    >
+  ) {
+    if ('standard' in network && network.specification) {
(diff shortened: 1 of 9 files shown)


src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
```diff
@@ -106,7 +106,7 @@ export function TransactionSimulation({
   const txInterpretQuery = useQuery({
     enabled: FEATURE_PAYMASTER_ENABLED
       ? shouldInterpretTransaction({ network, eligibilityQuery })
```
zerts, line 108: Should we add check for `supports_simulations` when `FEATURE_PAYMASTER_ENABLED=on`?

src/modules/networks/Networks.ts
```diff
@@ -76,10 +76,22 @@ export class Networks {
...
+  static getChainId(
+    network: Partial<
+      Pick<NetworkConfig, 'standard' | 'specification' | 'external_id'>
+    >
+  ) {
+    if ('standard' in network && network.specification) {
+      return network.standard === 'eip155'
+        ? normalizeChainId(network.specification.eip155.id)
+        : null;
+    } else if (network.external_id) {
```
zerts, line 88: Why do we need this condition? external_id is deprecated and shouldn't be persisted anywhere any more.

zerts (approved):

src/modules/networks/Networks.ts
everdimension: This method is called by code that deals with the persisted network configs

PR #594, opened 2024-08-19
e2e tests (playwright)

vyorkin:


src/ui/index.tsx
```diff
@@ -70,7 +70,11 @@ async function initializeUI({
   try {
     await registerServiceWorker();
     initializeChannels();
-    await maybeOpenOnboarding();
+
+    if (process.env.NODE_ENV !== 'test') {
```
zerts, line 74: I'm not sure that we need such checks in the code. Or this will create a different behaviour between prod build and test run. May be it is better to provide a predefined config for test like we have some wallets in it

playwright.config.ts
```diff
@@ -0,0 +1,31 @@
...
+  // 1 minute, default is 30 seconds
+  timeout: 60 * 1000,
+  // Run tests in files in parallel
+  fullyParallel: true,
+  // Fail the build on CI if you accidentally left test.only in the source code
+  forbidOnly: !!process.env.CI,
+  // Retry on CI only
+  retries: process.env.CI ? 2 : 0,
+  // Opt out of parallel tests on CI
+  workers: process.env.CI ? 1 : undefined,
```
zerts, line 15: How does this aligns with `fullyParallel: true`?

e2e/test.ts
```diff
@@ -0,0 +1,115 @@
...
+      });
+      if (!popupFileName) {
+        throw Error('popup file not found');
+      }
+      await use(`chrome-extension://${extensionId}/${popupFileName}`);
+    },
+  });
+}
+
+export async function takeScreenshot(page: Page, screenshotPath: string) {
```
zerts, line 113: This function is not used anywhere. Do we need it?

playwright.config.ts
vyorkin: as I understand:
`fullyParallel: true` enables parallel tests _inside_ a single file
`workers: process.env.CI ? 1 : undefined` ensures only one file runs at a time in CI.
so it kinda limits file-level concurrency in CI but keeps parallelism within file

.github/workflows/run-tests.yml
```diff
@@ -18,20 +34,26 @@ jobs:
...
+        run: |
+          npm run test:unit
+
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps
+
+      - name: Run Playwright tests
+        run: xvfb-run npx playwright test
+
+      - name: Upload Playwright report
```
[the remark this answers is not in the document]
vyorkin: Ah, I renamed the folder to `e2e-report`

src/ui/index.tsx
vyorkin: Actually I noticed that it doesn't affect that single test, so I just removed this condition

e2e/utils/wallet.ts
```diff
@@ -0,0 +1,15 @@
+import { entropyToMnemonic } from '@ethersproject/hdnode';
+import { Wallet, randomBytes } from 'ethers';
+
+export function generateRandomRecoveryPhrase() {
+  const entropy = randomBytes(16);
+  return entropyToMnemonic(entropy);
+}
```
everdimension, line 7: what about
```
import { Wallet as EthersV5Wallet } from '@ethersproject/wallet';

export function generateRandomRecoveryPhrase() {
  return EthersV5Wallet.createRandom().mnemonic.phrase;
}
```

Is this the same?

e2e/utils/wallet.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function generateRandomWallet() {
```
everdimension, line 9: I think this should conform to wallets that are being created in `src/shared/wallet/create.ts`:
```suggestion
export function generateRandomWallet(): BareWallet {
```

e2e/utils/wallet.ts
```diff
@@ -0,0 +1,15 @@
+import { entropyToMnemonic } from '@ethersproject/hdnode';
+import { Wallet, randomBytes } from 'ethers';
```
everdimension, line 2: prefer `Wallet` from `'@ethersproject/wallet'`?

src/ui/shared/truncateAddress.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const address = '7358B726830A2E222f9b139E90483A37142bcBE5';
+    const expected = `7358B7${ellipsis}2bcBE5`;
+    expect(truncateAddress(address)).toBe(expected);
+  });
+
+  test('custom padding', () => {
+    const address = '0x7358B726830A2E222f9b139E90483A37142bcBE5';
+    const expected = `0x7358${ellipsis}cBE5`;
+    expect(truncateAddress(address, 4)).toBe(expected);
+  });
```
everdimension, line 21: Let's test at least two different padding values?

e2e/tests/test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export function createExtensionTest(testConfig?: Config) {
+  return baseTest.extend<TestContext & Fixtures & Options>({
+    // eslint-disable-next-line no-empty-pattern
+    context: async ({}, use) => {
+      // Determine headless mode:
+      // - in CI: always run headless
+      // - in dev: use RUN_HEADED_IN_DEV value unless overridden by testConfig
+      const headless =
+        !process.env.CI && (testConfig?.headless ?? !RUN_HEADED_IN_DEV);
```
everdimension, line 144: Contradicts comment

e2e/utils/wallet.ts
vyorkin: I think so! thanks ✅

e2e/utils/wallet.ts
vyorkin: updated ✅

e2e/utils/wallet.ts
vyorkin: updated ✅

e2e/tests/test.ts
vyorkin: updated

src/ui/shared/truncateAddress.test.ts
vyorkin: added 2 more tests with 4 and 8 padding

src/ui/pages/GetStarted/GetStarted.tsx
```diff
@@ -266,7 +266,6 @@ function WalletGroupSelect() {
         {title}{' '}
         <span
           title="Each group contains wallets that are associated with same recovery phrase, stored locally on your device. Zerion does not have access to this data.
-
```
zerts, line 269: Do we need this change?

src/ui/index.tsx
```diff
@@ -70,7 +70,9 @@ async function initializeUI({
   try {
     await registerServiceWorker();
     initializeChannels();
+
```
zerts, line 73: Do we need this change?

src/background/index.ts
```diff
@@ -212,7 +212,6 @@ initialize().then((values) => {
 browser.runtime.onInstalled.addListener(({ reason, previousVersion }) => {
   runtimeStore.handleInstalledEvent({ reason, previousVersion });
   if (reason === 'install') {
-    userLifecycleStore.handleRuntimeInstalledEvent();
```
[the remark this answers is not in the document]
vyorkin: not ok :)

src/ui/index.tsx
vyorkin: ah, not really, just my habits to edit code while reading... updated ✅

zerts (approved):

src/ui/pages/GetStarted/GetStarted.tsx
vyorkin: reverted ✅

everdimension (approved):

PR #595, opened 2024-08-20
Automatically enable testnet mode

everdimension:
Fix testmode edge cases and update flows


src/ui/components/EmptyView/EmptyView.tsx
```diff
@@ -21,3 +22,36 @@ export function EmptyView({
     </FillView>
   );
 }
+
+export function EmptyView2(
```
vyorkin, line 26: 😄

src/ui/pages/Settings/Settings.tsx
```diff
@@ -303,7 +303,23 @@ function DeveloperTools() {
...
+                  style={{
+                    padding: '2px 4px',
+                    background: 'var(--neutral-200)',
+                    borderRadius: 4,
+                    color: 'var(--black)',
+                  }}
+                >
+                  T
+                </kbd>{' '}
+                shortcut to switch quickly between mode
```
vyorkin, line 320:
```suggestion
                shortcut to switch quickly between modes
```
I believe "between modes" is correct plurality

src/ui/pages/Settings/Settings.tsx
```diff
@@ -303,7 +303,23 @@ function DeveloperTools() {
                 testnetMode: event.target.checked ? { on: true } : null,
               });
             }}
-            detailText="Enables viewing and interacting with test networks"
+            detailText={
+              <div>
+                Enables viewing and interacting with test networks <br />
```
vyorkin, line 308:
```suggestion
                Enables viewing and interacting with test networks. <br />
```
Not sure about adding the period here, looks like it was omitted intentionally

vyorkin (approved):

src/ui/components/EmptyView/EmptyView.tsx
everdimension: I think current name conveys better that we plan to deprecate the first one

src/background/Wallet/Wallet.ts
```diff
@@ -1538,7 +1543,7 @@ export class Wallet {
     const network = await fetchNetworkByChainId({
       preferences: Model.getPreferences(this.record),
       chainId,
-      apiEnv: 'current',
```
zerts, line 1541: Why do we need this change?

src/ui/pages/Overview/Overview.tsx
```diff
@@ -297,6 +354,50 @@ function OverviewComponent() {
...
+      <FillView>
+        <EmptyView2
+          title="Wrong Environment"
+          message={
+            <div>
+              {preferences?.testnetMode?.on ? (
+                <UnstyledButton
+                  className="underline hover:no-underline"
+                  onClick={() => {
+                    setPreferences({ testnetMode: null });
```
zerts, line 378: should we make it `{ on: false }` in this case?

zerts (approved):

src/background/Wallet/Wallet.ts
everdimension: Case on uniswap: we can be connected to a mainnet network, and dapp gets balance through the wallet rpc methods. But if in the wallet we have testmode enabled, then here we won't be able to get data for the mainnet network to which the dapp is currently connected, and the dapp's UI will appear to be broken (unable to display balance)

src/ui/pages/Overview/Overview.tsx
everdimension: My reasoning here is that this button is helpful to those who do not need testmode and ended up in it accidentally. Those who need test toggle to remain can use either the `T` shortcut or the toggle at the bottom of the page

PR #596, opened 2024-08-21
PoC: Sidepanel

everdimension:
- [x] When sidepanel is open, react to updated tab origin and to "chain changed" event
- [x] Adapt layout for sidepanel
- [ ] What if dapp request comes NOT from the current (active) tab? Handle this case


zerts:
In Arc if you choose "Prefer Sidepanel", extension goes into broken state. Cause sidepanel is not supported, so nothing opens.

zerts:
If you prefer sidebar and sign/send/connect screen opens, you lose context and route where you were before in extension.

zerts:
If I prefer sidebar, close it and try to sign transaction, sidebar doesn't open on this request -> I need to do it manually.

src/ui/shared/routing/helpers.ts
```diff
@@ -0,0 +1,5 @@
+import { emitter } from '../events';
+
+export function navigateProgrammatically(params: { pathname: string }) {
+  emitter.emit('navigationRequest', params);
```
zerts, line 4: why do we need this wrapper over `navigate`?

src/ui/shared/events.ts
```diff
@@ -14,6 +14,10 @@ export const emitter = createNanoEvents<{
   errorScreenView: (data: { message: string; location: string }) => void;
   loaderScreenView: (data: { location: string; duration: number }) => void;
   reloadExtension: () => void;
+  navigationRequest: (params: { pathname: string }) => void;
+  // this event means that some dapp-related data is updated (e.g. current account, chain)
+  ethereumEvent: () => void;
+  'sidepanel/activeTabUpdated': () => void;
```
zerts, line 20: Should this event be called just `activeTabUpdated`?

zerts:
After choosing "Open sidepanel once" and closing it, dapp requests don't open popup

src/shared/sidepanel/BrowserState.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor() {
+    this.getInitialWindow();
+    this.addListeners();
+  }
+
+  async getInitialWindow() {
+    const currentWindow = await browser.windows.getCurrent();
+    this.currentWindowId = currentWindow.id;
+    this.initialWindowId = currentWindow.id;
```
zerts, line 15: The fact that getter changing the sate of the class seems a bit confusing

src/shared/sidepanel/sidepanel-support.ts
```diff
@@ -0,0 +1,3 @@
+export function isSidepanelSupported() {
+  return globalThis.chrome && 'sidePanel' in chrome;
```
zerts, line 2: Hm, interesting, why this didn't prevent sidebar controls from appearing in Arc?

src/shared/sidepanel/sidepanel-commands.background.ts
```diff
@@ -0,0 +1,56 @@
...
+         * the request, so it's ok. But if it is open, it will receive the request asyncronously, and therefore
+         * it will react to it AFTER we call the {sidePanel.open} method later, so the {open} method will
+         * not work, and that's what we need.
+         */
+        closeSidepanel({ windowId: currentWindowId ?? null });
+      } catch {
+        // it's ok, it might've not been open at all
+      }
+      const url = getSidepanelUrl();
+      if (!currentWindowId) {
```
zerts, line 30: Should we move this condition to the beginning of this function? To avoid closing sidepanel without opening it next

zerts (approved):

src/ui/shared/routing/helpers.ts
everdimension: because `navigate` works only in context of react router
we should migrate to creating a `router` instance later: https://reactrouter.com/en/main/upgrading/v6-data

src/ui/shared/events.ts
everdimension: I don't know :)

src/shared/sidepanel/sidepanel-support.ts
everdimension: Yeah, look like a bug in Arc

src/shared/sidepanel/sidepanel-commands.background.ts
everdimension: Hm, I'm thinking that if user pressed a shortcut and sidepanel is opened and it's possible to close it, it should be closed. That way the user interaction won't seem broken

everdimension:
Closed in favor of https://github.com/zeriontech/zerion-wallet-extension/pull/617

PR #602, opened 2024-08-30
Animation for Login View

everdimension:


src/ui/pages/Login/login-animation-lottie-v3.json
vyorkin, line 1: 10 mb 😱

everdimension: yeeeah.... newer file is 15 😱 
I asked to look into it :)

src/background/Wallet/Wallet.ts
```diff
@@ -223,9 +224,13 @@ export class Wallet {
         emitter.emit('globalPreferencesChange', state, prevState);
       })
     );
+    // todo: group into "syncWithExternalStores" or "updateExternalStores"?
     this.disposer.add(
       this.walletStore.on('change', this.updateChainConfigStore.bind(this))
     );
+    this.disposer.add(
+      this.walletStore.on('change', this.updateLastAddress.bind(this))
+    );
```
vyorkin, line 233: (I would postpone grouping/extracting until we need to have the same calls anywhere else)

src/background/Wallet/Wallet.ts
```diff
@@ -799,6 +805,29 @@ export class Wallet {
...
+    }
+  }
+
+  async getLastUsedAddress({
+    context,
+    params: { userId },
+  }: WalletMethodParams<{ userId: string }>) {
+    this.verifyInternalOrigin(context);
+    const state = lastUsedAddress.getState();
+    if (state && state?.walletModelId === userId) {
```
vyorkin, line 824: Maybe I don't get the idea but from the first sight it is not clear why are we comparing `walletModelId` to `userId`

vyorkin (approved):

src/background/Wallet/Wallet.ts
everdimension: > until we need to have the same two calls anywhere else

But we do :) On lines 255-256, that's the reason for the suggestion

src/background/user-activity.ts
```diff
@@ -45,3 +46,10 @@ export async function handleAlarm(alarm: browser.Alarms.Alarm) {
     }
   }
 }
+
+// TODO: use LocallyEncrypted value for {{ address: LocallyEncrypted } | undefined}
+type State = { address: string; walletModelId: string } | null; // Record< | undefined>;
+class LastUsedAddress extends PersistentStore<State> {}
+
+const STORAGE_KEY = 'last-selected-address';
+export const lastUsedAddress = new LastUsedAddress(null, STORAGE_KEY);
```
zerts, line 55: What about `lastUsedAddressStore` ?

src/background/user-activity.ts
```diff
@@ -45,3 +46,10 @@ export async function handleAlarm(alarm: browser.Alarms.Alarm) {
     }
   }
 }
+
+// TODO: use LocallyEncrypted value for {{ address: LocallyEncrypted } | undefined}
+type State = { address: string; walletModelId: string } | null; // Record< | undefined>;
+class LastUsedAddress extends PersistentStore<State> {}
+
+const STORAGE_KEY = 'last-selected-address';
```
zerts, line 54: 'last-used-address'?
Not critical, just interesting, why key is different from the name from the beginning?

src/ui/components/WalletAvatar/AvatarIcon/AvatarIcon.tsx
```diff
@@ -10,17 +10,27 @@ export function AvatarIcon({
...
+    if (!hasMediaContent) {
+      // means we're loading blockie and it's "ready" immediately
+      onReady?.();
+    }
+  }, [hasMediaContent, onReady]);
   return (
     <div className={s.root}>
       <div className={active ? s.activeIndicatorClip : undefined}>
-        {nft?.metadata?.content ? (
+        {nft?.metadata && mediaContent ? (
```
zerts, line 33: Sad that ts can't understand the narrowing here :(

src/background/Wallet/Wallet.ts
```diff
@@ -799,6 +805,29 @@ export class Wallet {
     chainConfigStore.setDefiSdkClient(client);
   }
 
+  private updateLastAddress() {
+    const currentAddress = this.readCurrentAddress();
+    if (this.id && currentAddress) {
+      lastUsedAddress.setState({
+        address: currentAddress,
+        walletModelId: this.id,
```
zerts, line 813: Is there a case when we request this address with another userId?
As I understand, it can happen after creating new user which should lead to clearing the whole data anyway.

src/ui/pages/Login/LayersAnimationLottie.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+    const item = lottieWeb.loadAnimation({
+      container,
+      animationData,
+      loop: false,
+      autoplay,
+      renderer: 'svg',
+    });
+    function loopOnCompleted() {
+      const FRAME_NUMBER_TO_START_LOOP = 39;
```
zerts, line 47: Probably the only line specific to this particular case. Everything else looks like can be reused easily.

src/ui/pages/Login/LayersAnimationLottie.tsx
```diff
@@ -0,0 +1,118 @@
...
+      item.removeEventListener('complete', loopOnCompleted)
+    );
+    this.options.onAnimationReady?.();
+    this.animationItem = item;
+  }
+
+  destroy() {
+    this.disposables.forEach((dispose) => dispose());
+    this.animationItem?.destroy();
+    this.animationItem = null;
```
zerts, line 61: Hah :)
Is this a GC workaround?

src/ui/pages/Login/Login.tsx
```diff
@@ -12,10 +19,78 @@ import { VStack } from 'src/ui/ui-kit/VStack';
...
+            onAnimationReady={() => setLottieIsReady(true)}
+          />
+        </DelayedRender>
+      </div>
+      <Spacer height={130} />
+      <div style={{ display: 'flex', justifyContent: 'center' }}>
+        <ZStack style={{ placeItems: 'center' }}>
+          <ZerionLogo style={{ width: 54, height: 54 }} />
+          {address ? (
+            <div
```
zerts, line 66: Why not use react-spring here?

zerts (approved):

src/background/user-activity.ts
everdimension: missed after refactoring

src/ui/pages/Login/Login.tsx
everdimension: Trying to experiment with new CSS features

src/background/Wallet/Wallet.ts
everdimension: I'm just following the idea that current user id and wallet model id must match

src/ui/pages/Login/LayersAnimationLottie.tsx
everdimension: Nope, just returning state to its initial value

src/ui/pages/Login/LayersAnimationLottie.tsx
everdimension: this and the path to the lottie file

everdimension: And the looping logic :)

PR #604, opened 2024-08-30
Fix build

vyorkin:


everdimension:
Thanks! Fixed in https://github.com/zeriontech/zerion-wallet-extension/pull/602, closing as this is a research PR

PR #607, opened 2024-09-04
[Testnet Mode]: Use testnet ZPI for paymaster requests

everdimension:


src/modules/zerion-api/zerion-api.ts
```diff
@@ -134,8 +132,44 @@ export class ZerionAPI {
...
+      gas: string;
+      gasPerPubdataByte: string;
+      maxFee: string;
+      maxPriorityFee: string;
+      value: string;
+      data: string;
+    },
+    { source }: BackendSourceParams
+  ) {
+    interface PaymasterParamsResponse {
```
zerts, line 155: What do you think about defining types outside of this file?

src/modules/zerion-api/zerion-api.ts
```diff
@@ -43,20 +43,23 @@ function getZpiHeaders() {
   };
 }
 
+type UrlInput = { endpoint: string } | { url: string | URL };
```
zerts, line 46: Do we need this endpoint case? I can't find it in the code

zerts (approved):

src/modules/zerion-api/zerion-api.ts
everdimension: Updated other methods to use this

zerts (approved):

PR #608, opened 2024-09-05
Put chain state in URL for Overview tab

everdimension:
So that going "back" from other routes
returns the user to selected chain

### DIFF
--- a/src/ui/pages/Overview/Overview.tsx
+++ b/src/ui/pages/Overview/Overview.tsx
@@ -1,5 +1,5 @@
-import React, { useEffect, useMemo, useState } from 'react';
-import { Route, Routes, useLocation } from 'react-router-dom';
+import React, { useEffect, useMemo, useRef } from 'react';
+import { Route, Routes, useLocation, useSearchParams } from 'react-router-dom';
 import { useAddressPortfolioDecomposition } from 'defi-sdk';
 import { RenderArea } from 'react-area';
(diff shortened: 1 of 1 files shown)


src/ui/pages/Overview/Overview.tsx
```diff
@@ -297,7 +298,12 @@ function OverviewComponent() {
   });
   const isReadonlyGroup =
     walletGroup && isReadonlyContainer(walletGroup.walletContainer);
-  const [filterChain, setFilterChain] = useState<string | null>(null);
+  const [searchParams, setSearchParams] = useSearchParams();
+  const filterChain = searchParams.get('chain') || null;
```
zerts, line 302: May be it is gonna be more reliable to name this `filterChain`?

src/ui/pages/Overview/Overview.tsx
```diff
@@ -398,6 +408,24 @@ function OverviewComponent() {
     </CenteredFillViewportView>
   );
 
+  /**
+   * Creates href such that "chain" search-param is preserved between
+   * tabs, but clicking on current tab resets searchParams
```
zerts, line 413: Right now clicking on selected tab scrolls you to the top. Is it ok to combine 2 actions here (scroll + filter reset)?

zerts: We have `All networks` button for this quick filter reset

zerts (approved):

PR #611, opened 2024-09-09
Fix network-store load calls order

zerts:


src/modules/networks/networks-store.ts
```diff
@@ -148,10 +149,14 @@ export class NetworksStore extends Store<State> {
         ?.filter((config) => !fulfilledNetworkIdSet.has(config.id))
         .map((config) => toNetworkConfig(config.value, config.id)) || [];
 
+    this.isReady = true;
     return this.updateNetworks();
   }
 
   private async fetchNetworkById(chainId: ChainId) {
+    if (!this.isReady) {
+      await this.load();
```
everdimension, line 158: why not in the `finally` block of the `load` method?

src/modules/networks/networks-store.ts
```diff
@@ -148,10 +149,14 @@ export class NetworksStore extends Store<State> {
         ?.filter((config) => !fulfilledNetworkIdSet.has(config.id))
         .map((config) => toNetworkConfig(config.value, config.id)) || [];
 
+    this.isReady = true;
```
everdimension, line 152: why not in the finally block of the load method?

everdimension (approved):

PR #616, opened 2024-09-22
Sanitize PR number from artifact

AdnaneKhan:
Technically the PR number could be anything from the first workflow if from a PR and therefore could be used to inject code. This fixes it.

### DIFF
--- a/.github/workflows/pr_attach_build.yml
+++ b/.github/workflows/pr_attach_build.yml
@@ -65,9 +65,11 @@ jobs:
       - name: Attach build artifacts to the PR thread
         if: ${{ github.event.workflow_run.conclusion == 'success' }}
         uses: actions/github-script@v3
+        env:
+          PR_NUMBER:  "${{ steps.get-build-artifacts.outputs.pr_number }}"
         with:
           script: |
-            const pr_number = "${{ steps.get-build-artifacts.outputs.pr_number }}";
+            const pr_number = process.env.PR_NUMBER;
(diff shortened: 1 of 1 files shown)


everdimension:
Can you please give more details about what problem this update is solving?

AdnaneKhan:
> Can you please give more details about what problem this update is solving?

The artifact uploaded on a `pull_request` workflow can be modified so that it contains a PR number that is in fact a script injection payload. You can see in the previous step it just reads the PR number from a file and sets it to the output variable.

```
   pr_number=$(cat ./pr_number)
          build_artifact_id=$(curl -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}/artifacts" | jq ".artifacts[1].id")
          # We can not use "url" or "archive_download_url" here,
          # hence we need to figure out the artifact's URL "manually"
          build_download_url=${repository_url}/suites/${check_suite_id}/artifacts/${build_artifact_id}
          set +x
          echo "pr_number=$pr_number" >> $GITHUB_OUTPUT
          echo "build_download_url=$build_download_url" >> $GITHUB_OUTPUT
```

So if someone modified the `pr.yml` workflow in their fork pull request to upload a malicious PR number that contained javascript code, they could conduct a poisoned pipeline execution attack.

## Reference 
* https://securitylab.github.com/resources/github-actions-untrusted-input/

PR #617, opened 2024-09-23
Feature/sidepanel support and flow [WLT-5590]

everdimension:


src/ui/style/global.module.css
```diff
@@ -39,6 +39,7 @@ body {
    */
   display: grid;
   --body-width: 425px;
+  --body-max-width: 800px;
```
zerts, line 42: should we add this to popup mode too?
also login page layout is a bit broken on wide screen

src/ui/shared/channels.ts
```diff
@@ -42,18 +46,33 @@ export const sessionCacheService = new PortMessageChannel({
 }) as RPCPort<SessionCacheService>;
 
 class WindowPort extends PortMessageChannel {
+  static maybeRestoreRouteForSidepanel() {
+    if (urlContext.windowType === 'sidepanel') {
+      // TODO: navigate to location that user was on before opening the request view?
```
zerts, line 51: Should we implement this before release?

src/background/Wallet/Wallet.ts
```diff
@@ -1631,6 +1639,10 @@ class PublicController {
     origin: string,
     props: NotificationWindowProps<T>
   ) {
+    if (debugValue && process.env.NODE_ENV === 'development') {
```
zerts, line 1642: Do we need to keep this?

src/shared/UrlContext.ts
```diff
@@ -6,12 +6,19 @@ import type {
...
+}
+
 function getUrlContext(): UrlContext {
   const params = new URL(window.location.href).searchParams;
   return {
     appMode: (params.get(UrlContextParam.appMode) as AppMode) || 'wallet',
-    windowType:
-      (params.get(UrlContextParam.windowType) as WindowType) || 'popup',
+    windowType: getWindowType(params),
+    // (params.get(UrlContextParam.windowType) as WindowType) || 'popup',
```
zerts, line 21: delete?

src/shared/UrlContext.ts
```diff
@@ -6,12 +6,19 @@ import type {
 } from './types/UrlContext';
 import { UrlContextParam } from './types/UrlContext';
 
+function getWindowType(params: URLSearchParams): WindowType {
+  if (window.location.pathname.startsWith('/sidepanel')) {
```
zerts, line 10: Do we use this case somewhere?

src/shared/sidepanel/types.ts
```diff
@@ -0,0 +1,21 @@
+import { isObj } from '../isObj';
+
+interface IsOpenRequest {
```
zerts, line 3: `IsSidepanelOpenRequest` ?

src/shared/sidepanel/types.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export type SidepanelMessageRequest = {
+  payload: IsOpenRequest | CloseSidepanelRequest;
+};
+
+export function isSidepanelMessageRequest(
+  x: unknown
+): x is SidepanelMessageRequest {
+  return isObj(x) && isObj(x.payload) && 'method' in x.payload;
```
zerts, line 20: should we check here that x.payload.method is one of options 'is-sidepanel-open'/'close-sidepanel'?

src/shared/sidepanel/sidepanel-support.ts
```diff
@@ -0,0 +1,3 @@
+export function isSidepanelSupported() {
+  return globalThis.chrome && 'sidePanel' in chrome;
```
zerts, line 2: `&& chrome.sidePanel != null`

src/shared/sidepanel/sidepanel-apis.ts
```diff
@@ -0,0 +1,52 @@
...
+    invariant(tab, 'Active tab not found');
+    invariant(tab.id, 'Active tab id not set');
+    return { tabId: tab.id };
+  }
+}
+
+export async function openSidePanel({
+  pathname,
+  searchParams,
+  openOptions,
```
zerts, line 26: Do we need this param?

src/shared/sidepanel/sidepanel-apis.ts
```diff
@@ -0,0 +1,52 @@
...
+  searchParams: URLSearchParams | null;
+  openOptions?: chrome.sidePanel.OpenOptions;
+}) {
+  const url = getSidepanelUrl();
+  // TODO: support templateType=sidepanel
+  // url.searchParams.append('templateType', 'dialog');
+  // const id = nanoid();
+  // searchParams.append('windowId', String(id));
+  // We use HashRouter currently
+  if (searchParams) {
```
zerts, line 38: As I see there is only one case of this function call and searchParams are always null

src/ui/style/global.module.css
everdimension: Popup can't be this wide

src/background/Wallet/Wallet.ts
everdimension: Nope

src/shared/UrlContext.ts
everdimension: yes, we use `urlContext.windowType` in a lot of places :)

src/shared/sidepanel/types.ts
everdimension: This would make it harder to add more methods
Good case for `zod` btw

src/shared/sidepanel/sidepanel-support.ts
everdimension: This is a canonical way to check for browser apis
There's no reason for the property to exist and be nullish, I think. And we're not checking deeper anyway

src/shared/sidepanel/sidepanel-apis.ts
everdimension: This is still a reasonable api I think

PR #622, opened 2024-10-07
Invitation Center

vyorkin:


src/ui/pages/Invite/SuccessDialog/GradientBorder.tsx
vyorkin, line 1: `GradientBroder` is taken from the webapp as is

PR #623, opened 2024-10-07
Update client event properties

everdimension:
For metabase events:
* remove user_id
* send wallet_address
* send correct wallet_provider

### DIFF
--- a/src/shared/analytics/analytics.background.ts
+++ b/src/shared/analytics/analytics.background.ts
@@ -40,21 +40,36 @@ function trackAppEvents({ account }: { account: Account }) {
 
   const getProvider = (address: string) =>
     getProviderForMetabase(queryWalletProvider(account, address));
+  const getCurrentAddress = () =>
+    account.getCurrentWallet().readCurrentAddress();
 
-  const createParams: typeof createBaseParams = (params) => {
-    const getUserId = () => account.getUser()?.id;
-    return createBaseParams({ ...params, userId: getUserId() });
+  const createParams: typeof createBaseParams = (eventParams) => {
+    const params: typeof eventParams & {
+      wallet_address?: string | null;
+      wallet_provider?: string | null;
+    } = eventParams;
+
+    if (!params.wallet_address) {
+      params.wallet_address = getCurrentAddress();
+    }
+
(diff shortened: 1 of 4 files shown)


src/shared/analytics/analytics.client.ts
```diff
@@ -1,85 +1,107 @@
...
+  getWalletProvider,
+}: {
+  getWalletProvider: (address: string) => Promise<string>;
+}) {
+  async function createWalletProperties() {
+    const address = readCachedCurrentAddress() ?? null;
+    const wallet_provider = address
+      ? await requestWithTimeout(() => getWalletProvider(address))
+      : null;
+    return { wallet_address: address?.toLowerCase(), wallet_provider } as const;
```
zerts, line 35: We should filter all `null` values from the final object for Mixpanel. 
Should we filter wallet_provider: null here?
I don't see that we send this to mixpanel, but may be I miss something

zerts: Loos like all good!

src/shared/analytics/analytics.background.ts
```diff
@@ -40,21 +40,36 @@ function trackAppEvents({ account }: { account: Account }) {
...
-    const getUserId = () => account.getUser()?.id;
-    return createBaseParams({ ...params, userId: getUserId() });
+  const createParams: typeof createBaseParams = (eventParams) => {
+    const params: typeof eventParams & {
+      wallet_address?: string | null;
+      wallet_provider?: string | null;
+    } = eventParams;
+
+    if (!params.wallet_address) {
+      params.wallet_address = getCurrentAddress();
```
zerts, line 53: We shouldn't send null here. I see that types for getCurrentAddress allow string | null

everdimension: We have been sending it already
I'll confirm with Anya

everdimension: For mixpanel we're getting rid of all nullish params before sending: https://github.com/zeriontech/zerion-wallet-extension/blob/8ec927fa7ecf45921219bdcad41651bf09bacf8c/src/shared/analytics/mixpanel.ts#L37-L47

PR #631, opened 2024-10-14
Register all owned wallets on initialize

vyorkin:


src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +253,45 @@ export class DnaService {
...
+          }))
+        ) || [];
+
+    for (const { address, origin } of ownedAddresses) {
+      try {
+        await this.registerWallet({
+          params: { address, origin },
+        });
+      } catch (error) {
+        console.log('registerWallet error', error); // eslint-disable-line no-console
```
zerts, line 285: As I see, we will try register each wallet only once. So if something goes wrong, we will not try again.
May be we should not mark register process as complete in this case and try again next time?

vyorkin: I'll update it not set the flag if at least one wallet fails to register

src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +253,49 @@ export class DnaService {
...
+
+    let didFail = false;
+    for (const { address, origin } of ownedAddresses) {
+      try {
+        await this.registerWallet({
+          params: { address, origin },
+        });
+      } catch (error) {
+        didFail = true;
+        console.log('registerWallet error', error); // eslint-disable-line no-console
```
zerts, line 287: Let's do console.warn

vyorkin: updated to `console.warn` ✅

src/modules/dna-service/dna.background.ts
```diff
@@ -9,8 +9,11 @@ import * as browserStorage from 'src/background/webapis/storage';
 import { normalizeAddress } from 'src/shared/normalizeAddress';
 import { emitter } from 'src/background/events';
 import { isReadonlyContainer } from 'src/shared/types/validators';
+import type { Wallet } from 'src/shared/types/Wallet';
+import { INTERNAL_SYMBOL_CONTEXT } from 'src/background/Wallet/Wallet';
 import type { DnaAction } from './types';
 
+const REGISTER_ALL_WALLETS_INVOKED_KEY = 'registerAllWalletsInvoked-14-10-2024';
```
everdimension, line 16: very minor but i think if the key has hyphens it's more readable to have the whole value like that

src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +253,49 @@ export class DnaService {
...
+        });
+      } catch (error) {
+        didFail = true;
+        console.warn('registerWallet error', error); // eslint-disable-line no-console
+      }
+    }
+
+    if (!didFail) {
+      await browserStorage.set(REGISTER_ALL_WALLETS_INVOKED_KEY, true);
+    }
```
everdimension, line 293: I think you're setting this value even if there are no walletGroups found. What should happen in that case?

src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +253,49 @@ export class DnaService {
...
+    for (const { address, origin } of ownedAddresses) {
+      try {
+        await this.registerWallet({
+          params: { address, origin },
+        });
+      } catch (error) {
+        didFail = true;
+        console.warn('registerWallet error', error); // eslint-disable-line no-console
+      }
+    }
```
everdimension, line 289: Do these calls need to be one after the other? Can we use `Promise.all` instead to make them in parallel?

src/modules/dna-service/dna.background.ts
vyorkin: If there are no walletGroups then we have no "unregistered wallets" to register

src/modules/dna-service/dna.background.ts
vyorkin: good point, I'll update the key

src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +253,49 @@ export class DnaService {
     return browserStorage.set(ACTION_QUEUE_KEY, []);
   }
 
+  async registerAllWallets() {
```
everdimension, line 256: Can you add notes for why we're doing this?

src/modules/dna-service/dna.background.ts
everdimension: I see, it mirrors the method name. In that case perhaps it's okay the way it is

src/modules/dna-service/dna.background.ts
vyorkin: Yep, you're right. If backend fails to register a single wallet it doesn't mean we shouldn't attempt to register others. I'll update it to `Promise.all`

src/modules/dna-service/dna.background.ts
everdimension: Shouldn't we wait for the login event maybe?

src/modules/dna-service/dna.background.ts
vyorkin: Ok, if it reads well for you then I leave it as is

src/modules/dna-service/dna.background.ts
vyorkin: I subscribed to the `authenticated` event, which required me to make `account` a parameter in the `initialize` function. This is because `this.record.walletManager.groups` comes from `this.record`, and `this.record` itself is set up during the login flow: `login -> setUser -> updateCredentials -> syncWithWalletStore`. Right after this setup, the `authenticated` event is triggered

src/modules/dna-service/dna.background.ts
vyorkin: Sure! ✅

src/modules/dna-service/dna.background.ts
vyorkin: I increased number of retries to 3, added `POST` to the list "retriable" methods. Now `registerWallet` requests are sent in parallel. Added clarifying comment as well ✅  Please check it out

src/shared/analytics/analytics.ts
```diff
@@ -9,6 +9,7 @@ type MetabaseEvent =
   | 'signed_message'
   | 'signed_transaction'
   | 'client_error'
+  | 'error_registering_dna_action'
```
everdimension, line 12: I don't think we need a special event for this

src/shared/analytics/analytics.background.ts
```diff
@@ -288,6 +289,12 @@ function trackAppEvents({ account }: { account: Account }) {
   emitter.on('eip6963SupportDetected', ({ origin }) => {
     eip6963Dapps.add(origin);
   });
+
+  dnaServiceEmitter.on('registerError', async (error, action) => {
+    const request_name = 'error_registering_dna_action';
+    const params = createParams({ request_name, error, action });
+    sendToMetabase(request_name, params);
```
everdimension, line 296:
```suggestion
    const request_name = 'client_error';
    const message = getError(error).message;
    const type = 'dna action';
    const params = createParams({ request_name, type, message, action });
    sendToMetabase(request_name, params);
```

src/modules/dna-service/dna.background.ts
```diff
@@ -248,7 +260,56 @@ export class DnaService {
...
+      (result) => result.status === 'rejected'
+    );
+
+    if (!hasRejectedRequests) {
+      await browserStorage.set(REGISTER_ALL_WALLETS_INVOKED_KEY, true);
+    }
+  }
+
+  initialize({ account }: { account: Account }) {
+    account.on('authenticated', this.registerAllWallets.bind(this));
```
everdimension, line 312: Does this mean we will only call registerAllWallets when using logs in? What if the user never logs out?

src/shared/analytics/analytics.ts
vyorkin: ah, ok, let it be just `client_error`

src/modules/dna-service/dna.background.ts
vyorkin: I see, that's a good point

everdimension (approved):

PR #635, opened 2024-10-22
Add claim xp banner + xp retrodrop check

vyorkin:
This PR show "Claim Your XP & Level" banner in case when response from `get-meta` has `membership.retro` field which it is not `null`
image


src/ui/pages/Overview/Overview.tsx
```diff
@@ -338,6 +340,15 @@ function OverviewComponent() {
     suspense: false,
   });
 
+  const { data: walletsMeta } = useQuery({
+    queryKey: ['ZerionAPI.getWalletsMeta', params.address],
+    queryFn: () => getWalletsMetaByChunks([params.address]),
```
zerts, line 345: Do we need to show this banner to watch address?

zerts (approved):

PR #636, opened 2024-10-22
Conditionally refetch positions more frequently

everdimension:
Set refetch interval to 4 seconds within 2 minutes of a last known submited transaction

### DIFF
--- a/src/modules/zerion-api/hooks/useWalletPositions.ts
+++ b/src/modules/zerion-api/hooks/useWalletPositions.ts
@@ -25,7 +25,7 @@ export function useHttpAddressPositions(
     suspense?: boolean;
     enabled?: boolean;
     keepPreviousData?: boolean;
-    refetchInterval?: number;
+    refetchInterval?: number | false;
   } = {}
 ) {
   return useQuery({
--- a/src/ui/transactions/usePositionsRefetchInterval.ts
+++ b/src/ui/transactions/usePositionsRefetchInterval.ts
@@ -17,17 +17,24 @@ localTransactionsStore.on('change', (state) => {
   }
 });
 
-function useNow() {
+function useNow(enabled: boolean) {
   const [now, setNow] = useState(() => Date.now());
   useEffect(() => {
-    const id = setInterval(() => setNow(Date.now()), 1000);
-    return () => clearInterval(id);
-  }, []);
+    if (enabled) {
+      const id = setInterval(() => {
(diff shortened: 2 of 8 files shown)


src/ui/transactions/usePositionsRefetchInterval.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/** A flag for perf optimisation: we don't need to track "useNow()" after a tx has been handled */
+let handledTimestamp: number | null = null;
+
+/**
+ * This hook returns a small refetchInterval value
+ * within two minutes of a last known submitted transaction
+ * {false} is a valid value for react query which means no refetch interval
+ */
+export function usePositionsRefetchInterval(fallback: number | false) {
```
zerts, line 47: Let's make a named param. To make it `usePositionsRefetchInterval({fallbackInterval: false})`

src/ui/transactions/usePositionsRefetchInterval.ts
```diff
@@ -0,0 +1,61 @@
...
+const URGENT_REFETCH_INTERVAl = 4000; // 4 seconds
+
+function useNow(enabled: boolean) {
+  const [now, setNow] = useState(() => Date.now());
+  useEffect(() => {
+    if (enabled) {
+      // doesn't really need to match URGENT_REFETCH_INTERVAl,
+      // but also we don't need to refresh more frequently than that
+      const refreshInterval = URGENT_REFETCH_INTERVAl;
+      const id = setInterval(() => {
```
zerts, line 30: Is it ok that we will update `now` only 4 sec after `enabled` was set to `true`?

zerts (approved):

src/modules/zerion-api/hooks/useWalletPositions.ts
```diff
@@ -25,7 +25,7 @@ export function useHttpAddressPositions(
     suspense?: boolean;
     enabled?: boolean;
     keepPreviousData?: boolean;
-    refetchInterval?: number;
+    refetchInterval?: number | false;
```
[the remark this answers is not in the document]
everdimension: 🤔 I don't think it should be

src/ui/transactions/usePositionsRefetchInterval.ts
everdimension: I think so

PR #640, opened 2024-10-24
Refactor formatTokenValue and formatCurrencyValue

everdimension:
Rely more on Intl.NumberFormat

### DIFF
--- a/src/modules/currency/currencies.ts
+++ b/src/modules/currency/currencies.ts
@@ -4,7 +4,7 @@ export interface CurrencyConfig {
   symbol: string;
   modifyParts?: (parts: Intl.NumberFormatPart[]) => Intl.NumberFormatPart[];
   options:
-    | Intl.NumberFormatOptions
+    | (Intl.NumberFormatOptions & { default?: never; lessThanOnde?: never })
     | {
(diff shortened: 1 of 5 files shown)


src/modules/currency/currencies.ts
```diff
@@ -1,16 +1,33 @@
...
-
 export interface CurrencyConfig {
   name: string;
   code: string;
   symbol: string;
   modifyParts?: (parts: Intl.NumberFormatPart[]) => Intl.NumberFormatPart[];
+  options:
+    | Intl.NumberFormatOptions
+    | {
+        default: null | Intl.NumberFormatOptions;
```
zerts, line 9: Is this a valid type definition to express `type = Intl.NumberFormatOptions & { lessThenOne: Intl.NumberFormatOptions }`?

zerts: Otherwise looks like `default` is always null

zerts (approved):

src/modules/currency/currencies.ts
everdimension: Thanks, fixed by adding "never" to the first type option

vyorkin (approved):

PR #641, opened 2024-10-24
Display gasPrices rounded to zero as <$0.01

everdimension:


src/modules/currency/currencies.ts
```diff
@@ -29,6 +29,17 @@ export function resolveOptions(
...
 }
 
+export function resolveOptions(
+  value: number,
+  config: CurrencyConfig | null,
+  runtimeOptions: Intl.NumberFormatOptions | null
+): Intl.NumberFormatOptions | null {
+  const currencyOptions = resolveCurrencyOptions(value, config);
+  return runtimeOptions
+    ? { ...currencyOptions, ...runtimeOptions }
```
zerts, line 39: Can there be a case where runtime option will override currency option? And can it be bad?

everdimension: That's the point though? :)

zerts (approved):

PR #642, opened 2024-10-25
Pass backend_env to ZPI

zerts:


zerts:
https://github.com/zeriontech/zerion-wallet-extension/pull/646

PR #644, opened 2024-10-25
XP drop onboarding wizard and scoring screens

vyorkin:


src/ui/features/xp-drop/XpDropOnboarding/XpDropScoring.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (counter > 6) {
+      navigate('/xp-drop/claim');
+    }
+  });
+
+  const visibleItems = criteriaItems
+    .slice(counter)
+    .map((item, index) => ({ item, key: index }));
+
+  const transitions = useTransition(visibleItems, {
```
vyorkin, line 83: I understand that react-spring isn’t necessary for the current list animation, but we’ll end up using it when we update it... So I just left it as it is for now

src/ui/features/xp-drop/XpDropOnboarding/XpDropScoring.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }, 1000);
+
+    return () => clearInterval(interval);
+  }, []);
+
+  useEffect(() => {
+    if (counter > 6) {
+      navigate('/xp-drop/claim');
+    }
+  });
```
zerts, line 77: do we need a deps array here?

src/ui/features/xp-drop/XpDropOnboarding/XpDropScoring.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setCounter((value) => value + 1);
+    }, 1000);
+
+    return () => clearInterval(interval);
+  }, []);
+
+  useEffect(() => {
+    if (counter > 6) {
```
zerts, line 74: may be it is better to use item's array length as a reference to avoid bugs in the future + less unnamed consts

src/ui/features/xp-drop/XpDropOnboarding/XpDropScoring.tsx
vyorkin: thanks! ✅

src/ui/features/xp-drop/XpDropOnboarding/XpDropScoring.tsx
vyorkin: good idea :) ✅

PR #646, opened 2024-10-30
Support backend_env in all backend resources

everdimension:
Websocket Api, Http Api and Transaction Api
Use createUrl helper for constructing urls


zerts (approved):

src/shared/createUrl.test.ts
```diff
@@ -0,0 +1,213 @@
+import { createUrl } from './createUrl';
+
+describe('createUrl', () => {
```
vyorkin, line 3: 🔥

PR #629, opened 2024-10-11
Referral program

vyorkin:

PR #648, opened 2024-10-31
XP drop check and claim

vyorkin:


src/background/Wallet/Wallet.ts
```diff
@@ -1049,11 +1049,26 @@ export class Wallet {
     }
   }
 
+  private getOfflineSignerByAddress(address: string) {
```
vyorkin, line 1052: Had to do the same thing I did in #629 cuz it is not merged yet

vyorkin: Actually I'll better try to re-use `SignMessageButton` (hence I don't need these changes in this PR)

src/ui/features/xp-drop/XpDropClaimFlow/XpDropClaim.tsx
```diff
@@ -0,0 +1,436 @@
...
+      // To account for this possible delay, we're adding a short wait here.
+      // Otherwise, we might still see the wallet for which we just claimed the XP drop in
+      // the wallet selector when we click on the "Next Wallet" button.
+      await wait(2000);
+    },
+    onSuccess: async () => {
+      // Refetch the wallet metadata to update the list of wallets eligible for the XP drop.
+      await walletsMetaQuery.refetch();
+      // If this is the last wallet, we want to display the success view.
+      if (eligibleAddresses.length === 1) {
```
vyorkin, line 274: Maybe `<=` would be better

vyorkin: ✅

PR #653, opened 2024-11-05
Feature/show gasback in tx views

everdimension:
TODO:
* [x] Show gasback in "Success State"
* [x] Show hint for gasback in a bottomsheet dialog


src/modules/remote-config/plugins/useFirebaseConfig.ts
everdimension, line 1: Moved this file out because otherwise `useQuery` got imported by the background script which shouldn't happen

src/modules/remote-config/types.ts
```diff
@@ -3,4 +3,9 @@ import type { WalletNameFlag } from 'src/shared/types/WalletNameFlag';
 export interface RemoteConfig {
   extension_wallet_name_flags: Record<string, WalletNameFlag[]>;
   extension_uninstall_link: string;
+  ios_loyalty_config: Partial<{
```
vyorkin, line 6: Wouldn't it make more sense to use `extension_loyalty_config` instead (that I created recently)? I don't know what is the point of having separate configs for loyalty though...

src/background/Wallet/Wallet.ts
```diff
@@ -1047,11 +1038,8 @@ export class Wallet {
     const networksStore = getNetworksStore(Model.getPreferences(this.record));
     const networks = await networksStore.loadNetworksByChainId(chainId);
     const nodeUrl = networks.getRpcUrlInternal(networks.getChainById(chainId));
-    if (FEATURE_PAYMASTER_ENABLED) {
-      return new ZksProvider(nodeUrl);
-    } else {
-      return new ethers.providers.JsonRpcProvider(nodeUrl);
-    }
+    // TODO: should we conditionally use regular provider for non-paymaster txs?
+    return new ZksProvider(nodeUrl); // return new ethers.providers.JsonRpcProvider(nodeUrl);
```
vyorkin, line 1042: We need ensure `ZksProvider` works for all transactions (paymaster/non-paymaster)... I’m not sure about the implementation differences between these two providers, but from the first sight, this doesn’t look entirely safe to me :)

src/modules/remote-config/plugins/useFirebaseConfig.ts
vyorkin: I did the same in one of my PR's

vyorkin (approved):

src/modules/remote-config/types.ts
everdimension: Yeah not sure, I think though this is a temporary location for this value and it'll be moved to the tb api

src/ui/pages/SwapForm/SuccessState/SuccessState.tsx
```diff
@@ -148,6 +152,14 @@ export function SuccessState({
           receivePosition={receivePosition}
         />
       </animated.div>
+      {gasbackValue && FEATURE_LOYALTY_FLOW === 'on' ? (
+        <animated.div style={trail[3]}>
```
zerts, line 156: should we update to `useTrail(4)` then?

src/ui/pages/SendForm/SuccessState/SuccessState.tsx
```diff
@@ -16,8 +16,37 @@ import { Button } from 'src/ui/ui-kit/Button';
...
+        borderRadius: 12,
+        background:
+          'linear-gradient(90deg, rgba(160, 36, 239, 0.20) 0%, rgba(253, 187, 108, 0.20) 100%)',
+      }}
+    >
+      <UIText kind="small/regular">Gasback</UIText>
+      <UIText
+        kind="small/accent"
+        style={{
+          background: 'linear-gradient(90deg, #6C6CF9 0%, #FF7583 100%)',
```
zerts, line 39: May be we need move this value to a global var?

src/ui/components/address-action/TransactionConfirmationView/TransactionConfirmationView.tsx
```diff
@@ -29,6 +33,8 @@ export function TransactionConfirmationView({
   eligibilityQuery,
   localAllowanceQuantityBase,
   onOpenAllowanceForm,
+  gasback: gasbackEstimation,
```
zerts, line 36: Maybe we can name this `gasbackValue` and `onGasbackChange`? Like in inputs.

src/modules/remote-config/plugins/useFirebaseConfig.ts
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from '@tanstack/react-query';
+import type { RemoteConfig } from '../types';
+import { fetchRemoteConfig } from './firebase';
+
+export function useFirebaseConfig<T extends keyof RemoteConfig>(
```
zerts, line 5: it looks like I've already seen this function extracted in one of @vyorkin 's prs. Someone will need to merge carefully :))

src/modules/ethereum/account-abstraction/rewards.ts
```diff
@@ -0,0 +1,33 @@
...
+  suppportsSimulations,
+  supportsSponsoredTransactions,
+}: {
+  paymasterEligible: null | boolean;
+  suppportsSimulations: boolean;
+  supportsSponsoredTransactions: boolean | undefined;
+}) {
+  const { data, ...query } = useFirebaseConfig(['loyalty_config'], {
+    enabled:
+      FEATURE_LOYALTY_FLOW === 'on' &&
```
zerts, line 21: Hah, I see that we can just leave `FEATURE_LOYALTY_FLOW && ...`

zerts (approved):

src/modules/ethereum/account-abstraction/rewards.ts
vyorkin: Yeah, I think so..., because we have this line in `config.ts`:
`process.env.FEATURE_LOYALTY_FLOW === 'on' ? 'on' : null`

.github/workflows/pr.yml
```diff
@@ -71,6 +71,7 @@ jobs:
           FEATURE_SEND_FORM: on
           FEATURE_FOOTER_BUG_BUTTON: off
           MIXPANEL_TOKEN_PUBLIC: ${{ secrets.MIXPANEL_TOKEN_PUBLIC_DEV }}
+          FEATURE_LOYALTY_FLOW: on
```
vyorkin, line 74: Should we also add this flag to `release.yml`?

src/modules/remote-config/types.ts
```diff
@@ -3,4 +3,10 @@ import type { WalletNameFlag } from 'src/shared/types/WalletNameFlag';
 export interface RemoteConfig {
   extension_wallet_name_flags: Record<string, WalletNameFlag[]>;
   extension_uninstall_link: string;
+  loyalty_config: Partial<{
+    referrerXpPercent: number;
+    isEnabled?: boolean;
```
vyorkin, line 8: It looks like it's already removed in Firebase, so we should probably remove it from this type too

src/ui/pages/SendTransaction/TransactionConfiguration/TransactionConfiguration.tsx
```diff
@@ -101,26 +203,71 @@ export function TransactionConfiguration({
...
+            chain={chain}
+            onChange={
+              onConfigurationChange
+                ? (nonce) => onConfigurationChange({ ...configuration, nonce })
+                : null
+            }
+          />
+        </div>
+      ) : null}
+      {FEATURE_LOYALTY_FLOW === 'on' ? (
```
vyorkin, line 235: Same here... looks like we could just check the truthiness of `FEATURE_LOYALTY_FLOW`, no need to `=== 'on'`

src/ui/pages/SendForm/SuccessState/SuccessState.tsx
```diff
@@ -110,6 +135,14 @@ export function SuccessState({
           />
         </animated.div>
       ) : null}
+      {gasbackValue && FEATURE_LOYALTY_FLOW === 'on' ? (
```
vyorkin, line 138:
```suggestion
      {gasbackValue && FEATURE_LOYALTY_FLOW ? (
```

vyorkin (approved):

src/modules/ethereum/account-abstraction/rewards.ts
everdimension: Yes but I think what I tried here is to avoid false-positives by doing two things:
1. converting value other than `'on'` to `null` early on (maybe it should be `false`)?
2. AND have a habit to explicitly check for `'on'` in code.

That way if one part fails the result would still be correct :) 

Is this overkill?

src/modules/remote-config/plugins/useFirebaseConfig.ts
everdimension: Fortunately it seems that we've created different files so they do not conflict :)

src/modules/remote-config/types.ts
everdimension: removed

src/ui/components/address-action/TransactionConfirmationView/TransactionConfirmationView.tsx
everdimension: I named it to be consistent with `onFeeValueReady`. It's a dirty side-effect to bubble up the value which is contained inside a child component, so it's not equivalent to a controlled input component, where value comes from the prop

src/ui/pages/SendForm/SuccessState/SuccessState.tsx
everdimension: it's not like we have a design system that consistently uses these gradients...

src/ui/pages/SendForm/SuccessState/SuccessState.tsx
everdimension: https://github.com/zeriontech/zerion-wallet-extension/pull/653/files#r1834131020

src/ui/pages/SendTransaction/TransactionConfiguration/TransactionConfiguration.tsx
everdimension: https://github.com/zeriontech/zerion-wallet-extension/pull/653/files#r1834131020

src/ui/pages/SwapForm/SuccessState/SuccessState.tsx
everdimension: 🔥

PR #654, opened 2024-11-07
Show Explore Rewards link for wallets with new rewards available

vyorkin:


src/ui/pages/WalletSelect/WalletList.tsx
```diff
@@ -76,121 +86,154 @@ function WalletListItem({
...
-          />
-          {isSelected ? <CheckIcon style={{ width: 24, height: 24 }} /> : null}
-        </HStack>
+              }
+            />
+            {isSelected ? (
+              <CheckIcon style={{ width: 24, height: 24 }} />
+            ) : null}
+          </HStack>
+          {exploreRewardsUrl ? (
```
zerts, line 211: It looks like it is better not to extend this component's signature even more, but trying to pass this layout with some prop like `renderItemFooter`. But this is debatable

vyorkin: agree, your suggestion looks more general

src/ui/pages/WalletSelect/WalletSelect.tsx
```diff
@@ -77,18 +78,40 @@ function PortfolioRow({ walletGroups }: { walletGroups: WalletGroup[] }) {
...
+  );
+
+  const { data: walletsMeta, isLoading: isLoadingWalletsMeta } = useQuery({
+    queryKey: ['ZerionAPI.getWalletsMeta', ownedAddresses],
+    queryFn: () =>
+      ownedAddresses ? getWalletsMetaByChunks(ownedAddresses) : null,
+    enabled: Boolean(ownedAddresses),
+  });
 
   const ownedAddressesCount = useMemo(
```
zerts, line 108: Isn't it just a `ownedAddresses.length` now

vyorkin: yes :)

src/ui/pages/WalletSelect/WalletSelect.tsx
```diff
@@ -165,11 +192,13 @@ export function WalletSelect() {
         >
           <WalletList
             walletGroups={walletGroups}
+            walletsMeta={walletsMeta || []}
```
zerts, line 195: Same question here about extending param list for WalletList. May be it is better to make it more abstract and make all wallet-meta related calculations here?

src/ui/pages/WalletSelect/WalletList.tsx
```diff
@@ -76,121 +86,154 @@ function WalletListItem({
...
+              style={{
+                borderRadius: '0 0 18px 18px',
+              }}
+            >
+              <HStack gap={8} alignItems="center" justifyContent="center">
+                <RewardsIcon
+                  style={{
+                    width: 20,
+                    height: 20,
+                    color: 'linear-gradient(90deg, #A024EF 0%, #FDBB6C 100%)',
```
zerts, line 227: I believe this will be worth adding as a variable soon. Something like `--loyalty-gradient`

src/ui/pages/WalletSelect/WalletList.tsx
```diff
@@ -225,6 +272,14 @@ export function WalletList({
   const supportsCssAnchor = CSS.supports('anchor-name: --name');
   for (const group of walletGroups) {
     for (const wallet of group.walletContainer.wallets) {
+      const walletMeta = walletsMeta?.find(
+        (meta) =>
+          normalizeAddress(meta.address) === normalizeAddress(wallet.address)
+      );
+      const exploreRewardsUrl =
+        showExploreRewards && walletMeta?.membership.newRewards
+          ? `${ZERION_ORIGIN}/rewards?section=rewards&address=${wallet.address}`
```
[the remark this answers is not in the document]
vyorkin: The problem is that I can't call this hook inside the loop... I extracted function `getWalletParams` so could call it inside the loop

src/ui/pages/WalletSelect/WalletList.tsx
vyorkin: these gradients are different everywhere... yeah, let's wait and see how many variations we'll end up with eventually :)

src/ui/pages/WalletSelect/WalletSelect.tsx
vyorkin: totally agree! refactored ✅

PR #669, opened 2024-11-18
Add missing check for loyalty enabled flag to onboarding flow

vyorkin:


vyorkin:
It seems we're never going to turn it off anyway. I'll reopen it in that case, though it's highly unlikely

PR #679, opened 2024-12-10
Change Password flow

everdimension:


zerts:
New change password flow allows to create a password without warnings/checks for password strength. May be it is worth adding there too

src/background/account/Account.ts
```diff
@@ -152,39 +197,60 @@ export class Account extends EventEmitter<AccountEvents> {
...
+    const updatedUser = await Account.updateUser(this.user);
+    const currentCredentials = await deriveUserKeys({
+      user: currentUser,
+      credentials: { password: currentPassword },
+    });
+    const newCredentials = await deriveUserKeys({
+      user: updatedUser,
+      credentials: { password: newPassword },
+    });
+    console.log({ currentCredentials, newCredentials });
```
zerts, line 221: Do not forget to remove :)

zerts (approved):

src/ui/pages/Security/ChangePassword.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { UIText } from 'src/ui/ui-kit/UIText';
+import { VStack } from 'src/ui/ui-kit/VStack';
+import { ChangePasswordSuccess } from './ChangePasswordSuccess';
+
+function CustomValidityInput({
+  customValidity,
+  ...props
+}: React.InputHTMLAttributes<HTMLInputElement> & { customValidity: string }) {
+  const ref = useRef<HTMLInputElement>(null);
+  useLayoutEffect(() => {
```
zerts, line 25: Can we just use `useCustomValidity` here instead of new component?

src/ui/pages/Security/ChangePassword.tsx
```diff
@@ -0,0 +1,173 @@
...
+      newPassword: string;
+    }) => {
+      invariant(user, 'User must be defined');
+      await accountPublicRPCPort.request('changePassword', {
+        user,
+        currentPassword,
+        newPassword,
+      });
+    },
+    onSuccess: async () => {
```
zerts, line 53: do we need async here?

src/ui/pages/Security/ChangePassword.tsx
```diff
@@ -0,0 +1,173 @@
...
+                <Input
+                  type="password"
+                  name="newPassword"
+                  placeholder="Set new or verify existing one"
+                  required={true}
+                />
+              </VStack>
+            </VStack>
+          </label>
+          <label style={{ all: 'unset' }}>
```
zerts, line 118: Can we refactor all `Unstyled***` components with this?
Not for this task, just interesting :)

src/background/account/Account.ts
```diff
@@ -156,6 +200,7 @@ export class Account extends EventEmitter<AccountEvents> {
     this.notificationWindow = notificationWindow;
     this.wallet = new Wallet(TEMPORARY_ID, null, this.notificationWindow);
     this.on('authenticated', () => {
+      // TODO: Call Account.writeCurrentUser() here, too?
```
zerts, line 203: Should we do anything with this TODO?

src/background/Wallet/WalletRecord.ts
```diff
@@ -492,7 +492,27 @@ export class WalletRecordModel {
...
+    record: WalletRecord,
+    {
+      credentials,
+      newCredentials,
+    }: { credentials: SessionCredentials; newCredentials: SessionCredentials }
+  ) {
+    // Async update flow for Immer: https://immerjs.github.io/immer/async/
+    const draft = createDraft(record);
+    for (const group of draft.walletManager.groups) {
+      if (isMnemonicContainer(group.walletContainer)) {
```
zerts, line 508: Shouldn't we use `isSignerContainer` here?

zerts (approved):

src/background/Wallet/model/WalletContainer.ts
```diff
@@ -227,6 +227,25 @@ export class MnemonicWalletContainer extends WalletContainerImpl {
...
+
+  async reEncryptWallets({
+    credentials,
+    newCredentials,
+  }: {
+    credentials: SessionCredentials;
+    newCredentials: SessionCredentials;
+  }) {
+    const { mnemonic: encryptedMnemonic } = this.getFirstWallet();
+    invariant(encryptedMnemonic, 'Must be a Mnemonic WalletContainer');
```
vyorkin, line 239: the message just don't make sense to me

src/ui/pages/Security/ChangePassword.tsx
everdimension: We can, but the point is to test the abstraction at a component level :)

src/ui/pages/Security/ChangePassword.tsx
everdimension: I guess not!

zerts (approved):

src/modules/persistent-store-base.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  getState() {
+    if (!this.isReady) {
+      throw new Error('Do not access getState() before checking ready()');
+    }
+    return super.getState();
+  }
+
+  setState(...args: Parameters<Store<T>['setState']>) {
```
zerts, line 45: Should we specify setState as async here?

src/modules/persistent-store-base.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (process.env.NODE_ENV === 'development') {
+        // Throw only in dev mode in case some production flow depends on
+        // setting state sooner. Before this refactoring it was possible, so
+        // it shouldn't really break anything
+        throw new Error(
+          'You are trying to write to a PersistentStore before {ready}'
+        );
+      }
+    }
+    super.setState(...args);
```
zerts, line 56: As I can see here, we have both sync and async state updates. Can this cause the racecondition when sync updates will be called in one order, but async saves will be executed in another order which will cause incorrect state after store reload?

src/background/account/Account.ts
```diff
@@ -422,6 +437,7 @@ export class AccountPublicRPC {
     user: PublicUser;
     password: string;
   }>): Promise<PublicUser | null> {
+    invariant(!this.account.isWriteLocked, 'Atomic operation in progress');
```
zerts, line 440: shouldn't logout also be blocked while updating is in progress?

src/background/Wallet/persistence.ts
```diff
@@ -42,25 +69,130 @@ export class WalletStore extends PersistentStore<WalletStoreState> {
...
+    id,
+    credentials,
+    newCredentials,
+  }: {
+    id: string;
+    credentials: SessionCredentials;
+    newCredentials: SessionCredentials;
+  }) {
+    await this.ready();
+    console.log('reading', { id, credentials, state: this.getState() });
```
zerts, line 188: remove?

zerts: Looks dangerous to expose credentials :)

src/background/Wallet/persistence.ts
```diff
@@ -42,25 +69,130 @@ export class WalletStore extends PersistentStore<WalletStoreState> {
   }
 
   /** Prefer WalletStore['save'] unless necessary */
-  async encryptAndSave(
+  private async encryptAndSave(
     id: string,
-    encryptionKey: string,
+    credentials: Credentials,
```
zerts, line 74: do we need to pass all Credentials here?

src/background/Wallet/persistence.ts
```diff
@@ -42,25 +69,130 @@ export class WalletStore extends PersistentStore<WalletStoreState> {
...
-  async save(id: string, encryptionKey: string, record: WalletRecord) {
+  async save(id: string, credentials: Credentials, record: WalletRecord) {
     if (this.lastRecord === record) {
       return;
     }
-    await this.encryptAndSave(id, encryptionKey, record);
+    await this.encryptAndSave(id, credentials, record);
+  }
+
+  async createBackup(id: string) {
```
zerts, line 96: what do you think about calling this param `userId` to make to purpose more clear?

src/background/Wallet/persistence.ts
```diff
@@ -42,25 +69,130 @@ export class WalletStore extends PersistentStore<WalletStoreState> {
...
+          name: 'internal_error',
+          message: 'Atomic wallet update failed. Restored from backup.',
+        });
+        console.log('Successfully restored wallet record from backup'); // eslint-disable-line no-console
+      } catch {
+        emitter.emit('globalError', {
+          name: 'internal_error',
+          message: 'Atomic wallet update failed. Restore from backup failed.',
+        });
+        throw new InternalBackupError(getError(error), { didRestore: false });
```
zerts, line 172: Do we need this param for this Error? It looks like we only use it with didRestore: false

src/modules/persistent-store-base.ts
everdimension: It wouldn't change the type though

src/background/Wallet/persistence.ts
everdimension: No, but it makes it more consistent with the other signatures. I think it's more proper to pass credentials, it's easier to abstract them this way

src/modules/persistent-store-base.ts
zerts: Yes! But will make it a bit more obvious :)

zerts:
https://github.com/zeriontech/zerion-wallet-extension/pull/878

PR #680, opened 2024-12-11
Glowup/hide balances mode

everdimension:


zerts:
Balances are not hidden in the send form (for address input and for the token balance)

zerts:
For manage wallets it is the same. Should I check all corners of the UI right now?

zerts:
The branch is too old. Better plan to implement it again :(

PR #681, opened 2024-12-15
Tests for typedData signing and paymaster serialization

everdimension:


src/modules/ethereum/message-signing/signTypedData.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        { name: 'version', type: 'string' },
+        { name: 'chainId', type: 'uint256' },
+      ],
+    },
+    domain: { name: 'zkSync', version: '2', chainId: '543210' },
+    primaryType: 'Transaction',
+    message: {
+      txType: 113,
+      from: '0xE093d671dA3D42fBf79BC333692a7A5f794EdDFb',
+      to: '0x6a6394f47dd0baf794808f2749c09bd4ee874e70',
```
zerts, line 36: Not sure is it important or not, to have everywhere `from` in mixed case and `to` in lowercase?

zerts (approved):

src/modules/ethereum/message-signing/signTypedData.test.ts
everdimension: Hmm I think I copied objects as is. Ideally this shouldn't matter

PR #684, opened 2024-12-17
Use dialog "close" event for onClosed cb

everdimension:


src/ui/ui-kit/ModalDialogs/BaseDialog.tsx
```diff
@@ -76,6 +77,15 @@ export const BaseDialog = React.forwardRef(
       };
     }, []);
 
+    useEffect(() => {
+      const dialogEl = dialogRef.current;
+      if (dialogEl) {
+        const onClosed = () => onClosedRef.current?.();
```
zerts, line 83: Name collision with the component's param. Not critical

zerts (approved):

PR #694, opened 2025-01-16
Track page views in Google Analytics

vyorkin:


src/shared/analytics/analytics.background.ts
```diff
@@ -348,7 +356,13 @@ export function initialize({ account }: { account: Account }) {
     willSendRequest: createAddProviderHook({ getWalletProvider }),
   });
   const handleUserId = () => mixpanelIdentify(account);
-  account.on('authenticated', () => handleUserId());
+
+  account.on('authenticated', () =>
```
zerts, line 360: I think, we can create a separate .on call for GA.
No need to combine them and probably no need to use Promise.allSettled()

.env.example
```diff
@@ -11,3 +11,5 @@ FEATURE_FOOTER_BUG_BUTTON=on
 FEATURE_SEND_FORM=
 MIXPANEL_TOKEN_PUBLIC=
 FEATURE_LOYALTY_FLOW=off
+GOOGLE_ANALYTICS_MEASUREMENT_ID=
+GOOGLE_ANALYTICS_API_SECRET=
```
zerts, line 15: This api key will be exposed. Probably it is ok, but everything that has secret in its name raises a concern

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+import { onIdle } from './analytics';
+
+type GoogleAnalyticsEvent = 'page_view';
+
+export const SESSION_ID_KEY = 'GOOGLE_ANALYTICS_V4_SESSION_ID';
+const CLIENT_ID_KEY = 'GOOGLE_ANALYTICS_V4_CLIENT_ID';
+const ENDPOINT = 'https://www.google-analytics.com/mp/collect';
+const DEFAULT_ENGAGEMENT_TIME_IN_MSEC = 100;
+
+function readClientid() {
```
zerts, line 17: this should be an async function.

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+  return browserStorage.get<string>(CLIENT_ID_KEY);
+}
+
+async function getClientId() {
+  let clientId = await readClientid();
+  if (!clientId) {
+    // Generate a unique client ID for Google Analytics V4.
+    // The actual value is not relevant.
+    clientId = crypto.randomUUID();
```
zerts, line 26: can't we reuse userId that was generated for mixpanel here?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+    await browserStorage.set(CLIENT_ID_KEY, clientId);
+  }
+  return clientId;
+}
+
+export async function resetGoogleAnalyticsSessionId() {
+  const sessionId = crypto.randomUUID() as string;
+  await chrome.storage.session.set({
```
zerts, line 34: Isn't this also `await browserStorage.set(CLIENT_ID_KEY, clientId);`?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+    [SESSION_ID_KEY]: sessionId,
+  });
+  return sessionId;
+}
+
+async function getSessionId() {
+  // A sessionId represents a period of continuous user interaction with the extension.
+  // Tracking it is necessary for Google Analytics V4.
+  const result = await browser.storage.session.get(SESSION_ID_KEY);
```
zerts, line 43: Isn't this `readClientid()`?

zerts: Ok, I see the difference, shouldn't we use `browserStorage.get<string>(SESSION_ID_KEY)` here?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+  let sessionId = result?.[SESSION_ID_KEY] as string | undefined;
```
zerts, line 44: Is it correct that we get result by SESSION_ID_KEY and then reading [SESSION_ID_KEY] in result?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+    sessionId = await resetGoogleAnalyticsSessionId();
+  }
+  return sessionId;
+}
+
+function getEventParams(event: GoogleAnalyticsEvent) {
+  if (event === 'page_view')
+    return {
+      page_title: document.title,
+      page_location: document.location.href,
```
zerts, line 55: do we need this .href or .hash?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  logToConsole(Loglevel.info, 'group', `Google Analytics: ${event}`);
+  logTable(Loglevel.info, eventPayload);
+  logToConsole(Loglevel.info, 'groupEnd');
+
+  if (process.env.NODE_ENV !== 'development') {
+    onIdle(() => {
+      fetch(
+        `${ENDPOINT}?measurement_id=${GOOGLE_ANALYTICS_MEASUREMENT_ID}&api_secret=${GOOGLE_ANALYTICS_API_SECRET}`,
```
zerts, line 95: May be it will be a bit more readable to use URL() constructor here?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+      page_title: document.title,
+      page_location: document.location.href,
+    };
+
+  return {};
+}
+
+export async function sendToGoogleAnalytics(
+  event: GoogleAnalyticsEvent,
+  extraParams: Record<string, unknown>
```
zerts, line 63: Do we need this extra params?

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+    },
+  };
+
+  logToConsole(Loglevel.info, 'group', `Google Analytics: ${event}`);
+  logTable(Loglevel.info, eventPayload);
+  logToConsole(Loglevel.info, 'groupEnd');
+
+  if (process.env.NODE_ENV !== 'development') {
+    onIdle(() => {
+      fetch(
```
zerts, line 94: What about using `ky` here?

src/shared/analytics/analytics.background.ts
vyorkin: Agree, separate handler is better ✅

src/shared/analytics/google-analytics.ts
vyorkin: updated ✅

src/shared/analytics/google-analytics.ts
vyorkin: `local` and `session` storage are different (according to [docs](https://developer.chrome.com/docs/extensions/reference/api/storage#storage-areas)). Probably we could use browserStorage (`local`) here as well, but the google analytics v4 docs suggests using session storage. I think that is because the session storage is cleared if the extension is disabled, reloaded or updated and when the browser restarts

src/shared/analytics/google-analytics.ts
vyorkin: yes

src/shared/analytics/google-analytics.ts
vyorkin: https://github.com/zeriontech/zerion-wallet-extension/pull/694#discussion_r1919635169

src/shared/analytics/google-analytics.ts
vyorkin: google analytics docs suggests using `.href`

src/shared/analytics/google-analytics.ts
vyorkin: Not yet, but we may need them if we decide to send additional events to GA

.env.example
vyorkin: 👍  Agree, it shouldn't cause much harm, but I'd prefer not to expose it

src/shared/analytics/analytics.background.ts
```diff
@@ -348,7 +361,13 @@ export function initialize({ account }: { account: Account }) {
...
+    // Google Analytics does not have a clear definition of a user session.
+    // Hence, we need to define what a user session means within the extension.
+    // For simplicity, we initiate a new session each time the user authenticates.
+    Promise.allSettled([handleUserId(), resetGoogleAnalyticsSessionId()])
```
everdimension, line 369: I don't get it, what does `.allSettled` achieve here? It's not being awaited nor caught

src/shared/analytics/google-analytics.ts
everdimension: we have a device id, why not reuse it?

.env.example
vyorkin: Network traffic inspection could indeed reveal these GA API tokens... Could be solved by creating our custom endpoint, just like we did for access to firebase config. I'm just not sure it's worth it at this point

src/shared/analytics/analytics.background.ts
vyorkin: Thats odd, doesn't make sense to me either... Let me update this part

src/shared/analytics/google-analytics.ts
vyorkin: wow, I missed it, thanks, this is exactly what I needed

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,181 @@
...
+  origin: globalThis.location.origin,
+  app_version: version,
+});
+
+export function gaBeginSession(account: Account) {
+  const userId = account.getUser()?.id;
+  if (!userId) {
+    return;
+  }
+  googleAnalyticsApi.setSessionId(userId);
```
zerts, line 164: Do we need to store this sessionId if it is a userId and we have it on every event anyway?

src/shared/analytics/analytics.background.ts
```diff
@@ -101,6 +102,13 @@ function trackAppEvents({ account }: { account: Account }) {
     mixpanelTrack(account, 'General: Screen Viewed', mixpanelParams);
   });
 
+  emitter.on('screenView', async (data) => {
+    gaCollect(account, 'page_view', {
+      page_title: data.title,
```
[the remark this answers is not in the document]
vyorkin: Yes, https://developer.chrome.com/docs/extensions/how-to/integrate/google-analytics-4
> The page_view event also requires the page_title and page_location parameters

src/shared/analytics/google-analytics.ts
vyorkin: I agree, while in mixpanel client it makes sense to store the user_id as a field within the API client class, GA api client is a bit simpler: we only have a single `collect` method that requires `session_id`, so we could just pass `user_id` as a parameter each time it’s called

vyorkin: updated ✅

src/shared/analytics/google-analytics.ts
vyorkin: updated ✅

src/shared/analytics/google-analytics.ts
vyorkin: not relevant anymore

src/shared/analytics/google-analytics.ts
vyorkin: not relevant anymore

src/shared/analytics/analytics.background.ts
```diff
@@ -177,6 +185,7 @@ function trackAppEvents({ account }: { account: Account }) {
         ...addressActionAnalytics,
       });
       sendToMetabase('signed_transaction', params);
+      gaCollect(account, 'signed_transaction', params);
```
zerts, line 188: may be we can introduce a helper `prepareGaParams(account, params)` that will get a userId from account and add it to params. To make google-analytics service independent from Account class implementation. Not a big deal though.

src/shared/analytics/analytics.background.ts
```diff
@@ -348,12 +358,11 @@ export function initialize({ account }: { account: Account }) {
     willSendRequest: createAddProviderHook({ getWalletProvider }),
   });
   const handleUserId = () => mixpanelIdentify(account);
+
```
zerts, line 361: I think, we can revert changes here and in the account.on('reset') listener :)
Not a big deal, though

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,165 @@
...
+
+const googleAnalyticsApi = new GoogleAnalyticsApi({
+  apiSecret: GOOGLE_ANALYTICS_API_SECRET ?? null,
+  measurementId: GOOGLE_ANALYTICS_MEASUREMENT_ID ?? null,
+  resolveDeviceId: () => deviceIdStore.getSavedState(),
+  debugMode: process.env.NODE_ENV !== 'production',
+  sendRequestsOverTheNetwork: process.env.NODE_ENV === 'production',
+});
+
+Object.assign(globalThis, { googleAnalyticsApi });
```
zerts, line 141: Do we need this?

zerts (approved):

src/shared/analytics/google-analytics.ts
vyorkin: Not really, I used it for debugging only, removed ✅

src/shared/analytics/analytics.background.ts
vyorkin: reverted ✅

src/env/config.ts
```diff
@@ -34,3 +34,12 @@ export const MIXPANEL_TOKEN_PUBLIC = process.env.MIXPANEL_TOKEN_PUBLIC;
...
 export const SLOW_MODE = false;
+export const GOOGLE_ANALYTICS_MEASUREMENT_ID = process.env.GOOGLE_ANALYTICS_MEASUREMENT_ID;
+export const GOOGLE_ANALYTICS_API_SECRET = process.env.GOOGLE_ANALYTICS_API_SECRET;
+
+if (!PROXY_URL) {
+  throw new Error('PROXY_URL must be defined in ENV');
+}
+if (!ZERION_API_URL) {
+  throw new Error('ZERION_API_URL must be defined in ENV');
+}
```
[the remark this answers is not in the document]
vyorkin: updated ✅ merging issue

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,185 @@
...
+
+type GoogleAnalyticsEvent =
+  | 'first_open'
+  | 'page_view'
+  | 'signed_message'
+  | 'signed_transaction';
+
+class GoogleAnalyticsApi {
+  baseParams?: Record<string, unknown>;
+  deviceId?: string;
```
everdimension, line 32: I think better to be explicit
```suggestion
  baseParams: null | Record<string, unknown>;
  deviceId: null | string;
```

everdimension (approved):

src/shared/analytics/google-analytics.ts
```diff
@@ -0,0 +1,185 @@
...
+if (!GOOGLE_ANALYTICS_API_SECRET) {
+  // eslint-disable-next-line no-console
+  console.warn('GOOGLE_ANALYTICS_API_SECRET env var not found');
+}
+if (!GOOGLE_ANALYTICS_MEASUREMENT_ID) {
+  // eslint-disable-next-line no-console
+  console.warn('GOOGLE_ANALYTICS_MEASUREMENT_ID env var not found');
+}
+
+const USE_PAYLOAD_VALIDATION_ENDPOINT = false;
```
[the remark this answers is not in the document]
vyorkin: The flag to use GA's "validation endpoint", helped me to figure out the reason why we didn't receive events (I was sending array instead of object)

PR #695, opened 2025-01-20
Refactor paymaster and regular tx interpretation into a single call

everdimension:
- [x] Pass `domain` param to `interpret signature`


src/ui/shared/requests/uiInterpretTransaction.ts
```diff
@@ -0,0 +1,135 @@
...
+  data?: null | { data: { eligible: boolean } };
+  status: 'error' | 'success' | 'loading';
+  isError: boolean;
+}
+
+/**
+ * Interprets Paymaster Transaction if it is eligible
+ * or a regular transaction otherwise
+ */
+export async function interpretTxBasedOnEligibility({
```
everdimension, line 31: Feel free to suggest a better name here :)

src/modules/ethereum/transactions/interpret.ts
```diff
@@ -101,6 +103,7 @@ export function interpretSignature({
             chain_id: chainId,
             currency,
             typed_data: typedData,
+            domain: origin,
```
zerts, line 106: is this a new param?

src/ui/shared/requests/uiInterpretTransaction.ts
zerts: I think, we can go with a simple `interpretTransaction`. Cause ideally, this should just interpret any transaction, everything else is just a details of implementation

src/ui/shared/requests/uiInterpretTransaction.ts
```diff
@@ -0,0 +1,131 @@
...
+  const preferences = await getPreferences();
+  const source = preferences?.testnetMode?.on ? 'testnet' : 'mainnet';
+  const networksStore = await getNetworksStore();
+  const networks = await networksStore.loadNetworksByChainId(
+    normalizeChainId(transaction.chainId)
+  );
+  const chain = networks.getChainById(normalizeChainId(transaction.chainId));
+  const network = chain ? networks?.getNetworkByName(chain) || null : null;
+  invariant(network, `Unidentified network: ${transaction.chainId}`);
+  if (!network.supports_simulations) {
```
zerts, line 50: should we return null while `eligibilityQueryStatus === 'loading'`?

everdimension: Hm, let me check

everdimension: fixed

zerts (approved):

PR #698, opened 2025-01-22
Fix pre-eip-1559 type mismatch error

everdimension:


src/background/messaging/mapRPCMessageToController.ts
```diff
@@ -57,14 +57,18 @@ export function mapRPCMessageToController<T>(
...
-          return formatJsonRpcWalletError(
+          const normalizedError = formatJsonRpcWalletError(
             id,
             error instanceof DOMException
               ? domExceptionToError(error).message
               : 'code' in error
               ? error
               : getError(error)
           );
+          console.group('Controller error'); // eslint-disable-line no-console
```
zerts, line 68: Do we need this in production?

everdimension: Yep, so that it's easier for users to report what happened :)

zerts (approved):

PR #700, opened 2025-01-23
Fix Overview and WalletSelect when backend is down

everdimension:


src/ui/pages/History/ActionsList/ActionsList.tsx
```diff
@@ -55,7 +55,7 @@ export function ActionsList({
                 const hash = addressAction.transaction.hash;
                 return {
                   key: isLocalAddressAction(addressAction)
-                    ? addressAction.relatedTransaction || hash
+                    ? `local-${addressAction.relatedTransaction || hash}`
```
zerts, line 58: why do we need this?

src/ui/pages/WalletSelect/WalletSelect.tsx
```diff
@@ -108,7 +108,10 @@ export function WalletSelect() {
   );
 
   const { data: walletsMeta, isLoading: isLoadingWalletsMeta } =
-    useWalletsMetaByChunks({ addresses: ownedAddresses });
+    useWalletsMetaByChunks({
```
zerts, line 111: wallet meta request that blocks render make wallet list opens very slow if fails, but probably this is out of scope for this fix

zerts (approved):

src/ui/pages/History/ActionsList/ActionsList.tsx
everdimension: It's not related but I decided not to create a separate PR for this
It fixed react error that multiple components for same key are rendered. Turns out it happens when you try to accelerate a tx and initial tx gets mined

src/ui/pages/WalletSelect/WalletSelect.tsx
everdimension: Yep, I think we can add a smaller timeout here, but I also think it can be done separately

PR #707, opened 2025-01-30
Use last swap chain as default in SwapForm

everdimension:


src/background/Wallet/model/types.ts
```diff
@@ -102,9 +103,15 @@ export interface WalletRecordVersion4
...
+export interface WalletRecordVersion5
+  extends Omit<WalletRecordVersion4, 'version'> {
+  /** This version introduces normalized addresses in "permissions", see ./versions.ts */
   version: 5;
-  /** This version introduces normalized addresses in "permissions" */
+}
+
+export interface WalletRecord extends Omit<WalletRecordVersion5, 'version'> {
+  version: 6;
+  activityRecord: ActivityRecord;
```
everdimension, line 114: I wasn't sure if this needed a version upgrade. We could make this as optional and keep version 5. But it seems more correct to upgrade

src/background/Wallet/WalletRecord.ts
```diff
@@ -836,6 +853,29 @@ export class WalletRecordModel {
...
+    record: WalletRecord,
+    { address, chain }: { address: string; chain: string }
+  ) {
+    return produce(record, (draft) => {
+      draft.activityRecord[address] ??= {};
+      draft.activityRecord[address].lastSwapChain = chain;
+    });
+  }
+
+  static tempResetActivity(record: WalletRecord) {
```
zerts, line 873: Do we need this method?

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -956,10 +961,51 @@ export function SwapFormComponent() {
...
+    queryKey: ['wallet/getLastSwapChainByAddress', address],
+    queryFn: () => walletPort.request('getLastSwapChainByAddress', { address }),
+  });
+  useEffect(() => {
+    if (prepared || !isFetchedAfterMount) {
+      return;
+    }
+    if (lastUsedChain) {
+      searchParams.set('chainInput', lastUsedChain);
+      setSearchParams(searchParams, { replace: true });
```
zerts, line 986: With this approach we will keep this chain for other addresses after changing them.
In this case user did nothing to set up the chain manually, so may be it will be better to select another default chain for other address?

zerts (approved):

src/background/Wallet/WalletRecord.ts
everdimension: I'll remove :)

src/ui/pages/SwapForm/SwapForm.tsx
everdimension: discussed)

PR #710, opened 2025-02-04
Create pending transaction state

zerts:


src/ui/shared/forms/SuccessState/useActionStatusByHash.ts
```diff
@@ -0,0 +1,39 @@
...
+    refetchOnWindowFocus: false,
+    refetchInterval: (data) => (data ? false : 3000),
+  });
+
+  const ethersv5Receipt = data ? toEthersV5Receipt(data) : null;
+  const nodeStatus = ethersv5Receipt
+    ? transactionReceiptToActionStatus({ receipt: ethersv5Receipt })
+    : null;
+
+  return localStatus || nodeStatus || 'pending';
```
zerts, line 38: Is it ok to use pending as a fallback option?

src/modules/ethereum/transactions/getTransactionReceipt.ts
```diff
@@ -0,0 +1,28 @@
+import type { TransactionReceipt } from 'ethers';
+import { SLOW_MODE } from 'src/env/config';
+import type { Chain } from 'src/modules/networks/Chain';
+import type { Networks } from 'src/modules/networks/Networks';
+import { sendRpcRequest } from 'src/shared/custom-rpc/rpc-request';
+import { wait } from 'src/shared/wait';
+
+export async function getTransactionReceipt({
```
everdimension, line 8: `TransactionPoller` is already performing this work
I think we should be able to tap into it from the UI

zerts: I was thinking about this, but didn't get the easy solution. Will think again

src/ui/shared/forms/SuccessState/SuccessStateLoader.tsx
```diff
@@ -0,0 +1,317 @@
...
+                </HStack>
+                <UIText
+                  kind="headline/hero"
+                  className={cn(
+                    styles.title,
+                    status !== 'confirmed' && styles.hidden
+                  )}
+                >
+                  {confirmedTitle}
+                </UIText>
```
everdimension, line 255: I know it's minor but I don't like that text selection is messed up in this area because of overlapping text elements with zero opacity
Can we fix this?

src/ui/shared/forms/SuccessState/SuccessStateLoader.tsx
```diff
@@ -0,0 +1,317 @@
...
+      : null;
+
+  return (
+    <PageColumn
+      style={{
+        width: 'clamp(320px, 100vw, 450px)',
+        marginInline: 'auto',
+        position: 'relative',
+      }}
+    >
```
everdimension, line 213: What do you think about making this whole element appear with a quick fade-in? I think currently it feels a bit off that it appears kinda unexpectedly but then has fluid animations within

src/ui/shared/forms/SuccessState/SuccessStateLoader.tsx
zerts: Good point!

src/ui/shared/forms/SuccessState/SuccessStateLoader.module.css
```diff
@@ -43,10 +56,12 @@
 
 .title.hidden {
   opacity: 0;
+  user-select: none;
```
everdimension, line 59: Does this work? I'd think it should be `pointer-events: none` instead

everdimension (approved):

src/ui/shared/forms/SuccessState/SuccessStateLoader.module.css
everdimension: I meant that the visible text should still be selectable :)

PR #712, opened 2025-02-06
Fix asset select scroll

zerts:


src/ui/pages/SendForm/AssetSelect/AssetSelect.tsx
```diff
@@ -472,6 +472,7 @@ function AssetSelectComponent({
     if (optionItem.type === ItemType.group) {
       return {
         key: optionItem.name,
+        pad: false,
```
zerts, line 475: Without this option SurfaceList tries to wrap this item which caused extra offset and broken scroll calculations which leads to add and remove item to the tail of the visible list

everdimension (approved):

PR #718, opened 2025-02-18
Update default slippage

vyorkin:


src/ui/pages/SwapForm/SlippageSettings/SlippageLine.tsx
```diff
@@ -8,15 +8,27 @@ import { formatPercent } from 'src/shared/units/formatPercent/formatPercent';
...
-  const { slippage } = useSelectorStore(swapView.store.configuration, [
-    'slippage',
-  ]);
+  const { slippage: userSlippage } = useSelectorStore(
+    swapView.store.configuration,
+    ['slippage']
+  );
+
+  const { defaultSlippagePercent } = getSlippageOptions(chain);
+  const slippage = userSlippage ?? defaultSlippagePercent / 100;
```
zerts, line 31: I think, it can be more consistent to use the same units all over the code.
What do you think about just have `defaultSlippage = 0.01`?

zerts: However, I can see that is was also this inconsistent before, so not a big problem then

src/ui/pages/SwapForm/SlippageSettings/SlippageSettings.tsx
```diff
@@ -139,18 +138,23 @@ function CustomValueOverlay({
...
-  const [percentValue, setPercentValue] = useState(() =>
-    String(toPercents(slippage))
-  );
+  const { slippage: userSlippage } = configuration;
+  const { defaultSlippagePercent, slippageOptions } = getSlippageOptions(chain);
+  const slippage = userSlippage
+    ? toPercents(userSlippage)
+    : defaultSlippagePercent;
+
+  const [percentValue, setPercentValue] = useState(() => String(slippage));
```
zerts, line 155: Can we cast String() in the previous expression? And use here just `useState(slippage)`?

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
```diff
@@ -0,0 +1,20 @@
+import type { Chain } from 'src/modules/networks/Chain';
+
+// [defaultOptionIndex, [option1, option2]]
+type SlippageOptions = [number, [number, number]];
```
zerts, line 4: What do you think about making this an object?
```
type SlippageOptions = {
  default: number,
  options: [number, number]
```?

File: src/ui/pages/SwapForm/SwapForm.tsx [hunk 4]
```diff
@@ -620,28 +620,32 @@ export function SwapFormComponent() {
...
-                    swapView.store.configuration.setState(value);
-                    slippageDialogRef.current?.close();
-                  }}
-                />
-              )}
-            />
-          </>
-        )}
+        renderWhenOpen={() => {
+          invariant(chain, 'Chain must be defined');
```
zerts, line 624: Can't this lead to an error for the user? Shouldn't we just wait until chain is not null?

zerts: Ok, I see that the line that should call for this view shows only when chain is not null. So may be there is no problem

zerts:
We need to change validation logic here:
May be we can to change validation to >1%, not >= 1%?
Zerts Screenshot 2025-02-18 at 12 13@2x

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
```diff
@@ -0,0 +1,20 @@
...
+
+interface SlippageConfig {
+  [chainId: string]: SlippageOptions;
+}
+
+const DEFAULT_SLIPPAGE_OPTIONS: SlippageOptions = [1, [0.5, 1.0]];
+const SLIPPAGE_BY_CHAIN: SlippageConfig = {
+  ethereum: [1, [0.2, 0.5]],
```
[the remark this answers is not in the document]
zerts: Ah, I see. Well let's make this an array then? This first value is a bit confusing :)

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
vyorkin: I initially implemented the `SlippageOptions` type as an object type, exactly as shown in your snippet. However, I later reconsidered this approach as I dislike the repetition of field names, especially considering we might switch to firebase configuration in the future. In my opinion, a more elegant solution would be to use a direct "tuple type" like this: `type SlippageOptions = [number, number, number];` with a descriptive comment above. But perhaps this is just my preference...

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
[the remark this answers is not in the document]
vyorkin: Yeah... maybe, this is why I added comment above the `SlippageOptions` types:
`// [defaultOptionIndex, [option1, option2]]`
The first field is the `defaultOptionIndex`

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: Yes, we have similar invariant below, shouldn't be a problem

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
vyorkin: Just discussed this with @everdimension , going to update this type to: `{ default: number; options: [number, number]; }`

src/ui/pages/SwapForm/SlippageSettings/SlippageSettings.tsx
vyorkin: ✅

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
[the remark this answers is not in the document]
vyorkin: Just updated this part, it is an object now

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
vyorkin: updated ✅

src/ui/pages/SwapForm/SlippageSettings/getSlippageOptions.ts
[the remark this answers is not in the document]
vyorkin: I changed these types as we discussed ✅

src/ui/pages/SwapForm/Quotes/useQuotes.ts
```diff
@@ -92,12 +98,18 @@ export function useQuotes({
         value,
         getDecimals({ asset: position.asset, chain })
       ).toFixed();
+
+      const { default: defaultSlippagePercent } = getSlippageOptions(chain);
+      const slippagePercent = String(
+        userSlippage ? toPercents(userSlippage) : defaultSlippagePercent
```
zerts, line 104: We shouldn't forget this PR top be merged https://github.com/zeriontech/transactions-ui/pull/15/files and package.json to be updated

zerts: Also here we need to use `??` instead of `?`
Slippage: 0 technically should be possible.

src/ui/pages/SwapForm/SlippageSettings/SlippageLine.tsx
```diff
@@ -8,22 +8,34 @@ import { formatPercent } from 'src/shared/units/formatPercent/formatPercent';
...
+  const { default: defaultSlippagePercent } = getSlippageOptions(chain);
+  const slippage = userSlippage ?? defaultSlippagePercent / 100;
 
   const price = swapView.receiveAsset?.price?.value || 0;
   const fiatValue = new BigNumber(receiveInput || 0)
     .times(price)
     .times(1 - slippage);
 
-  return slippage >= HIGH_SLIPPAGE_THRESHOLD ? (
+  return slippage > HIGH_SLIPPAGE_THRESHOLD ? (
```
zerts, line 38: There was a discussion about this logic not so long ago but I think, this is ok for now. I don't want to make HIGH_SLIPPAGE_THRESHOLD dependant on chain too in this PR

src/ui/pages/SwapForm/SlippageSettings/SlippageSettings.tsx
```diff
@@ -139,18 +138,25 @@ function CustomValueOverlay({
...
 }) {
-  const { slippage } = configuration;
-  const [percentValue, setPercentValue] = useState(() =>
-    String(toPercents(slippage))
+  const { slippage: userSlippage } = configuration;
+  const { default: defaultSlippagePercent, options: slippageOptions } =
+    getSlippageOptions(chain);
+
+  const slippage = String(
+    userSlippage ? toPercents(userSlippage) : defaultSlippagePercent
```
zerts, line 154: https://github.com/zeriontech/zerion-wallet-extension/pull/718/files#r1966830269
https://github.com/zeriontech/zerion-wallet-extension/pull/718/files#r1966830670

everdimension (changes requested):
Please address this https://github.com/zeriontech/zerion-wallet-extension/pull/718#discussion_r1966830269

src/ui/pages/SwapForm/Quotes/useQuotes.ts
vyorkin: `userSlippage == null ? toPercents(userSlippage) : defaultSlippagePercent`

src/ui/pages/SwapForm/SlippageSettings/SlippageLine.tsx
```diff
@@ -8,22 +8,34 @@ import { formatPercent } from 'src/shared/units/formatPercent/formatPercent';
...
-  const { slippage } = useSelectorStore(swapView.store.configuration, [
-    'slippage',
-  ]);
+  const { slippage: userSlippage } = useSelectorStore(
+    swapView.store.configuration,
+    ['slippage']
+  );
+
+  const { default: defaultSlippagePercent } = getSlippageOptions(chain);
+  const slippage = userSlippage ?? defaultSlippagePercent / 100;
```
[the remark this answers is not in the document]
vyorkin: Actually good idea! I refactored `getSlippageSettings`, so now it calculates everything that possible callers need

src/ui/pages/SwapForm/SlippageSettings/SlippageLine.tsx
```diff
@@ -8,22 +8,34 @@ import { formatPercent } from 'src/shared/units/formatPercent/formatPercent';
...
   const { receiveInput } = useSelectorStore(swapView.store, ['receiveInput']);
-  const { slippage } = useSelectorStore(swapView.store.configuration, [
-    'slippage',
-  ]);
+  const { slippage: userSlippage } = useSelectorStore(
+    swapView.store.configuration,
+    ['slippage']
+  );
+
+  const { default: defaultSlippagePercent } = getSlippageOptions(chain);
```
zerts: As I understand it, we can just use `slippage` from `getSlippageOptions` here

PR #720, opened 2025-02-24
eeeeDetect and analyze background script reload events

everdimension:


src/background/account/Account.ts
```diff
@@ -46,27 +61,29 @@ export class Account extends EventEmitter<AccountEvents> {
   isPendingNewUser: boolean;
 
   private static async writeCurrentUser(user: User) {
-    await browserStorage.set(currentUserKey, user);
+    await BrowserStorage.set(currentUserKey, user);
```
vyorkin, line 64: `User` data contains `salt`:
```
export interface User {
  id: string;
  salt: string;
}
```
I'm not sure how risky it is to store the `salt` in browser storage or if we can keep user data in session storage

vyorkin (approved):

src/background/account/Account.ts
everdimension: Encrypted storage cannot be unlocked without it, this is part of necessary data

src/background/account/Account.ts
```diff
@@ -46,27 +61,29 @@ export class Account extends EventEmitter<AccountEvents> {
...
+    return await SessionStorage.get<Credentials>(credentialsKey);
   }
 
   private static async removeCredentials() {
-    return await browserStorage.remove(credentialsKey);
+    await SessionStorage.remove(credentialsKey);
+    await LoginActivity.recordLogout();
   }
 
   static async readCurrentUser() {
```
zerts, line 81: Let's move these 2 functions to the top to place it near the `writeCurrentUser`?

src/background/account/Account.ts
```diff
@@ -152,6 +169,16 @@ export class Account extends EventEmitter<AccountEvents> {
     await this.setUser(user, { password }, { isNewUser: false });
   }
 
+  isAuthenticated(): boolean {
```
zerts, line 172: Let's move this function below the getUser?

src/shared/analytics/analytics.background.ts
```diff
@@ -331,6 +334,46 @@ function trackAppEvents({ account }: { account: Account }) {
...
+    const isAuthenticated = account.isAuthenticated();
+    const { loggedInAt } = await LoginActivity.getState();
+    const didLogoutIntentionally = loggedInAt == null;
+
+    if (didNotOnboardYet || isAuthenticated || didLogoutIntentionally) {
+      return;
+    }
+
+    const sessionExpiry = await estimateSessionExpiry();
+    const { previousVersion } = packageVersionStore;
```
zerts, line 365: Should we use getState here?

zerts: Ah, I see

src/background/account/Account.ts
vyorkin: Is it possible to make the `salt` a hash of the password instead of using `crypto.getRandomValues(new Uint8Array(32))`, so that we wouldn’t need to store it in browser storage (it could be stored in the session store).
But I’m still not sure if this is a good idea

src/ui/App/RouteRestoration.ts
```diff
@@ -10,7 +10,7 @@ export function registerPersistentRoute(value: string) {
 }
 
 export async function resetPersistedRoutes() {
-  return browserStorage.remove('routeRestoration');
+  return BrowserStorage.remove('routeRestoration');
```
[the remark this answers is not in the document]
everdimension: Good idea!

src/shared/analytics/analytics.background.ts
```diff
@@ -331,6 +334,46 @@ function trackAppEvents({ account }: { account: Account }) {
...
+    // The browser restart is considered an expected logout.
+    const { startupEvent, installedEvent } = runtimeStore.getState();
+
+    const isIntentionalBrowserRestart = Boolean(startupEvent);
+
+    const likelyReason = installedEvent
+      ? installedEvent.reason
+      : isIntentionalBrowserRestart
+      ? 'browser restart'
+      : 're-enabled by user';
```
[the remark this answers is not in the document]
everdimension: I expect such reloads to be caught by `onInstalled` listener and come with `installedEvent.reason`

src/shared/packageVersion.ts
```diff
@@ -3,3 +3,4 @@ import { version as packageVersion } from '../../package.json';
 const isProd = process.env.NODE_ENV === 'production';
 
 export const version = isProd ? packageVersion : `${packageVersion}-dev`;
+export const productionVersion = packageVersion;
```
zerts, line 6: Why do we need to clone this variable and not use `packageVersion` everywhere?

src/background/user-activity.ts
```diff
@@ -34,13 +34,27 @@ async function getIdleTimeout() {
...
   const lastActive = await getLastActive();
   const autoLockTimeout = await getIdleTimeout();
   if (autoLockTimeout === 'none') {
-    return;
+    return { autoLockTimeout, shouldExpire: false, timeToExpiry: Infinity };
+  }
+  if (lastActive != null) {
+    const timeToExpiry = autoLockTimeout - (Date.now() - lastActive);
+    return { autoLockTimeout, shouldExpire: timeToExpiry < 0, timeToExpiry };
+  } else {
```
zerts, line 50: We don't need `else` here, but this is a minor

zerts: We can also combine `autoLockTimeout === 'none' && lastActive == null` in one condition

src/shared/packageVersion.ts
everdimension: Well,
1. I want this place to be the only file that accesses `package.json` because it's not common
2. If we export `version` and `packageVersion`, I think it will be hard to understand difference between them. But `productionVersion` at least implies some caution in order to use it.

PR #722, opened 2025-03-03
Asset page v1

zerts:


LoznianuAnamaria:
@vyorkin @everdimension Could you please review this PR?

src/ui/components/AssetLink/AssetLink.tsx
```diff
@@ -3,10 +3,38 @@ import React from 'react';
...
   asset,
   title,
+}: {
+  asset: Pick<Asset, 'symbol' | 'name' | 'asset_code'>;
+  title?: string;
+}) {
+  const { preferences } = usePreferences();
+  const content = title || asset.symbol || asset.name;
+  if (preferences?.testnetMode?.on) {
+    return content;
```
vyorkin, line 18: Does it make sense to apply the same style when we're in `testnetMode`? I mean `overflow: hidden`, etc

src/ui/pages/AssetInfo/AssetDescription.tsx
```diff
@@ -0,0 +1,121 @@
...
+  // on Safari line-camp works correctly only with inline elements inside
+  const paragraphs = text.split(/(\r?\n|\r)+/g);
+  const markup: React.ReactNode[] = [];
+  paragraphs.forEach((block, index, array) => {
+    markup.push(<span key={index}>{block}</span>);
+    markup.push(<br key={array.length + index} />);
+  });
+
+  useEffect(() => {
+    if (ref.current) {
```
vyorkin, line 29: Looks like it could be possible to just:
`if (ref.current && ref.current.scrollHeight > ref.current.clientHeight) { ... `

src/ui/pages/AssetInfo/AssetGlobalStats.tsx
```diff
@@ -0,0 +1,138 @@
...
+      ) : null}
+      {assetFullInfo.fungible.meta.fullyDilutedValuation != null ? (
+        <AssetStatsChip
+          title="FDV"
+          fullTitle="Fully Diluted Valuation"
+          value={formatCurrencyValue(
+            assetFullInfo.fungible.meta.fullyDilutedValuation,
+            'en',
+            currency,
+            {
```
vyorkin, line 70: The formatting options are repeated multiple times in this file. We could consider extracting these options into a constant

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  return (
+    <VStack gap={8} style={{ opacity: actionsAreLoading ? 0.8 : 1 }}>
+      <VStack gap={4}>
+        <UIText kind="headline/h3">History</UIText>
+        <VStack gap={0}>
+          {value.map((action) => (
+            <AssetHistoryItem
+              key={action.id}
```
vyorkin, line 244: Sometimes in dev env I got this error
```
Encountered two children with the same key, `34fb69bd6e4d56b4841b013359984a72`
```
Probably because there exist duplicate action.id's

vyorkin (commented):
Screen Recording 2025-03-25 at 16.42.51 Screen jumps when clicking the "More Transactions" button

vyorkin:
Focus outline disappears on hover 

https://github.com/user-attachments/assets/f1d07c8e-f106-4079-9e70-2bd66656da87

vyorkin:
Not sure if it makes sense to save scroll position here 

https://github.com/user-attachments/assets/e3626f93-e58d-4d88-afdf-fca6fe3edbf4

src/ui/pages/AssetInfo/AssetDescription.tsx
```diff
@@ -0,0 +1,121 @@
...
+    <VStack gap={12}>
+      <UIText kind="headline/h3">About {assetFullInfo.fungible.name}</UIText>
+      <VStack gap={8}>
+        {assetFullInfo.extra.description ? (
+          <TextPreview text={assetFullInfo.extra.description} />
+        ) : null}
+        <div style={{ display: 'flex', flexWrap: 'wrap' }}>
+          {assetFullInfo.extra.relevantResources.map((resource) => (
+            <TextAnchor
+              key={resource.name}
```
vyorkin, line 103: Could you please double check it safe to use the `resource.name` as key here?
I see warnings like these in dev console:
```
react-dom.development.js:86 Warning: Encountered two children with the same key, `dex.guru`
```
stacktrace:
```
t div
    at div
    at VStack (chrome-extension://mpiiagdnopgpnfbhnpkgmdbeehgnmchj/popup.70bf0a75.js:57083:19)
    at div
    at VStack (chrome-extension://mpiiagdnopgpnfbhnpkgmdbeehgnmchj/popup.70bf0a75.js:57083:19)
    at AssetDescription (chrome-extension://mpiiagdnopgpnfbhnpkgmdbeehgnmchj/popup.70bf0a75.js:234554:29)
```

src/ui/components/AssetLink/AssetLink.tsx
zerts: Hm. Makes sense

zerts: But here I just copied previous implementation. Let's keep it this way and fix if needed in the future separately

src/ui/pages/AssetInfo/AssetHistory.tsx
zerts: changed to hash. Hopefully, this should be more reliable

src/ui/pages/AssetInfo/AssetDescription.tsx
zerts: Hm, this should be reported to backend, I suppose. All names should be different for sure

src/ui/components/AllowanceForm/AllowanceForm.tsx
```diff
@@ -19,13 +19,13 @@ import type { Chain } from 'src/modules/networks/Chain';
...
 import { Content } from 'react-area';
 import { Button } from 'src/ui/ui-kit/Button';
-import { AssetLink } from 'src/ui/components/AssetLink';
 import { formatCurrencyValue } from 'src/shared/units/formatCurrencyValue';
 import { isUnlimitedApproval } from 'src/ui/pages/History/isUnlimitedApproval';
 import { UnstyledButton } from 'src/ui/ui-kit/UnstyledButton';
 import { UNLIMITED_APPROVAL_AMOUNT } from 'src/modules/ethereum/constants';
 import { focusNode } from 'src/ui/shared/focusNode';
 import { useCurrency } from 'src/modules/currency/useCurrency';
+import { AssetAnchor } from '../AssetLink';
```
everdimension, line 28: why a change of path?

zerts: Oops, autocomplete + few iterations

src/modules/zerion-api/requests/wallet-get-asset-details.ts
```diff
@@ -0,0 +1,69 @@
...
+      iconUrl: string;
+      premium: boolean;
+      address: string;
+    };
+    chains: NetworkShortInfo[];
+    value: number;
+    convertedQuantity: number;
+    percentageAllocation: number;
+  }[];
+  apps: {
```
everdimension, line 38: `Array<...>` seems more readable here... :)

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+                src={network.icon_url}
+                alt={network.name}
+                width={36}
+                height={36}
+              />
+              <VStack gap={0}>
+                <UIText kind="body/accent">{network.name}</UIText>
+                {address ? (
+                  <UIText kind="small/regular" color="var(--neutral-500)">
+                    {middleTruncate({ value: address })}
```
everdimension, line 179: Why not use existing `truncateAddress`?
image

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+                  </UIText>
+                ) : null}
+              </VStack>
+            </HStack>
+            <HStack gap={0}>
+              {address ? <CopyAddressButton address={address} /> : null}
```
everdimension, line 185: missing `title` for accessibility

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+              <UnstyledAnchor
+                rel="noopener noreferrer"
+                target="_blank"
+                href={
```
everdimension, line 189: missing `title` for accessibility

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+              action={action}
+            />
+          ))}
+        </VStack>
+      </VStack>
+      {hasNext ? (
+        <Button
+          kind="neutral"
+          onClick={fetchMore}
+          disabled={actionsAreLoading}
```
everdimension, line 258: Missing visual change for `actionsAreLoading` state

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+          style={{
+            ['--button-background' as string]: 'var(--neutral-200)',
+            ['--button-background-hover' as string]: 'var(--neutral-300)',
+          }}
+        >
+          <HStack gap={8} alignItems="center" justifyContent="center">
+            <UIText kind="body/accent">More Transactions</UIText>
+            {actionsAreLoading ? <CircleSpinner /> : null}
```
everdimension, line 266: weird, I don't see this spinner (https://github.com/zeriontech/zerion-wallet-extension/pull/722/files#r2018965737)

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+export function AssetInfo() {
+  const { asset_code } = useParams();
+  const navigationType = useNavigationType();
+  useEffect(() => {
+    if (navigationType === NavigationType.Push) {
+      window.scrollTo(0, 0);
+    }
+  }, [navigationType]);
+  useBackgroundKind({ kind: 'white' });
```
everdimension, line 95:
```suggestion
  useBackgroundKind(whiteBackground);
```

`whiteBackgroundKind` is an existing export alongside `useBackgroundKind`

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+      addresses: [singleAddressNormalized],
+    },
+    { source: useHttpClientSource() },
+    { enabled: ready }
+  );
+
+  const { data: wallet } = useQuery({
+    queryKey: ['wallet/uiGetWalletByAddress', singleAddress, null],
+    queryFn: () =>
+      walletPort.request('uiGetWalletByAddress', {
```
everdimension, line 128: can't `uiGetCurrentWallet` be used here?

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+      return NetworkId.Zero;
+    }
+    return chainData.reduce(
+      (acc, chain) =>
+        chain.value > acc.value
+          ? { value: chain.value, chain: chain.chain.id }
+          : acc,
+      { value: chainData[0].value ?? 0, chain: chainData[0].chain.id }
+    ).chain;
+  }, [walletData]);
```
everdimension, line 148:
```suggestion
  const chainWithTheBiggestBalance = walletData?.data?.chainsDistribution.at(0)?.chain.id || NetworkId.Zero;
```

It looks as if backend returns this list already sorted.

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  if (!assetFullInfo?.fungible || !wallet || !walletData) {
```
everdimension, line 150: If we have received `assetFullInfo`, but it doesn't have the `fungible` prop, should we show an error state or an empty state instead of a loading state?

src/ui/pages/AssetInfo/AssetAddressDetails.tsx
```diff
@@ -0,0 +1,740 @@
...
+    <div
+      style={{
+        width: 100,
+        height: 24,
+        backgroundColor: 'var(--neutral-200)',
+        borderRadius: 12,
+      }}
+    />
+  );
+}
```
everdimension, line 64: Came up with a slightly better design:
```suggestion
function LoadingSkeleton() {
  return (
    <div
      style={{
        width: 100,
        height: 24,
      }}
    />
  );
}
```

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+import { AssetHistory } from './AssetHistory';
+import { AssetAddressStats } from './AssetAddressDetails';
+import { AssetGlobalStats } from './AssetGlobalStats';
+import { AssetTitleAndChart } from './AssetTitleAndChart';
+import { AssetResources } from './AssetResources';
+import { AssetHeader } from './AssetHeader';
+import { AssetDescription } from './AssetDescription';
+
+const SCROLL_THRESHOLD = 80;
+const SHOW_BRIDGE_BUTTON = false; // TODO: make true after bridge is implemented
```
everdimension, line 41: 👍

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+export function AssetInfo() {
+  const { asset_code } = useParams();
+  const navigationType = useNavigationType();
+  useEffect(() => {
+    if (navigationType === NavigationType.Push) {
+      window.scrollTo(0, 0);
+    }
+  }, [navigationType]);
+  useBackgroundKind({ kind: 'white' });
+  invariant(asset_code, 'Asset Code is required');
```
everdimension, line 96: I prefer to put these assertions next to where the variable is introduced

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+            alt={network.name}
+            width={24}
+            height={24}
+          />
+          <UIText kind="small/accent">
+            {middleTruncate({
+              value: address,
+              leadingLettersCount: 5,
+              trailingLettersCount: 4,
+            })}
```
everdimension, line 104: Is this better for some reason than `truncateAddress`?

zerts: I need to check. Tbh, I just forgot about other ways to truncate addresses

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+      {isSuccess ? (
+        <CheckIcon style={{ color: 'var(--positive-500)' }} />
+      ) : (
+        <CopyIcon className="content-hover" />
+      )}
+    </UnstyledButton>
+  );
+}
+
+function AssetImplementationsDialog({
```
everdimension, line 145:
```suggestion
function AssetImplementationsDialogContent({
```
Perhaps this is a better name? Since this component doesn't render the actual dialog component

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+              <UnstyledAnchor
+                rel="noopener noreferrer"
+                target="_blank"
+                href={
+                  address && networks
+                    ? networks.getExplorerTokenUrlByName(
+                        createChain(network.id),
+                        address
+                      )
+                    : network.explorer_home_url || ''
```
everdimension, line 195: Maybe this href can be calculated above, and we shouldn't render the link at all if the href is falsy?

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+              </UnstyledAnchor>
+            </HStack>
+          </HStack>
+        ))}
+      </VStack>
+    </VStack>
+  );
+}
+
+function ResourseButton({
```
everdimension, line 215:
```suggestion
function ResourceButton({
```

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+      kind="regular"
+      size={36}
+      href={resource.url}
+      style={{
+        padding: 6,
+        border: '2px solid var(--neutral-200)',
+        ['--button-background' as string]: 'var(--white)',
+        ['--button-background-hover' as string]: 'var(--neutral-200)',
+      }}
+      aria-label={resource.displayableName}
```
everdimension, line 237: I think `title` is still better here, because I didn't understand what the icon was :)

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+        .filter(isTruthy),
+    [assetFullInfo.fungible.implementations, networks]
+  );
+
+  const mainImplementation = useMemo(() => {
+    return (
+      implementations.find(
+        (implementation) =>
+          implementation.network.id === assetFullInfo.extra.mainChain
+      ) || implementations[0]
```
everdimension, line 271:
```suggestion
      ) || implementations.at(0)
```

this will type the value as optional

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+              }
+            }
+            icon={<DexscreenerIcon style={{ width: 20, height: 20 }} />}
+          />
+        </HStack>
+        <HStack gap={4}>
+          {mainImplementation ? (
+            <MainImplementationButton implementation={mainImplementation} />
+          ) : null}
+          {implementations.length ? (
```
everdimension, line 314: maybe you meant this?
```suggestion
          {implementations > 1 ? (
```
It seems that `implementations` can't be empty if `mainImplementation` is defined

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,365 @@
...
+            >
+              <Button
+                kind="ghost"
+                value="cancel"
+                size={36}
+                style={{ width: 36, padding: 8 }}
+                onClick={() => dialogRef.current?.close()}
+              >
+                <ArrowLeftIcon style={{ width: 20, height: 20 }} />
+              </Button>
```
everdimension, line 354: Technically html dialogs are designed like this:
```suggestion
              <form method="dialog" onSubmit={(event) => event.stopPropagation()}>
                <Button
                  kind="ghost"
                  value="cancel"
                  size={36}
                  style={{ width: 36, padding: 8 }}
                >
                  <ArrowLeftIcon style={{ width: 20, height: 20 }} />
                </Button>
              </form>
```

Not crucial, but this makes this component decoupled from the `dialogRef` variable

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const actionValue = normalizedAmountAction
+    ? formatCurrencyValue(normalizedAmountAction, 'en', currency)
+    : null;
+
+  return (
+    <>
+      <KeyboardShortcut
+        combination="backspace"
+        onKeyDown={handleDialogDismiss}
+      />
```
everdimension, line 141: This conflicts with the Navigation Bar's backspace handler. But do we need it here? I think escape is enough

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+  assetId,
+  address,
+  assetFullInfo,
+}: {
+  assetId: string;
+  address: string;
+  assetFullInfo?: AssetFullInfo;
+}) {
+  const { networks } = useNetworks();
+  const { singleAddressNormalized, ready } = useAddressParams();
```
everdimension, line 210: I think `address` and `singleAddressNormalized` are being mixed up. It seems `address` from props should be enough

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const { currency } = useCurrency();
+  const {
+    value,
+    // TODO: this flag doesn't work, needs to be fixed
+    isFetching: actionsAreLoading,
```
everdimension, line 215: we can hack up a solution with `useInfiniteQuery` to fix this, wdyt? Check out `ImportLedgerAddresses.tsx` for reference

src/ui/pages/AssetInfo/AssetDescription.tsx
everdimension: What's the outcome?

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,272 @@
...
+    : null;
+
+  return (
+    <>
+      <KeyboardShortcut
+        combination="backspace"
+        onKeyDown={handleDialogDismiss}
+      />
+      <UnstyledButton onClick={handleDialogOpen} className={styles.historyItem}>
+        <div className={styles.historyItemBackdrop} />
```
everdimension, line 143: Why not reuse `PageFullBleed` + `SurfaceList`? Then you wouldn't need this helper

src/ui/pages/AssetInfo/AssetResources.tsx
zerts: Forgot about this function. Thank you!

src/ui/pages/AssetInfo/AssetResources.tsx
zerts: Technically, CopyAddressButton, has aria-label.
But I can add title too

src/ui/pages/AssetInfo/AssetHistory.tsx
zerts: Yes, it looks like a problem with defi-sdk. Need to investigate.

src/ui/pages/AssetInfo/AssetInfo.tsx
zerts: Oh, nice! Forgot about this one

src/ui/pages/AssetInfo/AssetInfo.tsx
zerts: Whoa! Nice, thank you!

src/ui/pages/AssetInfo/AssetResources.tsx
zerts: This one is a thing to be improved in the future. Right now I show this button even if length = 1, cause in the dialog there is also an option to open the link in explorer, not only copy the address on chain

src/ui/pages/AssetInfo/AssetResources.tsx
zerts: Interesting!

src/ui/pages/AssetInfo/AssetHistory.tsx
zerts: Good idea, let me try!

zerts: I fixed this partially. 
Using SurfaceList is hard the same way it was in the main History.
Because of the Dialog associated with every item of the list, it is hard to use onClick outside of the component, so by default items will not be wrapped with the decorations which is the main point of the SurfaceList at this point.
+ It looks like we will need to update this list soon too and design will be updated significantly

src/ui/pages/AssetInfo/AssetResources.tsx
everdimension: idea: maybe not truncate here at all (this needs a slight design update)

src/ui/pages/AssetInfo/AssetResources.tsx
everdimension: if `title` is added, `aria-label` is no longer needed

src/ui/pages/AssetInfo/AssetDescription.tsx
zerts: I didn't face any problems so far. But if I do, I will report

src/ui/pages/AssetInfo/AssetResources.tsx
everdimension: Ah, sorry, didn't get your question

For `CopyAddressButton` aria-label should be enough, because visual users recognize the common copy icon

But for the explorer name it's hard to understand from the icon the purpose of the link :)

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,370 @@
...
+    >
+      <VStack gap={0}>
+        {implementations.map(({ address, network }) => {
+          const explorerHref =
+            address && networks
+              ? networks.getExplorerTokenUrlByName(
+                  createChain(network.id),
+                  address
+                )
+              : network.explorer_home_url || '';
```
everdimension, line 161: can we fallback to `null`? Or maybe no explicit fallback is needed at all here?

src/ui/pages/AssetInfo/AssetResources.tsx
everdimension: Makes sense!

src/ui/pages/AssetInfo/AssetResources.tsx
```diff
@@ -0,0 +1,371 @@
...
+              resource={resourcesById[TWITTER_ID]}
+              icon={<XIcon style={{ width: 20, height: 20 }} />}
+            />
+          ) : null}
+          {resourcesById[WARPCAST_ID] ? (
+            <ResourceButton
+              resource={resourcesById[WARPCAST_ID]}
+              icon={<WarpcastIcon style={{ width: 20, height: 20 }} />}
+            />
+          ) : null}
```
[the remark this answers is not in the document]
zerts: Yes. We've agreed to skip such buttons for now

src/ui/pages/AssetInfo/AssetInfo.tsx
```diff
@@ -0,0 +1,253 @@
...
+            }}
+          >
+            <Button
+              kind="primary"
+              size={48}
+              as={UnstyledLink}
+              to={
+                isEmptyBalance
+                  ? `/swap-form?chainInput=${chainForSwap}&receiveTokenInput=${asset_code}`
+                  : `/swap-form?chainInput=${chainForSwap}&spendTokenInput=${asset_code}`
```
everdimension, line 215: Doesn't SwapForm determine the correct chain itself internally?

zerts: Let me check.

src/ui/pages/AssetInfo/AssetHistory.tsx
```diff
@@ -0,0 +1,264 @@
...
+        ? {
+            ...outgoingTransfers[0],
+            quantity: outgoingTransfers
+              .reduce((acc, item) => acc.plus(item.quantity), new BigNumber(0))
+              .toFixed(),
+          }
+        : undefined;
+    return {
+      transfer: aggregatedIncomingTransfer || aggregatedOutgoingTransfer,
+      isIncoming: Boolean(aggregatedIncomingTransfer),
```
everdimension, line 86: isn't this same as 
```suggestion
      isIncoming: Boolean(incomingTransfers?.length),
```
?

everdimension (approved):

src/ui/pages/AssetInfo/AssetInfo.tsx
everdimension: Ok cool!

src/ui/pages/AssetInfo/AssetHistory.tsx
zerts: Yes, this is also ok. Not sure that this is much easier, but I can change it

everdimension: I mean, it doesn't rely on a derived value, so if you later remove `aggregatedIncomingTransfer`, you will still be able to keep `isIncoming`

src/ui/pages/AssetInfo/AssetAddressDetails.tsx
```diff
@@ -0,0 +1,742 @@
...
+        </UnstyledButton>
+      </div>
+
+      <div
+        style={{
+          overflow: 'hidden',
+          maxHeight: showNetworkDistribution
+            ? (walletAssetDetails.chainsDistribution?.length || 0) * 40
+            : 0,
+          transition: 'max-height 0.3s ease-in-out',
```
everdimension, line 321: does `maxHeight: auto` work with transitions yet? :)

PR #723, opened 2025-03-05
Clear pending transactions

everdimension:


src/ui/pages/Settings/Settings.tsx
```diff
@@ -340,6 +343,44 @@ function ToggleSettingLine({
...
+        disabled={mutation.isLoading}
+        style={{ padding: 20 }}
+      >
+        <Media
+          image={null}
+          text={<UIText kind="body/accent">Clear Pending Transactions</UIText>}
+          vGap={0}
+          detailText={
+            <UIText kind="small/regular" color="var(--neutral-500)">
+              This can fix stuck transactions
```
zerts, line 375: I'm not sure what I feel about this description. Not sure that this actually FIXES anything :)

src/ui/pages/Settings/Settings.tsx
```diff
@@ -340,6 +343,44 @@ function ToggleSettingLine({
...
+      <PopoverToast ref={toastRef}>Pending transactions cleared</PopoverToast>
+
+      <Frame
+        as={UnstyledButton}
+        interactiveStyles={true}
+        onClick={() => clearPendingTransactions()}
+        disabled={mutation.isLoading}
+        style={{ padding: 20 }}
+      >
+        <Media
```
zerts, line 369: Do we need Media here?

src/ui/pages/Settings/Settings.tsx
```diff
@@ -340,6 +343,44 @@ function ToggleSettingLine({
...
+      toastRef.current?.showToast();
+    },
+  });
+  return (
+    <>
+      <PopoverToast ref={toastRef}>Pending transactions cleared</PopoverToast>
+
+      <Frame
+        as={UnstyledButton}
+        interactiveStyles={true}
```
zerts, line 364: May be we can name this just `interactive`? Usually we use ___Styles for CSSProperties param

src/ui/ui-kit/Frame/styles.module.css
```diff
@@ -0,0 +1,20 @@
...
+  padding: 8px;
+}
+
+@media (hover: hover) {
+  .interactive:enabled:hover {
+    background-color: var(--neutral-100);
+  }
+}
+
+.interactive:enabled:focus {
```
zerts, line 14: It looks like we show outline on focus anyway. Do we need to change background color too?

zerts (approved):

src/ui/ui-kit/Frame/styles.module.css
everdimension: 🤔 i think you're right, we do not

src/ui/pages/Settings/Settings.tsx
everdimension: Yeah not really. I reused the layout of the `ToggleSettingLine` but I guess I'll refactor it, too, then

src/ui/pages/Settings/Settings.tsx
vyorkin: I would leave the name as it is. Ideally, if `interactiveStyles` doesn't make sense without having `as={...Button}` or `as={...Link}` then it should be impossible to construct such a `Frame` component (I mean it shouldn't type-check in this case)

PR #732, opened 2025-03-15
Cloudflare proof of concept

zerts:


src/modules/phishing-defence/phishing-defence-service.ts
```diff
@@ -97,6 +97,9 @@ class PhishingDefence {
       };
     }
     const origin = url ? this.getSafeOrigin(url) : null;
+    if (origin && this.websiteStatus[origin] === 'error') {
+      await this.checkDapp(url);
+    }
```
everdimension, line 102: Can you please explain these lines? I don't understand at all what this does and why the condition is like this

src/ui/features/turnstile/Turnstile.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const turnstileWidgetWidth = innerWidth - 32;
+  const [showDialog, setShowDialog] = useState(false);
+  const dialogRef = useRef<HTMLDialogElementInterface | null>(null);
+
+  useEffect(() => {
+    return emitter.on('openTurnstile', () => {
+      setShowDialog(true);
+      dialogRef.current?.showModal();
+    });
+  }, []);
```
everdimension, line 23: We shouldn't be calling any unnecessary hooks and rendering any unnecessary jsx unless we are reacting to this event. The component should be split into two, and the outer one should only be listening to this event and rendering `null` until the event occurs.

src/ui/features/referral-program/WebAppMessageHandler.tsx
```diff
@@ -4,9 +4,9 @@ import { isObj } from 'src/shared/isObj';
 import { ZerionAPI } from 'src/modules/zerion-api/zerion-api.client';
 import { saveReferrerData } from './shared/storage';
 
-const ZERION_WEB_APP_URL = new URL('https://app.zerion.io');
+const ZERION_WEB_APP_URL = new URL('https://beta.zerion.io');
```
everdimension, line 7: is this temprorary?

src/ui/features/turnstile/Turnstile.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <BottomSheetDialog
+      ref={dialogRef}
+      height="fit-content"
+      closeOnClickOutside={false}
+      containerStyle={{ backgroundColor: 'var(--z-index-0)' }}
+    >
+      {showDialog ? (
```
everdimension, line 42: This is probably unnecessary because you can use `renderWhenOpen` prop of the `BottomSheetDialog` instead

src/ui/features/referral-program/WebAppMessageHandler.tsx
zerts: Yes!

src/ui/features/turnstile/Turnstile.tsx
zerts: Ah, yes, thank you! I missed this option when I developed this one.

src/ui/features/turnstile/Turnstile.tsx
zerts: Agree! I will update this!

src/modules/phishing-defence/phishing-defence-service.ts
zerts: Sure, let me update the pr and provide all context

everdimension: Thanks!

> // to provide the correct data after passing the captcha we need to check the status again

what is "correct data" here?

zerts: Hm. The actual data we have about the dapp, not error because of cloudflare block.

src/ui/features/referral-program/WebAppMessageHandler.tsx
```diff
@@ -33,10 +34,14 @@ export function WebAppMessageHandler({
   pathname,
   callbackName,
   callbackFn,
+  hidden,
+  style,
 }: {
   pathname: string;
   callbackName: WebAppCallbackMethod;
-  callbackFn: (params?: unknown) => Promise<void>;
+  callbackFn: (params?: unknown) => void | Promise<void>;
```
everdimension, line 42: Where is this new signature used?

src/modules/phishing-defence/phishing-defence-service.ts
everdimension: I finally figured out what this does.

So after completing the captcha we call "refetchQueries", and in case of `<PhishingDefenceStatus />` being open, the refetching will get old data because the result is outdated in `phishingDefenceService`.

**So this code serves as a "retry" in case of an error result**

 I guess there are other ways to achieve this but perhaps it's okay to leave it as is for now
 
 Ideally though request statuses are better organized by some caching lib

src/ui/features/turnstile/helpers.ts
```diff
@@ -0,0 +1,7 @@
+import { emitter } from 'src/ui/shared/events';
+
+export function openTurnstileWidgetIfNeeded(response: Response) {
+  if (response.headers.get('cf-mitigated') === 'challenge') {
+    emitter.emit('openTurnstile');
```
everdimension, line 5: Maybe a local emitter (created somewhere within `src/ui/features/turnstile/`) is more appropriate here, because the purpose of this emitter is not making the event globally known, but a mechanism to change the state of an existing React Component

everdimension (approved):

src/ui/features/referral-program/WebAppMessageHandler.tsx
everdimension: Ah, I thought you added the Promise type, but you added the `void` one. Np!

src/ui/features/turnstile/Turnstile.tsx
```diff
@@ -0,0 +1,62 @@
...
+      />
+    </BottomSheetDialog>
+  );
+}
+
+export function TurnstileTokenHandler() {
+  const [showDialog, setShowDialog] = useState(false);
+  useEffect(() => {
+    return emitter.on('openTurnstile', () => {
+      setShowDialog(true);
```
everdimension, line 57: Hm, do we never set this back to `false`? I guess it's minor but it feels incomplete:)

src/modules/zerion-api/zerion-api.client.ts
```diff
@@ -1,9 +1,19 @@
...
 import { getAddressProviderHeader } from './requests/shared.client';
 import type { ZerionApiContext } from './zerion-api-bare';
 import { ZerionApiBare } from './zerion-api-bare';
 
 const context: ZerionApiContext = {
   getAddressProviderHeader,
+  getKyOptions: () => ({
+    hooks: {
+      afterResponse: [
+        (_, __, response) => {
```
vyorkin, line 11: double underscore %)

everdimension: I think it's standard for a second unused param

PR #740, opened 2025-03-19
Add a warning for high value loss in swap

vyorkin:


src/ui/pages/SwapForm/shared/price-impact.ts
```diff
@@ -1,11 +1,84 @@
...
+    }
+  | {
+      kind: 'loss';
+      level: PriceImpactLevel;
+      ratio: BigNumber;
+      percentage: string;
+    };
+
+function toPercentage(ratio: BigNumber) {
+  return `${ratio.isLessThan(0) ? '' : '+'}${formatPercent(
```
zerts, line 29: should we use `minus` here and format the absolute value?

src/ui/pages/SwapForm/shared/price-impact.ts
```diff
@@ -1,11 +1,84 @@
...
 
-export function exceedsPriceImpactThreshold({
-  relativeChange,
+type PriceImpactLevel = 'low' | 'medium' | 'high';
+
+export type PriceImpact =
+  | { kind: 'n/a' }
+  | {
+      kind: 'profit';
+      ratio: BigNumber;
```
zerts, line 15: If I understand it correctly, we don't need `ratio` field in the result object.

zerts: Also what do you think about just make object with 2 props:
```
{
  kind: 'loss' | 'big-loss' | 'neutral'
  percentage: number
}
```
?

It should cover all cases from the task with less cases in types.

src/ui/pages/SwapForm/shared/price-impact.ts
vyorkin: positive numbers are positive by default/implicitly (I mean when we see numbers like 5, 7, 123 they are assumed to be positive so we must add the + sign here), but negative number always prefixed with the `-` sign to be considered negative

zerts: + it is better to do all formatting in the JSX. Cause one day in the future we will support different locales and passing locale to these inner functions will be a bit of a challenge

src/ui/pages/SwapForm/shared/price-impact.ts
vyorkin: I chose to retain the `ratio` number as it provides context for understanding the derivation of percentage and other values. It serves as the source data for the percentage (which is primarily for visual purposes rather than conditional logic). Since it's simply a numeric value (carrying it around should't impact performance that much) and might be useful for potential future conditional logic, I decided to keep it

vyorkin: just read [this comment](https://github.com/zeriontech/zerion-wallet-extension/pull/740#discussion_r2007847769) and now I'm thinking about removing `percentage` formatted value, but keeping the `ratio` :)

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -154,7 +160,10 @@ function FormHint({
     : null;
   const exceedsBalance = Number(spendInput) > Number(positionBalanceCommon);
 
-  let message: string | null = null;
+  const isHighPriceImpact =
+    priceImpact?.kind === 'loss' && priceImpact.level === 'high';
```
zerts, line 164: Based on the design the warning on the swap button should be shown when impact > 5%

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -261,6 +281,25 @@ export function SwapFormComponent() {
     spendPosition,
   ]);
 
+  const { receiveAsset, spendAsset } = swapView;
+
+  const priceImpact = useMemo(() => {
```
zerts, line 286: Let's extract this to the separate file

zerts: Ok, I see, not critical

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -840,7 +883,12 @@ export function SwapFormComponent() {
             </React.Suspense>
           ) : null}
         </VStack>
-        {quote ? <ProtocolFeeLine quote={quote} /> : null}
+        <VStack gap={16}>
+          {quote ? <ProtocolFeeLine quote={quote} /> : null}
+          {priceImpact && !isApproveMode ? (
```
zerts, line 888: isn't if(priceImpact) always true?

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
```diff
@@ -3,19 +3,28 @@ import type { SwapFormView } from '@zeriontech/transactions';
...
 import React from 'react';
 import { formatCurrencyValue } from 'src/shared/units/formatCurrencyValue';
-import { formatPercent } from 'src/shared/units/formatPercent/formatPercent';
 import { UIText } from 'src/ui/ui-kit/UIText';
 import { isNumeric } from 'src/shared/isNumeric';
 import { useCurrency } from 'src/modules/currency/useCurrency';
-import { exceedsPriceImpactThreshold } from '../../shared/price-impact';
+import type { PriceImpact } from '../../shared/price-impact';
+
+interface FiatInputValueProps {
```
zerts, line 11: I can see potential merge conflicts with bridge form. We should merge this carefully

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
```diff
@@ -41,40 +48,59 @@ export function FiatInputValue({
...
+  const isPriceImpactWarning = isSignificantPriceImpact(priceImpact);
 
   return (
-    <UIText
-      kind="small/regular"
+    <FiatInputValue
+      name="receiveInput"
+      swapView={swapView}
+      percentageChangeText={
+        priceImpact?.kind === 'loss' || priceImpact?.kind === 'profit'
```
zerts, line 91: I think, we can show 0% too

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
```diff
@@ -41,40 +48,59 @@ export function FiatInputValue({
...
+      {isPrimaryInput ? null : '≈'}
+      {formatCurrencyValue(fiatValue, 'en', currency)} {percentageChangeText}
+    </UIText>
+  );
+}
 
-  const formattedDiff = diff
-    ? `${diff.isLessThan(0) ? '' : '+'}${formatPercent(diff.times(100), 'en')}%`
-    : null;
+export function SpendFiatInputValue({ swapView }: { swapView: SwapFormView }) {
```
zerts, line 59: Do we need this wrapper?

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
vyorkin: There is another task about that "Do not display in the swap form the fiat value of tokens for whhich we don't have price" :)

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
vyorkin: Just though it reads better, I mean having `SpendFiatInputValue` / `ReceiveFiatInputValue` instead of just `FiatIputValue` and `ReceiveFiatInputValue`

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
```diff
@@ -3,19 +3,28 @@ import type { SwapFormView } from '@zeriontech/transactions';
...
 
 export function FiatInputValue({
-  swapView,
   name,
-}: {
-  swapView: SwapFormView;
-  name: 'spendInput' | 'receiveInput';
-}) {
+  swapView,
+  percentageChangeText,
```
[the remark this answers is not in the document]
vyorkin: Yeah, I already improved this a bit, as I mentioned [here](https://github.com/zeriontech/zerion-wallet-extension/pull/740#discussion_r2007896396)

src/ui/pages/SwapForm/shared/price-impact.ts
vyorkin: updated ✅

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: good catch, updated ✅

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: yeah, I already extracted a separate function `calculatePriceImpact`, but wanted to keep all mentions of `primaryInput` here, in SwapForm

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
vyorkin: updated ✅

src/shared/units/formatPercent/formatPercentChange.ts
```diff
@@ -0,0 +1,22 @@
...
+export function formatPercentChange(
+  value: number,
+  locale: string
+): PercentChangeInfo {
+  return {
+    isPositive: value > 0,
+    isNonNegative: value >= 0,
+    isNegative: value < 0,
+    isZero: value === 0,
+    formatted: `${formatPercent(value, locale)}%`,
```
zerts, line 20: We tend to format negative values as `${minus}${format(Math.abs(value))}`
May be here, if we show minus with percent value, this is relevant too :)

zerts: I see that this is just a copy of code from before. Then this is not critical at all

src/ui/pages/SwapForm/shared/price-impact.ts
```diff
@@ -1,11 +1,86 @@
...
+  // -0.15 <= x < -0.05 => medium loss
+  // x < -0.15 => high loss
+
+  if (ratio == null) {
+    return { kind: 'n/a' };
+  } else if (ratio > 0) {
+    return { kind: 'profit', ratio };
+  } else if (ratio === 0) {
+    return { kind: 'zero' };
+  } else if (ratio >= LOW_LOSS_RATIO) {
```
zerts, line 38: shouldn't there be ratio <= LOW_LOSS_RATIO?

src/ui/pages/SwapForm/fieldsets/FiatInputValue/FiatInputValue.tsx
```diff
@@ -37,44 +49,76 @@ export function FiatInputValue({
...
+    <FiatInputValue
+      name="receiveInput"
+      swapView={swapView}
+      percentageChange={
+        showPercentageChange ? (
+          <PercentChange
+            value={priceImpactPercentage}
+            locale="en"
+            render={(change) => (
+              <UIText kind="small/regular" color="var(--negative-500)">
```
zerts, line 108: negative color should be used only for big loss 
Right now it looks like this sometimes :)
Zerts Screenshot 2025-03-25 at 10 18@2x

vyorkin: fixed ✅

src/shared/units/formatPercent/formatPercentChange.ts
vyorkin: As I can see `formatPercent` does what I mentioned:
```

export function formatPercent(value: BigNumber.Value, locale: string) {
  const valueAsNumber = toNumber(value);
  const formatter =
    valueAsNumber < 1
      ? getFormatterTwoDigits(locale)
      : getFormatterOneDigit(locale);
  const sign = valueAsNumber < 0 ? minus : '';
  return `${sign}${formatter.format(Math.abs(valueAsNumber))}`;
}

```

src/ui/components/PercentChange/PercentChange.tsx
```diff
@@ -0,0 +1,17 @@
...
+}: {
+  value: number | null;
+  locale: string;
+  render: (changeInfo: PercentChangeInfo) => JSX.Element;
+}): JSX.Element | null {
+  if (value == null) {
+    return null;
+  }
+  return render(formatPercentChange(value, locale));
+}
```
everdimension, line 17: What's the point of this component?

vyorkin: moped is not mine :) but good point, I'll remove it right now

zerts (approved):

src/ui/components/PercentChange/PercentChange.tsx
vyorkin: updated ✅

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -155,17 +161,31 @@ function FormHint({
     : null;
   const exceedsBalance = Number(spendInput) > Number(positionBalanceCommon);
 
-  let message: string | null = null;
+  const showPriceImpactWarning =
+    priceImpact?.kind === 'loss' && priceImpact.level === 'medium';
```
zerts, line 165: Should we also check `priceImpact.level === 'high'` ?

vyorkin, line 165:
```suggestion
    priceImpact?.kind === 'loss' && (priceImpact.level === 'medium' || priceImpact.level === 'high');
```

vyorkin: thanks, I don't know what I was thinking ✅

src/shared/units/formatPercent/formatPercentChange.ts
everdimension, line 1: I don't see the point in this extraction
The helper fields that are created with this function are still only used in `Overview` and nowhere else.
In other components using plain `formatPercent` would be enough

src/ui/pages/SwapForm/shared/PriceImpactLine/PriceImpactLine.tsx
```diff
@@ -0,0 +1,36 @@
...
+    priceImpact.kind === 'loss' && priceImpact.level === 'high';
+
+  const priceImpactPercentage = priceImpact
+    ? getPriceImpactPercentage(priceImpact)
+    : null;
+
+  const percentageChange = useMemo(
+    () =>
+      priceImpactPercentage
+        ? formatPercentChange(priceImpactPercentage, 'en')
```
everdimension, line 19: `formatPercent`?

src/shared/units/formatPercent/formatPercentChange.ts
vyorkin: right, it seems that we don't need other fields anywhere else, updated ✅

src/ui/pages/SwapForm/shared/PriceImpactLine/PriceImpactLine.tsx
vyorkin: ✅

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -272,6 +293,25 @@ export function SwapFormComponent() {
...
+  const { receiveAsset, spendAsset } = swapView;
+
+  const priceImpact = useMemo(() => {
+    const inputValue =
+      (primaryInput === 'spend' ? spendInput : receiveInput) || null;
+    const outputValue =
+      (primaryInput === 'spend' ? receiveInput : spendInput) || null;
+
+    const inputAsset = primaryInput === 'spend' ? spendAsset : receiveAsset;
+    const outputAsset = primaryInput === 'spend' ? receiveAsset : spendAsset;
```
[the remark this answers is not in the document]
vyorkin: yes, makes sense, updated ✅

PR #741, opened 2025-03-21
Update extension name

zerts:


zerts:
Not the most optimal solution. Will go with GH

PR #742, opened 2025-03-22
Playground page that shows previews of components

everdimension:


src/ui-lab/previews/registerPreview.ts
```diff
@@ -0,0 +1,31 @@
...
+export const previews: PreviewConfig[] = [];
+
+/**
+ * Call this function near a component to place a preview
+ * of it on the `/playground` route.
+ * NOTE: Intended to be used during development and
+ * should not be commited to final code. In order to create permanent previews
+ * use the `registerPreviewPermanent` function
+ */
+export function registerPreview(
```
zerts, line 17: Why do we need this one?

src/ui/pages/Settings/Settings.tsx
```diff
@@ -149,6 +153,16 @@ function SettingsMain() {
                 </HStack>
               </AngleRightRow>
             </FrameListItemLink>
+            {hasTestWallet ? (
```
zerts, line 156: This is minor, but having test wallet as a watched address also allows this item to appear. Do we want to avoid this?

PR #743, opened 2025-03-25
QA Flow Github

zerts:


vyorkin (approved):

.github/workflows/pr.yml
```diff
@@ -75,6 +75,43 @@ jobs:
           npm install
           npm run build:production
 
+      - name: Update Dev Manifest file
```
everdimension, line 78: Can we have a separate workflow for this?

We'd need to find a way to import the `env` vars from this one to avoid duplication

package.json
```diff
@@ -15,7 +15,8 @@
     "lint": "eslint . --cache --fix --ext .ts,.tsx,.js",
     "format": "prettier --loglevel warn --write \"**/*.{ts,tsx,css,md}\"",
     "prepare": "husky install",
-    "postversion": "node ./scripts/manifest-version.js"
+    "postversion": "node ./scripts/manifest-version.js",
+    "update-dev-manifest": "node ./scripts/update-dev-manifest.js"
```
everdimension, line 19: i think we can avoid putting this in `package.json`. We can call the script directly from the workflow, right?

.github/workflows/pr.yml
```diff
@@ -75,6 +75,43 @@ jobs:
...
+        if: ${{ success() }}
+        env:
+          GIT_BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+        run: npm run update-dev-manifest
+
+      - name: Create branch in another repository and copy files
+        if: ${{ success() }}
+        env:
+          REPO_TOKEN: ${{ secrets.QA_REPOSITORY_PAT }}
```
everdimension, line 87: `_PATH`?

scripts/update-dev-manifest.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-env node */
+const fs = require('fs/promises');
+const path = require('path');
+const manifest = require('../dist/manifest.json');
+
+const GIT_BRANCH_NAME = process.env.GIT_BRANCH_NAME;
+
+const dist = path.join(__dirname, '../dist');
+
+async function updateDevName() {
```
everdimension, line 10: maybe make path to manifest.json a param?

.github/workflows/pr.yml
zerts: This is Personal Access Token :)

PR #746, opened 2025-04-06
Update QA Icon

zerts:


.github/workflows/pr_push_qa.yml
```diff
@@ -56,6 +56,13 @@ jobs:
         with:
           node-version: '20.x'
 
+      - name: Replace icon with QA icon
```
zerts, line 59: This approach will keep the main code and disabled state logic untouched.

vyorkin (approved):

PR #750, opened 2025-04-09
Support send to another address in bridge form

vyorkin:


src/ui/pages/BridgeForm/BridgeForm.tsx
```diff
@@ -420,7 +432,8 @@ function BridgeFormComponent() {
   ]);
 
   const quotesData = useQuotes({
-    address,
+    from: address,
+    to,
```
zerts, line 436: Didn't check it yet, but looks like we need to make sure that `showReceiverAddressInput: false` to pass to param here.

vyorkin: Thanks! I think it makes sense, actually, that would be quite a critical bug... updated ✅

zerts (approved):

src/ui/ui-kit/FormFieldset/FormFieldset.tsx
```diff
@@ -66,12 +66,16 @@ export const FormFieldset = React.forwardRef<
...
-            <UIText kind="small/regular" as="label" ref={labelRef}>
-              {title}
-            </UIText>
-            {endTitle ? <UIText kind="small/regular">{endTitle}</UIText> : null}
-          </HStack>
+          {title || endTitle ? (
+            <HStack gap={8} justifyContent="space-between" alignItems="center">
+              <UIText kind="small/regular" as="label" ref={labelRef}>
+                {title}
+              </UIText>
```
everdimension, line 73: in case of falsy `title` well have empty node here, I assume this is on purpose, right? So that the `endTitle` stays on end?

src/ui/pages/BridgeForm/BridgeForm.tsx
```diff
@@ -356,15 +360,23 @@ function BridgeFormComponent() {
   const formState = useMemo(
     () => ({
       ...defaultFormValues,
+      to: null,
```
everdimension, line 363: why is this needed if it's null anyway?

vyorkin: The type of the field is `to: string | null`.
Because the `to` field can not be "undefined" – we either want to send to another address or not, there is no 3rd option

src/ui/ui-kit/FormFieldset/FormFieldset.tsx
vyorkin: yes that was the intention

src/ui/pages/BridgeForm/BridgeForm.tsx
everdimension: Got it! It's `string | undefined` though in SendForm

PR #755, opened 2025-04-16
Fix onboarding open from sidepanel

zerts:


vyorkin (approved):

src/ui/features/onboarding/initialization.ts
```diff
@@ -1,34 +1,17 @@
...
+  if (!isOnboarding && !hasExistingUser) {
     openOnboarding();
+    await chrome.sidePanel.setPanelBehavior({
+      openPanelOnActionClick: false,
+    });
+    window.close();
     throw new OnboardingInterrupt();
   }
-
-  const appMode = getAppMode({ hasExistingUser });
```
zerts, line 25: Need to check why this was done and is it safe to remove this?

zerts: That was added during the task of making `esc` close dialogs. However, everything with dialog I was able to find, still works

everdimension (approved):

PR #757, opened 2025-04-16
Remove ExploreZeroBanner

vyorkin:


src/background/Wallet/model/types.ts
```diff
@@ -48,6 +48,7 @@ interface PublicPreferences {
   mintDnaBannerDismissed?: boolean;
   upgradeDnaBannerDismissed?: boolean;
   inviteFriendsBannerDismissed?: boolean;
+  /** @deprecated */
```
zerts, line 51: I think, we can remove this from default value for WalletRecord then :)

everdimension (approved):

PR #759, opened 2025-04-17
Premium Avatar Border

zerts:


src/ui/shared/requests/useWalletsMetaByChunks.ts
```diff
@@ -1,22 +1,48 @@
...
     suspense,
     useErrorBoundary,
+    staleTime,
   });
+
+  // Update query cache with wallet meta data for each address
+  // to make it prefetched for single address requests (e.g. wallet avatar)
+  // useLayoutEffect is used here to add data before UI is updated and isLoading flag from the query unlocks the render
+  // so new components will not request the data again
+  useLayoutEffect(() => {
```
everdimension, line 33: can't we do this inside `queryFn` above instead?

src/ui/shared/requests/useWalletsMetaByChunks.ts
```diff
@@ -1,22 +1,39 @@
...
     enabled: enabled && addresses.length > 0,
     queryKey: ['ZerionAPI.getWalletsMetaByChunks', addresses],
-    queryFn: () => ZerionAPI.getWalletsMetaByChunks(addresses),
+    queryFn: async () => {
+      const result = await ZerionAPI.getWalletsMetaByChunks(addresses);
+      if (result.length > 1) {
+        result.forEach((walletMeta) => {
+          const normalizedAddress = normalizeAddress(walletMeta.address);
+          queryClient.setQueryData(
+            ['ZerionAPI.getWalletsMetaByChunks', [normalizedAddress]],
```
everdimension, line 28: Why is "by chunks" if it's a single address?
I think it's more appropriate to set this as `['walletGetMeta', normalizedAddress]`

everdimension: Let's research `initialData` approach? https://tkdodo.eu/blog/seeding-the-query-cache#seeding-details-from-lists

zerts: Because the logic is to avoid single address fetches if we've already fetched a group
I will rename the fetcher

src/ui/components/WalletAvatar/WalletAvatar.tsx
```diff
@@ -18,7 +19,13 @@ export function WalletAvatar({
   icon?: React.ReactNode;
   onReady?(): void;
 }) {
-  const { data: nft, isLoading } = useProfileNft(address);
+  const { data, isLoading } = useWalletsMetaByChunks({
```
everdimension, line 22: Since this is a request for a single address, `useWalletMeta` should be used

See https://github.com/zeriontech/zerion-wallet-extension/pull/759/files#r2177227801

src/ui/ui-kit/MediaContent/MediaContent.tsx
```diff
@@ -31,14 +32,26 @@ export const MediaError = ({
...
+  content: WalletMetaMediaContent
+): MediaContentValue {
+  return {
+    image_preview_url: content.imagePreviewUrl,
+    image_url: content.imageUrl,
+    audio_url: content.audioUrl,
+    video_url: content.videoUrl,
+    type: content.type,
+  };
+}
```
everdimension, line 53: 👍

PR #760, opened 2025-04-17
Premium Status support

zerts:


src/ui/features/premium/getPremiumStatus.ts
```diff
@@ -0,0 +1,59 @@
...
+      meta.membership.premium?.plan !== 'Restricted'
+    ) {
+      return true;
+    }
+    if (
+      normalizedAddress &&
+      Boolean(meta.membership.premium) &&
+      normalizeAddress(meta.address) === normalizedAddress
+    ) {
+      return true;
```
everdimension, line 26: Not sure I get the logic
If any other address being premium is enough, why pass current address at all?
But also if address is current then "restricted" value doesn't matter?

If the described logic is correct then I think it requires a comment

zerts: Done!
I hope this will help a bit :)

src/ui/features/premium/getPremiumStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+export function usePremiumStatus({ address }: { address?: string }) {
+  const normalizedAddress = address ? normalizeAddress(address) : undefined;
+  const addresses = useAllSignerOrHwAddresses();
+
+  const walletsMetaQuery = useWalletsMetaByChunks({
+    addresses: addresses || [],
+    enabled: Boolean(addresses?.length),
+    suspense: false,
+    useErrorBoundary: false,
+  });
```
everdimension, line 51: Should we give this query a long `staleTime`?

zerts: Probably we can

everdimension (approved):
to fix: https://github.com/zeriontech/zerion-wallet-extension/pull/760/files#r2178059228

PR #763, opened 2025-04-21
Add extra fields to analytics for bridge and send

vyorkin:


src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -139,9 +143,9 @@ function FormHint({
     : null;
   const exceedsBalance = Number(spendInput) > Number(positionBalanceCommon);
 
-  const showPriceImpactWarning =
-    priceImpact?.kind === 'loss' &&
-    (priceImpact.level === 'medium' || priceImpact.level === 'high');
+  const priceImpactWarningVisible = priceImpact
```
zerts, line 146: Is it `Boolean(priceImpact && isSignificantValueLoss(priceImpact))`?

vyorkin: ah, yes :)

src/ui/pages/SwapForm/SwapForm.tsx
```diff
@@ -534,6 +550,8 @@ export function SwapFormComponent() {
           chain,
         }),
         quote: selectedQuote,
+        warningWasShown: priceImpactWarningVisible,
+        outputAmountColor: priceImpactWarningVisible ? 'red' : 'grey',
```
zerts, line 554: There is a case where we show red text but doesn't show warning. 
Looks like this should be taken into account here

src/ui/components/FiatInputValue/FiatInputValue.tsx
```diff
@@ -88,9 +91,9 @@ export function ReceiveFiatInputValue({
 }: {
   priceImpact: PriceImpact | null;
 } & FieldInputValueProps) {
-  const isSignificantLoss =
-    priceImpact?.kind === 'loss' &&
-    (priceImpact.level === 'medium' || priceImpact.level === 'high');
+  const isSignificantLoss = priceImpact
```
zerts, line 94: https://github.com/zeriontech/zerion-wallet-extension/pull/763/files#r2053970462

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: hm I think you're right

src/shared/analytics/shared/transactionContextToAnalytics.ts
```diff
@@ -106,7 +106,10 @@ export function addressActionToAnalytics({
     asset_name_received: toMaybeArr(incoming?.map(getAssetName)),
     asset_address_sent: toMaybeArr(outgoing?.map(getAssetAddress)),
     asset_address_received: toMaybeArr(incoming?.map(getAssetAddress)),
+    warning_was_shown: warningWasShown,
```
zerts, line 109: Do you send these fields for Approve transactions?
If I remember right, we still need send them, but with the default values

src/ui/pages/SwapForm/SwapForm.tsx
zerts: Also, there is an unfortunate detail. Because we should reflect UI state in the analytics, it looks like we need calculate warning text here, on this level, and pass it as a param to the component. Otherwise, we will get inconsistency one day :(

src/ui/pages/BridgeForm/BridgeForm.tsx
```diff
@@ -709,6 +734,8 @@ function BridgeFormComponent() {
           chain: spendChain,
         }),
         quote: selectedQuote,
+        warningWasShown: isHighPriceImpact,
```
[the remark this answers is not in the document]
vyorkin: updated ✅

src/ui/components/FiatInputValue/FiatInputValue.tsx
vyorkin: updated ✅

src/ui/pages/SwapForm/SwapForm.tsx
vyorkin: there will be some "unavoidable" duplication (every duplication is avoidable at some cost :)) in `SwapForm` and BridgeForm` `"anyway", so I'm not sure if it is worth DRY'ing `outputAmountColor`. I mean the best thing I could do is to extract a function to calculate that `outputAmountColor` which is going to be called in `SwapFrom` and `BridgeForm`... but I'm not even sure if this extraction makes sense

vyorkin: but I fixed the condition behind `outputAmountColor` ✅

zerts (approved):

PR #775, opened 2025-05-07
Support for eip-7702 transactions

everdimension:
Update ethers libraries
Handle authorizationList tx property

### DIFF
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -3,6 +3,7 @@ import { Provider as ZksProvider } from 'zksync-ethers';
 import type { Emitter } from 'nanoevents';
 import { createNanoEvents } from 'nanoevents';
 import { nanoid } from 'nanoid';
+import omit from 'lodash/omit';
 import { isTruthy } from 'is-truthy-ts';
 import type {
   NotificationWindow,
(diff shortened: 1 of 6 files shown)


src/background/Wallet/Wallet.ts
```diff
@@ -1203,7 +1204,10 @@ export class Wallet {
           context,
           params: { typedData, ...transactionContextParams },
         });
-        const rawTransaction = serializePaymasterTx({ transaction, signature });
+        const rawTransaction = serializePaymasterTx({
```
zerts, line 1207: Shouldn't we create `typedData` from transaction without `authorizationList` too 5 lines above?

src/modules/ethereum/types/IncomingTransaction.ts
```diff
@@ -8,10 +8,7 @@ type AddressStr = string;
  * Represents a tx object that can be received from
  * a dapp, an API response or created manually
  */
-export type IncomingTransaction = Pick<
-  TransactionRequest,
-  'nonce' | 'data' | 'customData' | 'accessList'
-> & {
+export type IncomingTransaction = TransactionRequest & {
```
zerts, line 11: 👍

zerts (approved):

src/background/Wallet/Wallet.ts
everdimension: It shouldn't be there anyway, but good catch! It's more proper to do this above

PR #781, opened 2025-05-13
Chart Data Points

zerts:


src/ui/components/chart/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { ChartConfiguration, Plugin } from 'chart.js/auto';
+
+export type ChartPoint<T = unknown> = [number, number, T];
+export type ParsedChartPoint<T = unknown> =
```
everdimension, line 4: do we need the fallback to unknown?

src/ui/style/theme.module.css
```diff
@@ -63,6 +63,8 @@
     0px 8px 28px rgba(22, 22, 26, 0.08);
   --elevation-300: 0px 24px 48px -16px rgba(21, 21, 31, 0.24),
     0px 4px 8px -2px rgba(21, 21, 31, 0.08);
+  --background-transparent: rgba(22, 22, 26, 0.9);
```
everdimension, line 66: Is this specified in figma's ui-kit?

src/ui/pages/AssetInfo/AssetChart/tooltip.ts
```diff
@@ -0,0 +1,199 @@
+import { type TooltipOptions, type Chart } from 'chart.js';
+import type { AssetChartActions } from 'src/modules/zerion-api/requests/asset-get-chart';
+import { createNode as r } from 'src/content-script/in-dapp-notifications/createNode';
```
everdimension, line 3: 🔥

src/ui/components/chart/Chart.tsx
```diff
@@ -142,12 +146,18 @@ function getSegmentColor({
...
 
-export function Chart({
+export function Chart<T>({
   chartPoints,
   onRangeSelect,
   style,
+  datasetConfig = {},
+  tooltip = {},
+  plugins = [],
+  interaction = {},
```
everdimension, line 156: I actually do not know whether using default values in useEffect dependencies will cause updates 🤔 Can you check this please?

src/ui/pages/AssetInfo/AssetChart/interaction.ts
```diff
@@ -0,0 +1,77 @@
...
+    axis: 'x',
+    intersect: false,
+  });
+  const activePoint = indexPoints[0];
+
+  if (!activePoint) {
+    return [];
+  }
+
+  const marneticRadius = 4;
```
everdimension, line 32: magnetic

src/ui/components/chart/types.ts
zerts: In some cases we can just work with ChartPoint type in some helpers where we don't need information about an extra field. So I decided that is it possible to use this type without mandatory generic type

PR #783, opened 2025-05-16
Handle dapp network state for Solana

everdimension:
Refactor requests for native balance to RPC Nodes


src/shared/PortMessageChannel.ts
```diff
@@ -43,6 +44,12 @@ export class PortMessageChannel {
     });
     this.emitter.on('message', (response) => {
       if (isJsonRpcPayload(response)) {
+        if (isJsonRpcError(response)) {
+          console.group('Port method error'); // eslint-disable-line no-console
```
zerts, line 48: Do we need this in production?

everdimension: Yes, it's similar to what I did here: https://github.com/zeriontech/zerion-wallet-extension/blob/cd1303408519e38fbcc6b60b72e92f21c58752e8/src/background/messaging/mapRPCMessageToController.ts#L60-L62

It's useful both for debugging during dev and to investigate user reports. Otherwise there used to be no way to understand which requests between the UI layer and background caused errors

src/background/Wallet/Wallet.ts
```diff
@@ -1017,14 +1018,17 @@ export class Wallet {
   }
 
   async switchChainForOrigin({
-    params: { chain, origin },
+    params: { evmChain, solanaChain, origin },
     context,
-  }: WalletMethodParams<{ chain: string; origin: string }>) {
+  }: WalletMethodParams<
+    AtLeastOneOf<{ evmChain: string; solanaChain: string }> & { origin: string }
```
zerts, line 1024: Aren't this the same type as on the line 1031? 
`Parameters<Wallet['setChainForOrigin']>[0])`

zerts (approved):

src/background/Wallet/Wallet.ts
everdimension: ts doesn't recognize that the `either-or` constraint is kept

src/ui/pages/Overview/ConnectionHeader/ConnectionHeader.tsx
```diff
@@ -119,22 +124,36 @@ export function ConnectionHeader() {
...
-  );
+  const { singleAddressNormalized: address } = useAddressParams();
+  const { data: isConnected } = useIsConnectedToActiveTab(address);
 
   const { data: siteChain, ...chainQuery } = useQuery({
-    queryKey: ['requestChainForOrigin', activeTabOrigin],
-    queryFn: () => requestChainForOrigin(activeTabOrigin),
+    queryKey: ['requestChainForOrigin', activeTabOrigin, address],
+    queryFn: async () =>
+      activeTabOrigin
```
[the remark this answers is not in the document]
everdimension: Good idea!

PR #784, opened 2025-05-16
Track app opened events

vyorkin:


src/ui/features/onboarding/Onboarding.tsx
```diff
@@ -13,7 +12,6 @@ import { Backup } from './Backup';
 
 export function Onboarding() {
   const navigate = useNavigate();
-  useScreenViewChange();
```
everdimension, line 16: 👍

src/ui/shared/useScreenViewChange.ts
```diff
@@ -1,36 +1,59 @@
...
+        event: 'appOpened',
         pathname,
-        address,
-        previous: previousPathname.current,
-        screenSize: `${window.screen.width}x${window.screen.height}`,
-        windowType: urlContext.windowType,
+        previousPathname: previousPathname.current,
+      }).then(() => {
+        previousPathname.current = pathname;
       });
```
everdimension, line 38: I think that maybe it doesn't make sense to await the call to update prev path name?
Wdyt?

```
trackScreenView({ ..., previousPathname: previousPathname.current });
previousPathname.current = pathname;
```

vyorkin: ah, yess, actually makes sense, no need to wait to update it

everdimension (approved):

src/shared/analytics/analytics.background.ts
```diff
@@ -86,24 +114,20 @@ function trackAppEvents({ account }: { account: Account }) {
...
-    const portfolio = await getOwnedWalletsPortolio(account);
-    const mixpanelParams: Record<string, unknown> = {
-      ...omit(params, ['request_name', 'wallet_address']),
-      total_balance: portfolio?.total_value ?? 0,
-      ...getChainBreakdown(portfolio),
-    };
-    mixpanelTrack(account, 'General: Screen Viewed', mixpanelParams);
-  });
+  emitter.on('screenView', (data) =>
+    trackScreenView({
```
zerts, line 118: There was a request from analytics:
`Please delete property [blockchain_id](id from backend)_balance from other events, except the previous one`
So we should remove chain balances from screen view events, and keep them only in app opened event

vyorkin: ah, thanks, nice catch, formal-natural language parsing issue

zerts:
https://github.com/zeriontech/zerion-wallet-extension/pull/823

PR #807, opened 2025-06-27
Fix address input select by name

zerts:


src/ui/pages/SendForm/fieldsets/AddressInput/AddressInput.tsx
```diff
@@ -51,6 +52,7 @@ function matches(query: string | null, item: Item, domainInfo?: string[]) {
   }
   const value = query.toLowerCase();
   return (
+    normalizedContains(normalizeAddress(item.address), query) || // Solana address is not transformed to lowercase
     normalizedContains(normalizeAddress(item.address), value) ||
```
everdimension, line 56: Should it be this instead?
```suggestion
    normalizedContains(normalizeAddress(item.address).toLowerCase(), value) ||
```

zerts: Yes, this works!

PR #812, opened 2025-07-03
Combine analytics userId with landing and web app

zerts:
The flow described in `src/shared/analytics/analyticsId.ts`

We will try to sync userId with the userId from web-app through the iframe method.
As a fallback, deviceId will be used

### DIFF
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -1719,7 +1719,7 @@ export class Wallet {
     this.verifyInternalOrigin(context);
     /**
      * Wait for analyticsId to be set before sending screenView event.
-     * This can happen on the first launch or suring the migration to thenew analytics ID storage
+     * This can happen on the first launch or during the migration to the new analytics ID
      */
     const analyticsId = await BrowserStorage.get<string>(analyticsIdKey);
     if (!analyticsId) {
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -109,12 +109,15 @@ import { isMatchForEcosystem } from 'src/shared/wallet/shared';
 import type { AtLeastOneOf } from 'src/shared/type-utils/OneOf';
 import type { StringBase64 } from 'src/shared/types/StringBase64';
 import { createApprovalTransaction } from 'src/modules/ethereum/transactions/appovals';
+import { wait } from 'src/shared/wait';
+import { analyticsIdKey } from 'src/shared/analytics/analyticsId';
 import type { DaylightEventParams, ScreenViewParams } from '../events';
 import { emitter } from '../events';
 import type { Credentials, SessionCredentials } from '../account/Credentials';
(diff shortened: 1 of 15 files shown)


src/shared/analytics/analyticsId.client.ts
```diff
@@ -0,0 +1,8 @@
+import { BrowserStorage } from 'src/background/webapis/storage';
+import { walletPort } from 'src/ui/shared/channels';
+import { analyticsIdKey } from './analyticsId';
+
+export async function setAnalyticsIdIfNeeded(id: string) {
```
everdimension, line 5: what does `ifNeeded` imply?

src/shared/analytics/mixpanel.ts
```diff
@@ -175,17 +163,16 @@ class MixpanelApi {
...
-  async identify(
-    userId: string,
-    userProfileProperties: Record<string, unknown>
-  ) {
+  async identify(userProfileProperties: Record<string, unknown>) {
     await this.ready();
-    this.userId = userId;
-    const $anon_distinct_id = `$device:${this.deviceId}`;
+    const userId = await getAnalyticsId();
+    const $anon_distinct_id = await deviceIdStore.getSavedState();
```
everdimension, line 169: why not `this.deviceId`?

src/background/Wallet/Wallet.ts
```diff
@@ -1705,10 +1708,30 @@ export class Wallet {
...
+    if (!analyticsId) {
+      await Promise.race([
+        wait(5000),
+        new Promise<void>((resolve) => {
+          emitter.on('analyticsIdSet', () => {
+            resolve();
+          });
+        }),
+      ]);
+    }
```
everdimension, line 1734: Do you think this should be inside `getAnalyticsId` instead?

src/background/Wallet/Wallet.ts
```diff
@@ -1705,10 +1708,30 @@ export class Wallet {
...
+    /**
+     * Wait for analyticsId to be set before sending screenView event.
+     * This can happen on the first launch or during the migration to the new analytics ID
+     */
+    const analyticsId = await BrowserStorage.get<string>(analyticsIdKey);
+    if (!analyticsId) {
+      await Promise.race([
+        wait(5000),
+        new Promise<void>((resolve) => {
+          emitter.on('analyticsIdSet', () => {
```
everdimension, line 1729: `browser.storage.onChanged.addListener` can be used instead

src/shared/analytics/analyticsId.client.ts
zerts: Thank you! This was left from the initial implementation

src/shared/analytics/mixpanel.ts
zerts: Thank you! I missed this one

src/background/Wallet/Wallet.ts
zerts: Thank you! This will be helpful!

PR #815, opened 2025-07-15
Configurable "data" in Send Transactions

everdimension:


src/ui/pages/SendForm/SendForm.tsx
```diff
@@ -433,6 +435,40 @@ function SendFormComponent() {
                 networkId={formState.tokenChain || null}
               />
             ) : null}
+            {preferences?.configurableTransactionData &&
+            addressType === 'evm' ? (
+              <FormFieldset
+                title="Data"
+                style={{ borderBottomRightRadius: 4 }}
+                startInput={
+                  <textarea
```
zerts, line 444: Do we need an UnstyledTextarea element?

zerts (approved):

src/ui/pages/SendForm/SendForm.tsx
everdimension: Probably! Once it's used somewhere else perhaps....

zerts (approved):

PR #816, opened 2025-07-16
UK Disclaimer

zerts:


src/ui/components/UKDisclaimer/useUKDetection.ts
```diff
@@ -0,0 +1,6 @@
+import { useStatsigFeatureGate } from 'src/modules/statsig/statsig.client';
+
+export function useUKDetection() {
+  const { data: isUK, isLoading } = useStatsigFeatureGate('crypto_disclaimer');
```
everdimension, line 4: isn't `data` an object `{ gateName: string; value: boolean }`?  https://github.com/zeriontech/zerion-wallet-extension/pull/816/files#diff-000da9f1e4061b37f11d97233e45c6fe7b1bd38d944d541cd6cf8249cf3fce75R71

PR #824, opened 2025-07-24
Fix double transaction interpretation request

zerts:


src/ui/pages/SendTransaction/SendTransaction.tsx
```diff
@@ -616,7 +616,7 @@ function SendTransactionContent({
   const client = useDefiSdkClient();
 
   const interpretQuery = useQuery({
-    enabled: Boolean(network?.supports_simulations),
+    enabled: Boolean(network?.supports_simulations && chainGasPrices),
```
everdimension, line 619: to be discussed

PR #825, opened 2025-07-24
Remove double quote requests in bridge

zerts:
Based on the request from backend. 
+ support loading view for quote list. Not perfect, but not broken


src/ui/pages/BridgeForm/QuoteList/QuoteList.tsx
```diff
@@ -206,6 +209,11 @@ export function QuoteList({
                   </label>
                 );
               })}
+              {isLoading ? (
+                <div style={{ height: 32 }}>
```
zerts, line 213: Added fixed height to avoid scroll blinking (when rotating, spinner slightly changes height which impacts `fit-content` dialog)

PR #828, opened 2025-07-28
Show High Price impact warning + native bridge hint

zerts:
In this pull requests there are few additions:
- Use data from quote to set protocol_fee, network_fee and gas_price params in the sign_transaction event
- Add warning_was_shown and output_text_color params to reflect price impact warning state
- Add price impact warning callout and extract all logic for detect what price impact warnings should we show to the top level of the form so the same calculation results are passed to analytics.

### DIFF
--- a/src/shared/analytics/shared/addressActionToAnalytics.ts
+++ b/src/shared/analytics/shared/addressActionToAnalytics.ts
@@ -121,11 +121,11 @@ export function addressActionToAnalytics({
     const zerion_fee_usd_amount =
       quote.protocolFee.amount.usdValue ?? undefined;
     const network_fee = quote.networkFee?.amount?.usdValue ?? undefined;
-    const currentTransaction =
+    const currentTransactionEvm =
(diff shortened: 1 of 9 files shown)


src/shared/analytics/shared/addressActionToAnalytics.ts
```diff
@@ -116,17 +118,22 @@ export function addressActionToAnalytics({
...
   if (quote) {
     const zerion_fee_percentage = quote.protocolFee.percentage;
-    const feeAmount = quote.protocolFee.amount.quantity;
-    const asset = incoming?.[0]?.asset;
     const zerion_fee_usd_amount =
-      feeAmount && asset
-        ? assetQuantityToValue({ quantity: feeAmount, asset }, chain)
+      quote.protocolFee.amount.usdValue ?? undefined;
+    const network_fee = quote.networkFee?.amount?.usdValue ?? undefined;
+    const currentTransaction =
```
everdimension, line 124: maybe
```suggestion
    const currentTransactionEvm =
```

PR #829, opened 2025-07-28
Track tx sign success

zerts:
This pull request is based on another one https://github.com/zeriontech/zerion-wallet-extension/pull/828
Cause all changes are about the same event. Please, take a look into the previous one first

Updated sign transaction event, so now it can be called in case of `transactionSent` or `transactionFailed` events.
Transaction failed event fired from the Wallet.ts and passes error message.
Todo for the future: support failed txs for Solana. Right now we don't parse tx errors for Solana


src/background/Wallet/Wallet.ts
```diff
@@ -1297,7 +1298,11 @@ export class Wallet {
         );
         return safeTxPlain;
       } catch (error) {
-        throw getEthersError(error);
+        const parsedError = getEthersError(error);
+        const parsed = parseError(parsedError);
```
everdimension, line 1302: uhm :)

improve naming

zerts: Ah, yes, good comment!
Will need to find a better name

everdimension: can simply do:

```
        const parsed = parseError(getEthersError(error));
```

everdimension: ah, i see you use it later, too

src/shared/analytics/analytics.background.ts
```diff
@@ -153,69 +155,99 @@ function trackAppEvents({ account }: { account: Account }) {
...
-      const initiatorName = isInternalOrigin ? 'Extension' : 'External Dapp';
-      const addressActionAnalytics = addressActionToAnalytics({
-        addressAction,
-        quote,
-        outputChain: outputChain ?? null,
-      });
-      const preferences = await account
-        .getCurrentWallet()
-        .getPreferences({ context: INTERNAL_SYMBOL_CONTEXT });
+  const trackTransactionSent = async (
```
everdimension, line 158: maybe `trackTransactionEvent`? "sent" implies success, I think

everdimension (approved):
https://github.com/zeriontech/zerion-wallet-extension/pull/829/files?diff=unified&w=1#r2245054581 and other comments

PR #830, opened 2025-07-28
Filled form event

zerts:
This pull request is based on another one https://github.com/zeriontech/zerion-wallet-extension/pull/829
Cause all changes are about the same event. Please, take a look into the previous one first

New event should be sent when there is a new ready quote in the Swap/Bridge form.

### DIFF
--- a/src/shared/analytics/analytics.background.ts
+++ b/src/shared/analytics/analytics.background.ts
@@ -11,6 +11,10 @@ import {
   getTxSender,
 } from 'src/modules/shared/transactions/helpers';
 import { statsigTrack } from 'src/modules/statsig/shared';
+import { getGas } from 'src/modules/ethereum/transactions/getGas';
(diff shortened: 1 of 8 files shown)


src/shared/analytics/analytics.background.ts
```diff
@@ -250,6 +271,90 @@ function trackAppEvents({ account }: { account: Account }) {
...
+
+    const slippageOptions = formState.inputChain
+      ? getSlippageOptions({
+          chain: createChain(formState.inputChain),
+          userSlippage:
+            'slippage' in formState && formState.slippage
+              ? Number(formState.slippage)
+              : null,
+        })
+      : undefined;
```
everdimension, line 311: These calculations really do not belong here : /
I think this value should be attached to the event

zerts: Agree. I was thinking about this too. Good to get the second option!

PR #831, opened 2025-07-29
Chart improvements

zerts:
Updates in this pull request
- Added ability to show/hide asset chart dots via preferences
- Updated style for multi action dots - now it is a custom icon
- Added animated right point
- Populated backend data with chart dot direction (in/out) in case it is not present. (For now it was decided to solved on the client side)

### DIFF
--- a/src/background/Wallet/WalletRecord.ts
+++ b/src/background/Wallet/WalletRecord.ts
@@ -857,7 +857,7 @@ export class WalletRecordModel {
       enableTestnets: false,
       testnetMode: null,
       enableHoldToSignButton: null,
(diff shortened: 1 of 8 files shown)


src/background/Wallet/WalletRecord.ts
```diff
@@ -857,6 +857,7 @@ export class WalletRecordModel {
       enableTestnets: false,
       testnetMode: null,
       enableHoldToSignButton: null,
+      transactionsOnAssetPageChartHidden: false,
```
everdimension, line 860: sounds weird and backwards

More commonly this is names something like `showTrades` or `showTradesOnChart` with a default of `true`

PR #833, opened 2025-08-04
Reduce interpret requests

everdimension:
Prefer to wait for gas-prices response before
making the interpret request


zerts (approved):
May be a comment would be helpful :)

PR #834, opened 2025-08-05
Search page

zerts:
+ Added Search page based on Combobox UX
+ Added Recent Searches as an array of fungibleIds in preferences (same way as the recentAddresses in Send Form)


src/background/Wallet/model/types.ts
```diff
@@ -45,6 +45,7 @@ interface PublicPreferences {
   configurableNonce?: boolean;
   invitationBannerDismissed?: boolean;
   recentAddresses?: string[];
+  recentSearch?: string[];
```
everdimension, line 48: This shouldn't be part of wallet record; instead it should be a separate storage entry, like `chain-configs`

src/ui/pages/Overview/Positions/Positions.tsx
```diff
@@ -545,6 +548,13 @@ function PositionList({
               component: assetPageEnabled ? (
                 <SurfaceItemLink
                   to={`/asset/${position.asset.id}`}
+                  onClick={() => {
+                    walletPort.request('assetClicked', {
+                      assetId: position.asset.id,
+                      pathname,
+                      section: 'Overview',
+                    });
+                  }}
```
everdimension, line 557: What if asset link is pressed using the keyboard?

Ideally "screen view" event should be enough for this. Maybe these links should have additional search params to let analytics know that they have been visited from search results.

everdimension (changes requested):
See comments

src/ui/pages/Overview/Positions/Positions.tsx
zerts: I believe this is an approach for analytics to track all usage of the search page.
However, the previous screen name really should be enough for this. I will ask about this.

everdimension (approved):

PR #848, opened 2025-09-09
Create client error event

zerts:
**General: Client Error** event was created for backend-related errors in trade forms.
In this pull request:
- onError handle was added to useEventSource to catch global errors 4**, 5**
- event.code -> error message mapper was extracted out of the useEventSource cause it was more quote related
- new analytics event for quote errors populated with information from the trading forms

Quote error can be detected in 2 different ways:
- global error for 4**/5** errors via onError callback
- quote errors via useEffect for the selectedQuote

### DIFF
--- a/src/shared/analytics/analytics.background.ts
+++ b/src/shared/analytics/analytics.background.ts
@@ -393,6 +393,10 @@ function trackAppEvents({ account }: { account: Account }) {
   });
 
   emitter.on('quoteError', async (quoteErrorContext, source) => {
+    const preferences = await globalPreferences.getPreferences();
+    if (!preferences.analyticsEnabled) {
+      return;
+    }
     const [inputTokenPositions, inputAssetData, outputAssetData] =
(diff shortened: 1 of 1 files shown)


src/ui/shared/requests/useQuotes.ts
```diff
@@ -148,6 +202,19 @@ export function useQuotes2({
           return updatedItem || item;
         });
       },
+      eventCodeToMessage: QUOTES_EVENT_CODE_TO_MESSAGE,
+      onError: ({ parsedError, rawEvent, requestUrl }) => {
+        handleQuoteError({
+          message: parsedError.message,
+          code: rawEvent.code,
+          backendMessage: rawEvent.message,
+          requestParams: Object.fromEntries(
```
zerts, line 211: Form state can change before onError call, so requestParams used to pass correct form state into the handleQuoteError callback.

PR #852, opened 2025-09-19
Update Deposit entry point and the Receive page

everdimension:


src/ui/pages/Receive/AddressDetails.tsx
```diff
@@ -1,13 +1,200 @@
...
+        </HStack>
+      ))}
+    </VStack>
+  );
+}
+
+function EthereumNetworksVisualStack() {
+  return (
+    <div style={{ display: 'flex' }}>
+      <img
```
zerts, line 48: do we need alt here?
and should we extract the same style as a separate var?

src/ui/pages/Overview/Positions/EmptyPositionsView.tsx
```diff
@@ -76,58 +175,53 @@ export function EmptyPositionsView() {
...
+            size={48}
+            kind="primary"
+            onClick={() => {
+              dialogRef.current?.showModal();
+            }}
           >
-            Transfer from Ethereum
+            <HStack gap={8} alignItems="center" justifyContent="center">
+              <ArrowCircleIcon />
+              <span>Fund</span>
```
zerts, line 215: Is this the final copy? Looks too short :))

src/ui/pages/Receive/AddressDetails.tsx
everdimension: Yes, thx! I was thinking about ways to abstract this and ended up skipping it :)

src/ui/pages/Overview/Positions/EmptyPositionsView.tsx
everdimension: Seems so!

PR #854, opened 2025-10-01
Switch to the internal rpc for dapp rpc requests

zerts:


zerts:
Should be fixed on the backend with the new public node endpoint

PR #952, opened 2026-07-04
WLT-1686: fix NaN for mixed-case hex Permit approval amount

zerts:
## Summary

An EIP-712 Permit `value` supplied as a **mixed-case** hex string (e.g. `0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF`) rendered as `NaN mETH` in the approval UI.

Root cause: `bignumber.js` parses all-uppercase (`0xFF`) or all-lowercase (`0xff`) hex, but returns `NaN` for mixed-case hex — `new BigNumber('0xFf')` → `NaN`. The raw Permit `value` flows into `baseToCommon` (`new BigNumber(value)`) in `SignTypedData.tsx` and into `new BigNumber(...)` in `AllowanceView`. The reported value is a valid uint256 (≈`2.13e19` base units, not max-uint256), so the fix is to parse it correctly and show the real amount.

## Changes

- Add exported `normalizeNumberValue` in `src/shared/units/convert.ts` — converts any `0x`-hex string to a decimal string via `BigInt` (case-insensitive) before handing it to bignumber.js; non-hex input passes through unchanged.
- Route `baseToCommon` / `commonToBase` through it, plus `AllowanceView`'s two direct `new BigNumber(...)` constructions.
- Unit tests in `convert.test.ts` for the reported value + regressions.

**Parse-only, display-only.** No change to signing behaviour — the signed payload keeps the dApp's original `message.value`. Verified that ethers v6 `TypedDataEncoder.hash` produces byte-identical hashes for the hex and decimal forms of the same uint256, so the signature is unaffected.

Closes WLT-1686

## Test plan

- `test:unit src/shared/units/convert.test.ts` — passes (hex parse + regressions).
- `type-check`, `lint`, `prettier` — clean.
- Manual: request `eth_signTypedData_v4` for a Permit with `value: "0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF"` → approval UI shows the real amount instead of `NaN`; "Edit allowance" screen shows the amount too.

### DIFF
--- a/src/shared/units/convert.ts
+++ b/src/shared/units/convert.ts
@@ -1,9 +1,28 @@
 import { BigNumber } from 'bignumber.js';
 
+const HEX_STRING_PATTERN = /^0x[0-9a-f]+$/i;
+
+/**
+ * bignumber.js parses all-uppercase or all-lowercase 0x-hex strings, but
+ * returns NaN for mixed-case hex (e.g. `new BigNumber('0xFf')` -> NaN).
+ * EIP-712 uint256 fields (e.g. a Permit `value`) can legitimately arrive as
+ * mixed-case hex, so normalize any 0x-prefixed hex string to a decimal string
+ * via BigInt (which parses hex case-insensitively) before handing it to
+ * bignumber.js.
+ */
+export function normalizeNumberValue(value: BigNumber.Value): BigNumber.Value {
+  if (typeof value === 'string' && HEX_STRING_PATTERN.test(value)) {
+    return BigInt(value).toString();
+  }
+  return value;
+}
+
 export function baseToCommon(value: BigNumber.Value, decimalPlaces: number) {
-  return new BigNumber(value).shiftedBy(0 - decimalPlaces);
+  return new BigNumber(normalizeNumberValue(value)).shiftedBy(
+    0 - decimalPlaces
+  );
 }
 
 export function commonToBase(value: BigNumber.Value, decimalPlaces: number) {
-  return new BigNumber(value).shiftedBy(decimalPlaces);
+  return new BigNumber(normalizeNumberValue(value)).shiftedBy(decimalPlaces);
 }
--- a/src/shared/units/convert.ts
+++ b/src/shared/units/convert.ts
@@ -1,6 +1,5 @@
 import { BigNumber } from 'bignumber.js';
-
-const HEX_STRING_PATTERN = /^0x[0-9a-f]+$/i;
+import { isHexString } from 'ethers';
 
 /**
  * bignumber.js parses all-uppercase or all-lowercase 0x-hex strings, but
@@ -11,7 +10,9 @@ const HEX_STRING_PATTERN = /^0x[0-9a-f]+$/i;
  * bignumber.js.
  */
 export function normalizeNumberValue(value: BigNumber.Value): BigNumber.Value {
-  if (typeof value === 'string' && HEX_STRING_PATTERN.test(value)) {
+  // isHexString also returns true for a bare '0x', which BigInt can't parse,
+  // so require at least one hex digit.
+  if (isHexString(value) && value.length > 2) {
     return BigInt(value).toString();
   }
   return value;
--- a/src/shared/units/convert.test.ts
+++ b/src/shared/units/convert.test.ts
@@ -1,4 +1,4 @@
-import { baseToCommon, commonToBase } from './convert';
+import { baseToCommon, commonToBase, normalizeNumberValue } from './convert';
 
 test('baseToCommon', () => {
   expect(baseToCommon('8408943752575239', 18).toString()).toBe(
@@ -7,6 +7,33 @@ test('baseToCommon', () => {
   expect(baseToCommon('33871609663570237017', 18).toString()).toBe(
     '33.871609663570237017'
   );
+});
+
+test('baseToCommon parses hex-encoded values', () => {
+  // 0xDE0B6B3A7640000 === 10 ** 18 (mixed-case hex, which bignumber.js
+  // mishandles on its own, returning NaN)
+  expect(baseToCommon('0xDE0b6B3a7640000', 18).toFixed()).toBe('1');
+  // The value from WLT-1686: bignumber.js returns NaN for this mixed-case hex
+  expect(baseToCommon('0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF', 18).isNaN()).toBe(
+    false
+  );
+  expect(baseToCommon('0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF', 18).toFixed()).toBe(
+    baseToCommon('21267647932558578408597187050162094079', 18).toFixed()
+  );
+});
+
+test('normalizeNumberValue', () => {
+  // mixed-case hex -> decimal string
+  expect(normalizeNumberValue('0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF')).toBe(
+    '21267647932558578408597187050162094079'
+  );
+  expect(normalizeNumberValue('0xff')).toBe('255');
+  // non-hex input is passed through unchanged
+  expect(normalizeNumberValue('33871609663570237017')).toBe(
+    '33871609663570237017'
+  );
+  expect(normalizeNumberValue('33.871')).toBe('33.871');
+  expect(normalizeNumberValue(42)).toBe(42);
 });
 
 test('commonToBase', () => {
--- a/src/ui/components/AllowanceView/AllowanceView.tsx
+++ b/src/ui/components/AllowanceView/AllowanceView.tsx
@@ -11,6 +11,7 @@ import { usePositionsRefetchInterval } from 'src/ui/transactions/usePositionsRef
 import type { NetworkConfig } from 'src/modules/networks/NetworkConfig';
 import { createChain } from 'src/modules/networks/Chain';
 import type { AnyAddressAction } from 'src/modules/ethereum/transactions/addressAction';
+import { normalizeNumberValue } from 'src/shared/units/convert';
 import { AllowanceForm } from '../AllowanceForm';
 import { NavigationBar } from '../NavigationBar';
 
@@ -82,9 +83,13 @@ export function AllowanceView({
         address={address}
         balance={balance}
         requestedAllowanceQuantityBase={
-          new BigNumber(requestedAllowanceQuantityBase)
+          new BigNumber(normalizeNumberValue(requestedAllowanceQuantityBase))
         }
-        value={new BigNumber(value || requestedAllowanceQuantityBase)}
+        value={
+          new BigNumber(
+            normalizeNumberValue(value || requestedAllowanceQuantityBase)
+          )
+        }
         onSubmit={onChange}
         footerRenderArea="sign-transaction-footer"
         addressAction={addressAction}


zerts:

**Result:** `pass`

### Screenshots

**Wallet overview — modified baseToCommon renders normal portfolio amounts (no regression)**

Wallet overview — modified baseToCommon renders normal portfolio amounts (no regression)

**Permit approval popup (mixed-case hex value from the bug report) — boots and renders, no NaN**

Permit approval popup (mixed-case hex value from the bug report) — boots and renders, no NaN

## Summary

Persona `default` (`0x52Fb…78e59`, "new zerion hot wallet") unlocked cleanly and I
drove the PR head (`c7523f0d`) as an unpacked MV3 build. Two flows: (1) the wallet
**overview**, which exercises the modified `baseToCommon` in its hot path across a large
multi-chain portfolio, and (2) the **Permit approval popup** (`#/signTypedData`) fed the
exact bug-report payload — an EIP-712 Permit with `value: "0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF"`
(mixed-case hex). Both boot and render with no crash, no error boundary, no `NaN`, and no
non-noise console/network errors. The `simulate-signature` backend was reachable (HTTP 200).

The change is a display-only, parse-only fix (mixed-case hex → decimal via `BigInt` before
bignumber.js), so it does not alter the signing payload. It boots and the affected flow
renders — smoke gate passes. Correctness/design is `/loop-review`'s call.

## Verification log
- Profile: `zerion-wallet-extension` — persona `default`, unlocked with the robust
  recipe (settle 2.5s, assert `inputValue`, wait to leave `#/login`).
- Built `dist/` from PR head (`npm run build:chrome`, `dist/manifest.json` present).
- Flows driven:
  - `#/overview` → real balances/USD values render across all chains, no `NaN`.
  - `#/signTypedData?...` with the reported mixed-case-hex Permit → approval popup renders
    (title "Permit", origin `app.uniswap.org`, wallet, raw-data preview, "No Risks Found",
    Cancel/Sign). No `NaN`, no crash.
- Console/network: clean (0 non-noise errors; `zpi.zerion.io/wallet/simulate-signature/v1`
  → 200).
- Unit tests for the fix (`src/shared/units/convert.test.ts`): 4 passed, including the exact
  reported value `0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF` (no longer `NaN`) and decimal/number
  pass-through regressions.

### Harness limitation (why the amount line isn't visible in the screenshot)
The approval screen only renders the interpreted **amount** (the `baseToCommon` line the fix
touches) when `simulate-signature` returns an approve action with a fungible. In this
harness that endpoint returned `"action": null` for the synthetic direct-navigation Permit,
so the popup shows the raw-data fallback view instead of the amount line. I confirmed this is
**environmental, not value-dependent**: a control run with a *decimal* value (and a separate
USDC permit) also returned `action: null`. The live approve-action UI needs a real
connected-dApp session / on-chain nonce this harness can't produce. The fix's actual logic
(mixed-case hex → real amount, never `NaN`) is therefore verified by the passing unit tests
above, and the modified `baseToCommon` is shown rendering correctly in the overview hot path.

zerts:

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | code | isUnlimitedApproval still constructs new BigNumber without normalizeNumberValue (latent, not triggered today) (`src/modules/ethereum/transactions/appovals.ts:10`) |

## Summary

Reviewed PR #952 at head `c7523f0d` against WLT-1686. The bug: an EIP-712 Permit
`value` supplied as a **mixed-case** hex string (e.g. `0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF`)
rendered as `NaN mETH` in the approval UI, because `bignumber.js` returns `NaN`
for mixed-case hex while parsing all-upper/all-lower hex fine.

The fix is correct and minimal (3 files, +56/-5). It adds `normalizeNumberValue`
in `src/shared/units/convert.ts` — any `0x`-hex string is converted to a decimal
string via `BigInt` (case-insensitive) before reaching bignumber.js; non-hex
input (decimals, numbers, `BigNumber`) passes through untouched. `baseToCommon`
and `commonToBase` are routed through it, and `AllowanceView`'s two direct
`new BigNumber(...)` constructions are wrapped too.

I traced the actual render path, not just the unit-tested helper. The permit
amount shown in the approval popup is `allowanceQuantityCommon` =
`baseToCommon(requestedAllowanceQuantityBase, decimals).toFixed()`
(`SignTypedData.tsx:588-594`), where `requestedAllowanceQuantityBase` is the raw
`message.value` (`SignTypedData.tsx:92-95, 525-527`). That value is exactly the
mixed-case hex, and it now flows through the fixed `baseToCommon`. The "Edit
allowance" screen (`AllowanceView` → `AllowanceForm`) receives the raw value
through the two newly-normalized `new BigNumber(...)` calls, so all downstream
`AllowanceForm`/`getCommonQuantity` consumers get an already-parsed `BigNumber`.
Both user-visible surfaces are covered.

Deterministic proof of the render string (running the actual repo conversion on
the exact reported value + 18 decimals): old path → `NaN`; fixed path →
`21267647932558578408.597187050162094079`. Signing behaviour is unchanged — the
fix only touches display conversion; the signed payload keeps the dApp's original
`message.value` (a valid uint256, byte-identical hash for hex vs decimal form).

Tests pass, type-check clean, no regressions found. Verdict: **approved**. The
one finding is a minor, out-of-scope latent-robustness note that does not
manifest today.

## Findings

### [minor/code] isUnlimitedApproval still constructs `new BigNumber` without `normalizeNumberValue`
`src/modules/ethereum/transactions/appovals.ts:10`

```ts
export function isUnlimitedApproval(value?: BigNumber.Value | null) {
  return new BigNumber(value?.toString() || 0).gte(UNLIMITED_APPROVAL_AMOUNT);
}
```

This is the one approval-amount `new BigNumber(...)` in the codebase that is
**not** routed through `normalizeNumberValue`. If a raw mixed-case hex string ever
reached it, `new BigNumber('0xFf…')` → `NaN` → `.gte(...)` → always `false`
(i.e. a max-uint256-in-mixed-case-hex approval could be mis-classified as
non-unlimited).

**This does not manifest today and is out of the PR's scope.** Every current
caller passes an already-normalized value: `ActInfo.tsx:55` receives the
`applyCustomAllowance`-overwritten common decimal (which is `allowanceQuantityCommon`,
already normalized via the fixed `baseToCommon`), and `AllowanceForm.tsx:57,63`
receives a normalized `BigNumber`. The PRD explicitly scoped the fix to where raw
hex *currently* reaches bignumber.js, so this is not a defect in the PR — just the
only remaining un-hardened spot noted for completeness. Non-blocking; a one-line
`normalizeNumberValue(...)` wrap here would close the gap if desired.

## Verification log

- **Head SHA:** `c7523f0d` (hard-reset worktree to `origin/extension-permit-approval-amount-shows-nan-for-hex-value-wlt`).
- **Diff scope:** exactly 3 files — `convert.ts`, `convert.test.ts`, `AllowanceView.tsx`. No leftover debug code.
- **Tests:**
  - `npm run test:unit -- src/shared/units/convert.test.ts` → 4/4 pass (incl. the exact reported value `0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF` no longer `NaN`, plus decimal/number pass-through regressions).
  - Full `npm run test:unit` → 355 tests pass; 34/35 suites pass. The one failing suite (`estimateNetworkFee.test.ts`) fails only because it requires a `.env` `TEST_WALLET_ADDRESS` — a pre-existing environmental failure in a file the diff does not touch (verified: not in diff).
  - `npm run type-check` → clean.
- **Edge cases verified in `node`:** every string the regex `/^0x[0-9a-f]+$/i` accepts is `BigInt`-parseable (incl. `0X`-prefix); the strings `BigInt` throws on (`0x`, `0xg`) are rejected by the regex, so `normalizeNumberValue` cannot throw on hex-shaped input.
- **Render-path trace:** confirmed the displayed amount goes through the fixed `baseToCommon` (`SignTypedData.tsx:590`) and the two normalized `new BigNumber(...)` in `AllowanceView` — the fix covers the UI, not just the helper. Deterministic render value computed with the real repo conversion (see `render-value-proof.txt` in the scratch dir): fixed → `21267647932558578408.597187050162094079`, old → `NaN`.
- **Live:** did not spin up a fresh Playwright build this iteration. The amount line in the approval popup only renders when `wallet/simulate-signature` returns an approve action with a fungible, which requires a real connected-dApp session the local harness cannot produce — documented in the iteration-1 QA smoke gate (same head SHA `c7523f0d`), which built the unpacked MV3 extension, drove the `baseToCommon` hot path (wallet overview — no `NaN`, no regression) and the Permit approval popup with the exact bug payload (boots/renders, no `NaN`, no crash, `simulate-signature` → 200). Given that documented environmental limit, a fresh drive would re-hit the same wall with no added signal; the fix's user-visible outcome is instead proven deterministically (unit test on the exact reported value through the real render-path helper + full code trace).
- **Animation:** no animation/motion code in diff — `review-animations` pass skipped.
- **Console/network:** clean per the iteration-1 QA smoke gate (0 non-noise errors).
- **Scratch dir:** `/tmp/loop-review-fLCFd7` (`render-value-proof.txt`).

src/shared/units/convert.ts
```diff
@@ -1,9 +1,28 @@
 import { BigNumber } from 'bignumber.js';
 
+const HEX_STRING_PATTERN = /^0x[0-9a-f]+$/i;
```
zerts: is there a way to check the hex status of the string available in the ethers library?

zerts: Good call — replaced the custom regex with ethers' `isHexString` (already a dependency, v6.14.0). Kept a `value.length > 2` guard because `isHexString` returns true for a bare `0x`, which `BigInt('0x')` throws on. Verified it still handles the mixed-case hex from this ticket plus the all-upper/all-lower cases; tests pass.

zerts:

**Result:** `pass`

### Screenshots

**Wallet overview — portfolio amounts via baseToCommon/normalizeNumberValue, no NaN**

Wallet overview — portfolio amounts via baseToCommon/normalizeNumberValue, no NaN

**Permit approval (signTypedData) with WLT-1686 mixed-case hex value — renders, no NaN**

Permit approval (signTypedData) with WLT-1686 mixed-case hex value — renders, no NaN

## Summary

Verified PR #952 at head `0fc573e5` (includes the review-fix commit that swaps the custom
hex regex for ethers' `isHexString`). Persona: `default` — unlocked with the seeded vault,
active wallet `0x52Fb…78e59` ("new zerion hot wallet"). The MV3 extension built cleanly
(`dist/manifest.json` present, `build:chrome` exit 0 — the new named `isHexString` import
from `ethers` bundles fine), the popup unlocked, and both affected surfaces rendered:

- **Overview** — full multi-chain portfolio ($777 total) with real balances and USD
  values, all routed through `baseToCommon` → `normalizeNumberValue`. No `NaN` anywhere
  (the hot path the fix touches is intact).
- **Permit approval** (`#/signTypedData`) driven with the exact WLT-1686 payload
  (`value: "0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF"`, mixed-case hex, mETH domain). The screen
  boots and renders: "Permit" title, `app.uniswap.org` origin, wallet row, raw typed-data
  Data panel, Cancel/Sign. No `NaN`, no crash, no error boundary. A DOM text scrape for
  `/NaN/` returned false.

Change boots and the affected flow renders → `qa: pass`.

## Verification log
- Profile: zerion-wallet-extension — MV3 extension, no dev server. Persona `default`,
  no flag override needed (the fix is on the always-on display path).
- Built: `dist/` from head `0fc573e5` (`build:chrome` exit 0).
- Flows: unlock → `#/overview` (portfolio render) → `#/signTypedData?...` (Permit approval
  with mixed-case hex Permit value).
- Console/network: the only non-noise entries are the Cloudflare Turnstile challenge
  (`simulate-signature` → 403, `brunhild.challenges.cloudflare.com` →
  `ERR_ADDRESS_UNREACHABLE`, plus a 401) — the documented headless-UA limitation, visible
  as the "Verifying…" Turnstile widget in `02`. Because the backend interpretation is
  blocked, the screen shows the "unable to simulate … proceed with caution" fallback, so
  the *interpreted* approve-amount line cannot render in this harness — independent of the
  fix. That exact parsing (mixed-case hex → correct decimal) is covered by the PR's passing
  unit tests. The `%c%d …NaN` console line seen in an earlier run is emitted by the
  externally-loaded Cloudflare challenge script (absent from `src/` and `dist/`), not the
  app.

zerts:

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | code | Uppercase 0X-prefix hex no longer normalized (theoretical-only regression vs old regex) (`src/shared/units/convert.ts:15`) |
| minor | code | isUnlimitedApproval still constructs new BigNumber without normalizeNumberValue (latent, not triggered today) (`src/modules/ethereum/transactions/appovals.ts:10`) |

## Summary

Reviewed PR #952 at head `0fc573e5` against WLT-1686. This round's purpose was to
confirm the human PR-review comment was genuinely addressed. **It was.**

The single human review thread (on `src/shared/units/convert.ts`, now resolved)
asked: *"is there a way to check the hex status of the string available in the
ethers library?"* The implementer replaced the custom regex with ethers'
`isHexString`, keeping a `value.length > 2` guard, and the change landed at the
current head:

```ts
import { isHexString } from 'ethers';
export function normalizeNumberValue(value: BigNumber.Value): BigNumber.Value {
  // isHexString also returns true for a bare '0x', which BigInt can't parse,
  // so require at least one hex digit.
  if (isHexString(value) && value.length > 2) {
    return BigInt(value).toString();
  }
  return value;
}
```

I verified the new implementation is correct and preserves the fix. Empirically
comparing `isHexString(v) && v.length > 2` against the old regex over 15 inputs:

- **Mixed-case hex** — `0xFf`, `0xDE0b6B3a7640000`, `0xABCdef123`, and the exact
  bug value `0xFFFFFFFFFFFEFFFFFFfFFFFFFFFFFFF` — all return `isHexString → true`
  and normalize identically. The core bug (mixed-case hex → `NaN`) stays fixed;
  deterministic render value is `21267647932558578408.597187050162094079` (was `NaN`).
- **Bare `0x`** — `isHexString('0x') → true`, but the `value.length > 2` guard
  skips it (returns `'0x'` unchanged, same as before), so `BigInt('0x')` never
  throws. The guard and its rationale are correct.
- **Non-string input** (numbers, `BigNumber`) — `isHexString` short-circuits to
  `false` before `.length` is evaluated, so no crash. Type-checks because
  `isHexString`'s `value is string` predicate narrows the type.
- New vs old behavior is **identical for every realistic input**; the only
  divergence is uppercase `0X`-prefix (see finding 1), which never appears in real
  payloads.

The rest of the fix (routing `baseToCommon`/`commonToBase` through
`normalizeNumberValue`, plus `AllowanceView`'s two normalized `new BigNumber`)
is unchanged from the iteration-1 head I already approved, and the render path is
unchanged. Unit tests pass (4/4 convert, 355 total), type-check clean, no
regression from the new `ethers` import. Verdict: **approved** — both findings are
minor and non-blocking.

## Findings

### [minor/code] Uppercase `0X`-prefix hex no longer normalized (theoretical-only)
`src/shared/units/convert.ts:15`

ethers' `isHexString` uses the regex `/^0x[0-9A-Fa-f]*$/` — lowercase `x` only —
so `isHexString('0Xff') → false`, whereas the old regex `/^0x[0-9a-f]+$/i` matched
uppercase-`X` prefixes. Result: a `0X`-prefixed value would now pass through
un-normalized and could render `NaN` in bignumber.js, where the old code parsed it.

**Not a real defect.** EIP-712 messages and JSON-RPC hex quantities are always
lowercase `0x` (the reported bug value included). No dApp emits `0X…`. This is a
theoretical micro-regression against an input that does not occur in practice, and
the implementer's choice to prefer the well-tested library predicate over a bespoke
regex — exactly what the reviewer asked for — is the right call. Non-blocking; noted
only for completeness.

### [minor/code] isUnlimitedApproval still constructs `new BigNumber` without `normalizeNumberValue`
`src/modules/ethereum/transactions/appovals.ts:10`

Carried over from the iteration-1 review (unchanged). `isUnlimitedApproval` is the
one approval-amount `new BigNumber(...)` not routed through `normalizeNumberValue`;
if a raw mixed-case hex ever reached it, `.gte(...)` would return `false` on `NaN`.
It does **not** manifest today — every current caller passes an already-normalized
value (`ActInfo.tsx:55` gets the `applyCustomAllowance`-overwritten common decimal;
`AllowanceForm.tsx:57,63` gets a normalized `BigNumber`). Out of the PR's scope,
non-blocking.

## Verification log

- **Head SHA:** `0fc573e5` (hard-reset to `origin/extension-permit-approval-amount-shows-nan-for-hex-value-wlt`). New commit since last review: "use ethers isHexString instead of custom regex".
- **Human review threads:** 1 thread (resolved/outdated) on `convert.ts` — reviewer `zerts` asked for an ethers-provided hex check; implementer replaced the regex with `isHexString` + `length > 2` guard and replied with reasoning. **Confirmed honored at head** (code present, behavior verified). No other human threads; PR-level comments are bots + `ai-runner` (skipped).
- **Diff scope:** 3 files — `convert.ts`, `convert.test.ts`, `AllowanceView.tsx`. No leftover debug code.
- **isHexString equivalence check** (`node`, real installed ethers v6): 15 inputs, new impl == old regex for every realistic case; mixed-case hex normalizes correctly; bare `0x` guarded; numbers short-circuit safely. See `ishexstring-proof.txt` in scratch dir.
- **Tests:**
  - `npm run test:unit -- src/shared/units/convert.test.ts` → 4/4 pass.
  - Full `npm run test:unit` → 355 tests pass; 34/35 suites. The one failing suite (`estimateNetworkFee.test.ts`) fails only for a missing `.env` `TEST_WALLET_ADDRESS` — pre-existing, in a file the diff does not touch. No regression from the new `ethers` import.
  - `npm run type-check` → clean (the `import { isHexString } from 'ethers'` and its type-predicate narrowing type-check fine).
- **Render path:** unchanged from the already-approved iteration; permit amount still flows through the now-fixed `baseToCommon` (`SignTypedData.tsx:590`) and the normalized `AllowanceView` constructions. Deterministic render value re-confirmed: `21267647932558578408.597187050162094079`.
- **Live:** did not spin up a fresh Playwright drive. The interpreted amount line only renders when `wallet/simulate-signature` returns an approve action with a fungible, which the local harness cannot produce (documented). The iteration QA smoke gate already drove **this exact head `0fc573e5`**: MV3 `build:chrome` exit 0 (the new named `isHexString` import bundles fine), popup unlocked, overview renders full portfolio via `baseToCommon`→`normalizeNumberValue` with no `NaN`, and the Permit approval popup with the exact WLT-1686 mixed-case-hex payload boots/renders with a DOM `/NaN/` scrape returning false. Since the fix is behaviorally identical to the approved iteration for all realistic inputs and the render path is unchanged, a fresh drive would re-hit the same simulate-signature/Turnstile wall with no added signal; the outcome is proven deterministically instead.
- **Animation:** no animation/motion code in diff — pass skipped.
- **Console/network:** clean per the QA smoke gate at head `0fc573e5` (only the documented Cloudflare Turnstile challenge noise from headless UA).
- **Scratch dir:** `/tmp/loop-review-BFqWCf` (`ishexstring-proof.txt`).

PR #956, opened 2026-07-08
WLT-1724: fix history filtering for local txs

zerts:
## Summary

Local (client-stored) transactions were leaking into the filtered transaction history, showing duplicated rows next to their backend twins and inflating traded volume (~2×). Root cause and decisions are documented in a PRD comment on the Linear issue.

- **`mergeLocalAndBackendActions` guard rewritten** (`src/ui/pages/History/History.tsx`): previously local actions were kept when newer than the *oldest loaded* backend action (threshold 0 when there was no next page), which the unfiltered view masked but filtered views exposed. Now a mined local action is shown only while it is *strictly newer than the newest backend action* — i.e. only while it bridges the backend indexing gap — matching Web. Pending local actions bypass the timestamp guard so stuck transactions stay reachable for Speed Up / Cancel; hash dedup remains as a second layer.
- **`performPurgeCheck` loop fix** (`src/background/transactions/TransactionService.ts`): `return` → `continue` — a single Solana item in the store aborted the whole purge run, so mined EVM transactions accumulated locally forever, feeding the leak.
- **Local `assetTypes=fungible` filter gap**: local actions are now required to contain a fungible transfer/approval when the fungible asset filter is active (the `nft` case already excluded all local items).

Closes WLT-1724

## Test plan

- `npm run type-check`, `npm run lint`, prettier — all pass
- Manual: with stale mined/dropped local txs in the store, open History with an action-type or asset-type filter → no duplicated trades, totals match the unfiltered view and Web
- Manual: send a transaction → it appears as pending in History (with and without matching filters) and stays visible while pending; after mining it is replaced by the backend entry once indexed

### DIFF
--- a/src/background/transactions/TransactionService.ts
+++ b/src/background/transactions/TransactionService.ts
@@ -220,7 +220,7 @@ export class TransactionService {
 
     for (const item of transactions) {
       if (!item.hash) {
-        return; // Do not handle Solana items
+        continue; // Do not handle Solana items
       }
       const chainId = normalizeChainId(item.transaction.chainId);
       const key = `${item.transaction.from}:${chainId}` as const;
--- a/src/ui/pages/History/History.tsx
+++ b/src/ui/pages/History/History.tsx
@@ -53,8 +53,7 @@ function sortActions<T extends { timestamp?: number }>(actions: T[]) {
 
 function mergeLocalAndBackendActions(
   local: AnyAddressAction[],
-  backend: AddressAction[],
-  hasMoreBackendActions: boolean
+  backend: AddressAction[]
 ) {
   const backendHashes = new Set(
     backend.flatMap(
@@ -63,11 +62,13 @@ function mergeLocalAndBackendActions(
     )
   );
 
-  const lastBackendActionDatetime = backend.at(-1)?.timestamp;
-  const lastBackendTimestamp =
-    lastBackendActionDatetime && hasMoreBackendActions
-      ? new Date(lastBackendActionDatetime).getTime()
-      : 0;
+  // Backend actions are sorted newest-first. A mined local action that is
+  // older than the newest backend action already belongs to the backend
+  // history, so rendering it would duplicate the backend entry even when the
+  // hashes don't match (e.g. replaced or accelerated transactions). Pending
+  // actions are kept regardless: they aren't indexed yet, and hiding them
+  // would remove the entry point for speeding up or cancelling them.
+  const newestBackendTimestamp = backend.at(0)?.timestamp ?? 0;
 
   const merged = local
     .filter(
@@ -78,7 +79,7 @@ function mergeLocalAndBackendActions(
           (act) =>
             act.transaction.hash && backendHashes.has(act.transaction.hash)
         ) &&
-        tx.timestamp >= lastBackendTimestamp
+        (tx.status === 'pending' || tx.timestamp > newestBackendTimestamp)
     )
     .concat(backend);
   return sortActions(merged);
@@ -157,6 +158,14 @@ function useMinedAndPendingAddressActions({
       }
       if (assetTypes?.length === 1 && assetTypes?.[0] === 'nft') {
         items = items.filter(() => false);
+      } else if (assetTypes?.length === 1 && assetTypes?.[0] === 'fungible') {
+        items = items.filter((item) =>
+          [item.content, ...(item.acts?.map((act) => act.content) ?? [])].some(
+            (content) =>
+              content?.transfers?.some((transfer) => transfer.fungible) ||
+              content?.approvals?.some((approval) => approval.fungible)
+          )
+        );
       }
       return items;
     },
@@ -180,14 +189,9 @@ function useMinedAndPendingAddressActions({
 
   return useMemo(() => {
     const backendItems = isSupportedByBackend && actions ? actions : [];
-    const hasMore = Boolean(isSupportedByBackend && queryData.hasNextPage);
     return {
       actions: localAddressActions
-        ? mergeLocalAndBackendActions(
-            localAddressActions,
-            backendItems,
-            hasMore
-          )
+        ? mergeLocalAndBackendActions(localAddressActions, backendItems)
         : null,
       ...localActionsQuery,
       isLoading:


zerts:

**Result:** `pass`

### Screenshots

**Unlocked overview, portfolio loaded**

Unlocked overview, portfolio loaded

**Unfiltered History — seeded pending Send visible at top, no stale local leak**

Unfiltered History — seeded pending Send visible at top, no stale local leak

**History with actionTypes=trade — only backend trades, seeded stale Jan-2024 local trade absent**

History with actionTypes=trade — only backend trades, seeded stale Jan-2024 local trade absent

**History with assetTypes=fungible — pending send passes (fungible transfer), NFT mints excluded, no stale leak**

History with assetTypes=fungible — pending send passes (fungible transfer), NFT mints excluded, no stale leak

## Summary

Persona "default" (active wallet `0x52Fb…78e59`, "new zerion hot wallet") on PR head
`a15626c8`. Built `dist/` via `npm run build:chrome`, seeded the captured
`chrome.storage.local` state **plus two synthetic local transactions** crafted to
exercise the fix: (1) a stale *confirmed* local trade dated Jan 5 2024 with a hash
unknown to the backend — the exact class of record that leaked into filtered history
before this fix — and (2) a fresh *pending* send with nonce 999999999 (so the poller
can't drop it) to verify the pending bypass. Unlocked and drove the History tab
unfiltered, with `actionTypes=trade`, and with `assetTypes=fungible`. The app boots,
all three History states render real backend data, the stale local trade never appears
in any view, the pending local send stays visible where the filters allow it, and no
non-noise console errors or request failures occurred.

## Verification log
- Profile: zerion-wallet-extension — persona "default", no feature-flag overrides needed
- Built: `npm run build:chrome` → `dist/` OK (no dev server for MV3; qaPort unused per profile)
- Flows: unlock → `#/overview/history` → `#/overview/history?actionTypes=trade` → `#/overview/history?assetTypes=fungible`
- Programmatic checks (page innerText): `unfiltered_shows_pending: true`, `unfiltered_stale_leak: false`, `filtered_stale_leak: false`, `filtered_has_rows: true`, `fungible_stale_leak: false`
- Trade-filtered view also correctly excludes the pending *send* (local actionTypes filter) while the fungible view includes it (it carries a 0-ETH fungible transfer)
- Console/network: clean after standard noise filter (0 errors, 0 failed requests; zpi.zerion.io responses healthy — portfolio and history data loaded)

zerts:

**Verdict:** `approved`

## Summary

Reviewed PR #956 at head `a15626c8b7f88160f41cd399f3a5867583f009fb` against WLT-1724 and the PRD comment on the Linear issue. The change does exactly what the PRD decided: `mergeLocalAndBackendActions` now keeps a non-pending local action only when it is strictly newer than the newest backend action (matching Web), pending actions bypass the timestamp guard so Speed Up / Cancel stays reachable, `performPurgeCheck` no longer aborts the whole purge run on the first Solana item (`return` → `continue`), and local actions are now required to carry a fungible transfer/approval under the `assetTypes=fungible` filter. Code review found no defects; type-check passes; live verification in a built extension with four purpose-crafted seeded local transactions confirmed the full expected visibility matrix across five History views with a clean console. No human review threads existed on the PR to cross-check. Approved.

Code-review notes (all verified against the code, none are findings):
- Timestamp comparison is sound: backend timestamps are normalized to ms in `walletGetActions` (`timestamp * 1000`, `src/modules/zerion-api/requests/wallet-get-actions.ts:309`) and local actions use `Date.now()` ms, so `tx.timestamp > newestBackendTimestamp` compares like units. Backend pages arrive newest-first, so `backend.at(0)` is the correct boundary.
- With `backendItems = []` (chain unsupported by backend, or nothing loaded) the threshold falls back to 0 and all local history still renders — no regression for backend-unsupported chains.
- Local statuses are exactly `confirmed | failed | pending | dropped` (`getTransactionObjectStatus`); only `pending` bypasses the guard, per the PRD deviation.
- The removed `hasMoreBackendActions` parameter has no other callers; `queryData.hasNextPage` is still used for the infinite-scroll `hasMore` at `src/ui/pages/History/History.tsx:431`.
- `parseAssetTypes` only ever yields `['fungible']`, `['nft']`, or `[]`, so the new `else if` branch covers every reachable case; local transfer/approval objects always carry a `fungible` field (possibly null, e.g. NFT sends), which the filter handles correctly.
- The `continue` fix in `performPurgeCheck` (`src/background/transactions/TransactionService.ts:223`) matches the function's own doc comment ("do not handle Solana items") — the old `return` silently cancelled purging for all EVM entries after the first Solana item, which is exactly how the reporter accumulated weeks of stale local trades.
- No animation code in diff.

## Findings

None. One pre-existing test failure and one accepted edge are documented in the verification log below; neither is caused by this diff.

## Verification log

- Tests: `npm run type-check` — pass. `npm run test:unit` — 353/354 pass; the single failure (`src/modules/ethereum/transactions/gasPrices/estimateNetworkFee.test.ts`, optimistic fee expects hardcoded values that drift with time) also fails identically on `main` at 76c2f438 — pre-existing, unrelated to the diff (does not touch gas pricing), does not block.
- E2E: `npx playwright test` (repo suite: onboarding/login/get-started/overview/settings — no History coverage) was started but had not completed after ~25 minutes of real-Chrome runs and was stopped at the user's request; partial artifacts showed only snapshot-comparison churn in onboarding/overview flows untouched by this diff. Inconclusive but out-of-scope for the changed code; the changed flow was verified live instead (below).
- Live: built `dist/` via `npm run build:chrome` at the PR head, seeded persona "default" (`chrome.storage.local` dump) **plus four crafted local transactions** for wallet `0x52Fb…78e59`, each targeting one branch of the new guard:
  - `QASTALE` — confirmed local trade, Jan 5 2024, hash unknown to backend (the exact leak class from the bug report);
  - `QADROP` — dropped local send, Feb 1 2024;
  - `QAPENDOLD` — pending local send timestamped Jan 15 2025, i.e. *older* than the newest backend action (Jul 2026) — visible only via the pending bypass branch;
  - `QAPENDNEW` — pending local send timestamped now (nonce 999999999 so the poller can't drop it).
  Drove five History views (`#/overview/history` unfiltered, `?actionTypes=trade`, `?actionTypes=send`, `?assetTypes=fungible`, `?assetTypes=nft`) at the popup viewport. Observed matrix, asserted programmatically via page text and confirmed by reading the screenshots:
  - `QASTALE` and `QADROP`: absent in **all five** views — the duplicated-trades/inflated-volume leak is fixed, including the trade-filtered view from the report;
  - `QAPENDOLD`: visible unfiltered, under `send`, and under `fungible` (rendered at its Jan 15 2025 position with a "Pending" badge) despite being older than the newest backend action — the pending bypass genuinely exercised; correctly hidden under `trade` and `nft`;
  - `QAPENDNEW`: visible at the top unfiltered / under `send` / under `fungible`; hidden under `trade` and `nft`;
  - backend rows render real data in every view (trades Jul 2026, NFT mints under `nft`, etc.), no error boundaries, no stuck skeletons.
  Screenshots + rendered-text dumps: `/tmp/loop-review-Axg3fs/art/` (`02-history-unfiltered.png` … `06-history-nft.png`, `text-*.txt`); driver script `/tmp/loop-review-Axg3fs/drive.mjs`.
- Console/network: clean — 0 console errors, 0 failed requests after the standard noise filter; `zpi.zerion.io` responses healthy (portfolio and history loaded).
- Accepted edge (documented in the PRD, matches Web): a mined-but-not-yet-indexed local tx can be briefly hidden if a newer backend action is already indexed; self-heals on the next backend poll. Not a finding.
- No human review threads or PR comments to reconcile (only the linear-linkback, CI build artifact, and the ai-runner QA gate comment, which was skipped per protocol).

PR #962, opened 2026-08-12
WLT-1998: show the full recipient address in address picker rows

zerts:
## Summary

In the Send flow's recipient picker, a row showed the truncated address on **both** lines — the title fell back to `0xd8f50…0de1e` and the subtitle repeated it. So at the moment the user clicked to commit a freshly pasted recipient, the full address wasn't visible anywhere on screen (the search input has scrolled past its head by then), and there was nothing to check a clipboard-hijack swap against.

The second line of each row now carries the **full** address.

- New shared `FullAddress` component. It ellipsizes the head and pins the last 6 characters, so when space runs out it's never the tail — the part people actually compare — that gets cut. A 42-char EVM address fits outright at `caption/regular` in the 425px popup; longer Solana addresses degrade gracefully.
- This markup was already hand-rolled in three places (SendForm2 selector, SwapForm2 selector, PerpsWithdraw) and the fix would have made it four, so the SendForm2 and SwapForm2 recipient selectors are migrated onto the shared component. `PerpsWithdraw` is left alone — unrelated surface.
- Line 1 keeps its existing fallback (name → handle → truncated). On a nameless pasted address that means the truncated form sits above the full one, which is strictly more information than the duplicate it replaces.

Because `ReceiverAddressDialog` is shared, the fix also lands in Swap, Address Book, the Overview send button, and Asset Info — the verification concern is identical on all of them. Decisions and rejected alternatives are written up in the [PRD comment on the issue](https://linear.app/zerion/issue/WLT-1998/extension-send-flow-cant-verify-full-recipient-address-before).

Closes WLT-1998

## Test plan

- [ ] Send flow → tap the `To:` field → paste an EVM address. The **Exact match** row's second line shows the full address, all 42 characters, matching what was pasted.
- [ ] Same with a Solana address from a Solana wallet — the head ellipsizes if needed, the last 6 characters stay visible.
- [ ] Rows under Recents / Address Book / My wallets / Watchlist show name or handle on line 1 and the full address on line 2; nothing wraps and the row height is unchanged.
- [ ] Hovering a row shows the full address as a native tooltip.
- [ ] Send form's `To:` field and Swap's recipient field still render their address line as before (no visual regression from the shared-component migration).
- [ ] Address Book, Overview send button, and Asset Info recipient pickers show the same full-address rows.

### DIFF
--- a/src/ui/components/FullAddress/FullAddress.module.css
+++ b/src/ui/components/FullAddress/FullAddress.module.css
@@ -1,0 +1,17 @@
+.row {
+  display: grid;
+  grid-template-columns: auto minmax(min-content, 1fr);
+  align-items: center;
+  min-width: 0;
+  justify-content: start;
+}
+
+.head {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.tail {
+  white-space: nowrap;
+}
--- a/src/ui/components/FullAddress/FullAddress.tsx
+++ b/src/ui/components/FullAddress/FullAddress.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import cn from 'classnames';
+import { UIText } from 'src/ui/ui-kit/UIText';
+import type { Kind as UITextKind } from 'src/ui/ui-kit/UIText';
+import * as styles from './FullAddress.module.css';
+
+const TAIL_LENGTH = 6;
+
+/**
+ * Renders the whole address so that it can be verified against the source it
+ * was copied from. When there isn't enough room for it, the head is ellipsized
+ * and the last {@link TAIL_LENGTH} characters stay pinned — the tail is the
+ * part people compare, so it must never be the part that gets cut.
+ */
+export function FullAddress({
+  address,
+  kind = 'caption/regular',
+  color = 'var(--neutral-500)',
+  className,
+}: {
+  address: string;
+  kind?: UITextKind;
+  color?: string;
+  className?: string;
+}) {
+  const head = address.slice(0, -TAIL_LENGTH);
+  const tail = address.slice(-TAIL_LENGTH);
+
+  return (
+    <div className={cn(styles.row, className)} title={address}>
+      <UIText kind={kind} color={color} className={styles.head}>
+        {head}
+      </UIText>
+      <UIText kind={kind} color={color} className={styles.tail}>
+        {tail}
+      </UIText>
+    </div>
+  );
+}
--- a/src/ui/components/FullAddress/index.ts
+++ b/src/ui/components/FullAddress/index.ts
@@ -1,0 +1,1 @@
+export { FullAddress } from './FullAddress';
--- a/src/ui/components/ReceiverAddressDialog/AddressListItem.tsx
+++ b/src/ui/components/ReceiverAddressDialog/AddressListItem.tsx
@@ -4,6 +4,7 @@ import { normalizeAddress } from 'src/shared/normalizeAddress';
 import { truncateAddress } from 'src/ui/shared/truncateAddress';
 import { useWalletsMetaByChunks } from 'src/ui/shared/requests/useWalletsMetaByChunks';
 import { BlockieImg } from 'src/ui/components/BlockieImg';
+import { FullAddress } from 'src/ui/components/FullAddress';
 import { UIText } from 'src/ui/ui-kit/UIText';
 import { VStack } from 'src/ui/ui-kit/VStack';
 
@@ -64,9 +65,9 @@ export function AddressListItem({
   const previewUrl = meta?.nft?.previewUrl ?? null;
   const firstHandle = meta?.identities?.[0]?.handle ?? null;
 
-  const truncated = truncateAddress(normalizeAddress(address), 5);
-  const title = localName || firstHandle || truncated;
-  const subtitle = truncated;
+  const normalizedAddress = normalizeAddress(address);
+  const title =
+    localName || firstHandle || truncateAddress(normalizedAddress, 5);
 
   return (
     <>
@@ -105,17 +106,7 @@ export function AddressListItem({
         >
           {title}
         </UIText>
-        <UIText
-          kind="caption/regular"
-          color="var(--neutral-500)"
-          style={{
-            overflow: 'hidden',
-            textOverflow: 'ellipsis',
-            whiteSpace: 'nowrap',
-          }}
-        >
-          {subtitle}
-        </UIText>
+        <FullAddress address={normalizedAddress} />
       </VStack>
     </>
   );
--- a/src/ui/pages/SendForm2/ReceiverAddressSelector/ReceiverAddressSelector.module.css
+++ b/src/ui/pages/SendForm2/ReceiverAddressSelector/ReceiverAddressSelector.module.css
@@ -61,24 +61,6 @@
   pointer-events: auto;
 }
 
-.addressRow {
-  display: grid;
-  grid-template-columns: auto minmax(min-content, 1fr);
-  align-items: center;
-  min-width: 0;
-  justify-content: start;
-}
-
-.addressHead {
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-}
-
-.addressTail {
-  white-space: nowrap;
-}
-
 .skeletonAvatar {
   width: 24px;
   height: 24px;
--- a/src/ui/pages/SendForm2/ReceiverAddressSelector/ReceiverAddressSelector.tsx
+++ b/src/ui/pages/SendForm2/ReceiverAddressSelector/ReceiverAddressSelector.tsx
@@ -6,6 +6,7 @@ import { isMatchForEcosystem } from 'src/shared/wallet/shared';
 import type { BlockchainType } from 'src/shared/wallet/classifiers';
 import { truncateAddress } from 'src/ui/shared/truncateAddress';
 import { BlockieImg } from 'src/ui/components/BlockieImg';
+import { FullAddress } from 'src/ui/components/FullAddress';
 import { UIText } from 'src/ui/ui-kit/UIText';
 import { UnstyledButton } from 'src/ui/ui-kit/UnstyledButton';
 import { Button } from 'src/ui/ui-kit/Button';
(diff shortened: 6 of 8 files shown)


zerts:

**Result:** `pass`

### Screenshots

**The ticket's scenario, fixed — search input has scrolled past the head, Exact match row shows the full address**

The ticket's scenario, fixed — search input has scrolled past the head, Exact match row shows the full address

**Recipient picker — every row carries the full address on line 2**

Recipient picker — every row carries the full address on line 2

**Send form after picking — To: field shows the full address, live fee quote**

Send form after picking — To: field shows the full address, live fee quote

**Degradation at 340px — head ellipsizes, tail stays pinned**

Degradation at 340px — head ellipsizes, tail stays pinned

**Solana picker — 44-char base58 addresses render in full**

Solana picker — 44-char base58 addresses render in full

**Swap receiver field after the shared-component migration, live cross-ecosystem quote**

Swap receiver field after the shared-component migration, live cross-ecosystem quote

**Send form before a recipient is chosen**

Send form before a recipient is chosen

## Summary

Persona "default" (active wallet `0x52Fb…78e59`), extension built from PR head `ddbdca0f` and
loaded unpacked. Both affected flows boot and render. Pasting a checksummed address
(`0xd8dA6BF2…A96045`) into the Send recipient picker reproduces the ticket's exact setup — the
search input has scrolled past the head of the address — and the **Exact match** row now shows the
complete 42-character address on its second line, so it is verifiable at the moment of clicking.
Picking it fills the `To:` field, which also shows the full address, and a live network fee ($0.04)
resolves. The Swap receiver selector, migrated onto the same shared `FullAddress` component in this
diff, renders correctly with a live cross-ecosystem quote. No console errors, no request failures.

## Verification log

- Profile: zerion-wallet-extension — persona "default", no flag override needed (both forms are ungated)
- Built: `npm run build:chrome` → `dist/manifest.json` present (exit 0); no dev server (extension)
- Flows:
  - `#/send-form/?inputChain=ethereum&inputFungibleId=eth&inputAmount=0.0001` → "Select recipient" → picker → paste address → pick → back to form
  - `#/swap-form?inputChain=ethereum&outputChain=solana&…` → "Solana Recipient Address" → picker → pick Solana wallet
- Console/network: clean. Send flow: 0 errors, 0 failed requests. Swap flow: one bare `400` during
  the cross-ecosystem pair transition, which this profile documents as expected quote noise; the
  quote itself resolved (1.868 USDT, $0.05 fee, "Bridge & Send Now" enabled).

### Numeric assertions (`page.evaluate`, CSS-layout change)

Exact-match row after pasting, at the real popup width:

| property | value |
|---|---|
| `line2Rendered` | `0xd8da6bf26964af9d7eed9e03e53415d37aa96045` (all 42 chars) |
| matches pasted address (lowercased) | yes |
| `headEllipsized` | `false` — nothing is cut |
| gap between head and tail | `0px` — tail sits flush, not pushed to the right edge |
| rendered width / container | 295px / 295px |
| row height | 60px — unchanged, still inside the virtualized 64px slot |

- **Saved rows** (8 rows across Recents / Address Book / My wallets / Watchlist): every one renders
  its full 42-char address with `headEllipsized: false` and a 0px gap. The `0x56df3…353cd` row
  exercises the no-name fallback — truncated on line 1, full address on line 2 — as specified.
- **Degradation path** (viewport forced to 340px): `headEllipsized: true`, head overflows by 35px,
  and the tail `a96045` stays visible. Confirms the head-first, tail-pinned behavior the diff intends.
- **Solana** (44-char base58, the tightest case): renders in full in the picker at popup width, no
  ellipsis. In the Swap *field* it ellipsizes the head — that field reserves 36px for the clear
  button — with the tail `dMq1Xu` pinned.

### Observation for review (not a gate failure)

When the head ellipsizes, a small blank gap renders between the `…` and the pinned tail (visible in
`04` and `07`). This is the browser's standard `text-overflow: ellipsis` clipping artifact and is
pre-existing — the identical grid already shipped in the Send form's `To:` field. Cosmetic, and
outside this gate's scope.

zerts:

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | visual | Blank gap between the ellipsis and the pinned tail when the head overflows (`src/ui/components/FullAddress/FullAddress.module.css:9`) |
| minor | code | FullAddress `kind`/`color` props are unused by every call site (`src/ui/components/FullAddress/FullAddress.tsx:17`) |

## Summary

Reviewed head `ddbdca0f` (PR #962, `WLT-1998`). The diff is small and does exactly
what the Linear issue and the PRD comment specify: `AddressListItem`'s second line
now renders the **full** normalized address instead of repeating the truncated form
from line 1, via a new shared `src/ui/components/FullAddress` primitive, and the
SendForm2 / SwapForm2 recipient selectors are migrated onto that primitive (their
duplicated `.addressRow/.addressHead/.addressTail` CSS is deleted). `PerpsWithdraw`
is left alone, as the PRD scoped.

Verified live in the built extension: pasting `0xd8dA6BF2…A96045` into the Send
recipient picker reproduces the ticket's exact setup — the search input has scrolled
past the head of the address (`scrollLeft: 35`) — and the **Exact match** row's second
line renders all 42 characters with `headEllipsized: false`, a `0px` gap between head
and tail, and a 60px row height (unchanged, still inside the virtualized 64px slot).
Saved rows across Recents / Address Book / My wallets / Watchlist all render the full
address; the `0x56df3…353cd` row exercises the nameless fallback (truncated on line 1,
full on line 2) as the PRD describes. Solana's tighter 44-char base58 addresses render
in full in the picker with case preserved (`normalizeAddress` only lowercases `0x…`),
and the head-ellipsize/tail-pinned degradation works when width is genuinely short.
The Overview send-button picker inherits the fix. `tsc --noEmit`, eslint on the touched
files, and the unit suite pass; the two failing tests in the repo also fail on
`origin/main` (confirmed by building and running the base tree), so neither is caused
by this diff.

Two `minor` findings, neither blocking. Worth noting for the author rather than as
findings: the PR's test-plan line "Swap's recipient field still renders its address
line as before" is not quite right — SwapForm2's old grid was `minmax(0, 1fr) auto`,
which right-aligned the tail against the field's edge, while the shared component uses
SendForm2's `auto minmax(min-content, 1fr)`, so the tail is now flush after the head.
That's the PRD's intended unification and strictly better (the old formulation opened a
gap mid-address whenever the address fit), so it's an improvement, not a regression.
Also, a fourth hand-rolled copy of this markup lives in the legacy
`SendForm/fieldsets/AddressInput` (the PRD counted three) — legacy surface, fine to skip.

## Findings

### [minor/visual] Blank gap between the ellipsis and the pinned tail when the head overflows
`.head` uses `text-overflow: ellipsis`, so when the head is clipped the browser leaves
the partial glyph's advance width blank before the pinned tail, rendering
`0xd8da6bf26964af9d7eed9e03e53…␣a96045`. Visible in `05-exact-match-narrow.png`
(340px viewport, head overflows by 35px) and `09-swap-receiver-field.png` (Swap's field
reserves 36px for its clear button, so a 44-char Solana address overflows by 25px there
even at the real 425px popup width). Cosmetic only — the tail, which is the part the
ticket cares about, always stays visible — and it is a pre-existing artifact of the same
grid already shipping in SendForm2's `To:` field, so this diff inherits rather than
introduces it. At the real popup width no EVM (295px of 295px) or Solana (321px) picker
row ellipsizes at all, so the ticket's own scenario never hits it.

### [minor/code] FullAddress `kind`/`color` props are unused by every call site
`FullAddress` accepts `kind` and `color` overrides, but all three call sites
(`AddressListItem`, SendForm2, SwapForm2) take the defaults, and the two remaining
hand-rolled copies that could later migrate onto it (`PerpsWithdraw:349`,
`SendForm/fieldsets/AddressInput:445`) also use exactly `caption/regular` +
`var(--neutral-500)`. Speculative API surface on a brand-new primitive; `className` is
likewise unused today. Harmless either way — worth dropping until a second style is
actually needed.

## Verification log

- Deps: `npm install` (no `node_modules` in this worktree) → exit 0. `.env` copied from
  the main clone (gitignored; required by the jest samples module).
- Type check: `npm run type-check` (`tsc --noEmit`) → exit 0.
- Lint: `npx eslint` on the four touched/new source files → exit 0, no output.
- Unit: `npm run test:unit` → 363 passed, 36/37 suites pass.
  `estimateNetworkFee.test.ts` fails (hardcoded `estimatedFee`/`maxFee` vs. computed
  values differing by ~0.002%). **Pre-existing**: the same suite fails identically on a
  clean `origin/main` tree (`git archive origin/main` → `/tmp/wlt1998-base`). Touches no
  file in the diff.
- E2E: `npx playwright test` → 20 passed, 1 failed —
  `e2e/tests/overview.spec.ts:22 › Overview › Matches snapshot` times out waiting for
  "Get started". **Pre-existing**: reproduces on the base tree after building its own
  `dist/`, and the trace's final frame shows the run still parked on the onboarding
  "Nicely Done" tab, i.e. a fixture/navigation issue in the harness, unrelated to the
  recipient picker. Also re-ran it twice at head (`--repeat-each=2`) — deterministic,
  not flaky.
- Build: `npm run build:chrome` → exit 0, `dist/manifest.json` present.
- Live: extension built from head, loaded unpacked into Playwright chromium
  (`--headless=new`), persona "default" seeded into `chrome.storage.local` and unlocked;
  viewport 440×900 against the app's fixed `--body-width: 425px` (the real popup width),
  plus a 340px pass for the degradation path. Screenshots + `page.evaluate`
  measurements in `/tmp/loop-review-RTZ0v7/shots` and
  `/tmp/loop-review-RTZ0v7/measurements.json`:
  - `03-picker-saved-rows.png` — 8 rows across Recents / Address Book / My wallets /
    Watchlist, every one full-address, `headEllipsized: false`, `gap: 0`, 60px rows.
  - `04-exact-match.png` — the ticket's scenario: input scrolled past the head, row
    line 2 = all 42 chars of the pasted address (lowercased), read back and compared.
  - `05-exact-match-narrow.png` — 340px: `headEllipsized: true`, tail `a96045` intact.
  - `06-send-form-filled.png` — `To:` field full address + live $0.04 network fee.
  - `08-swap-picker-solana.png` — 44-char base58 rows in full, case preserved.
  - `09-swap-receiver-field.png` — post-migration Swap field, tail `44u1JJ` pinned, no
    overlap with the clear button (273px of a 309px column that reserves 36px).
  - `11-overview-send-picker.png` — the Overview send-button consumer inherits the fix.
  - Address Book and AssetInfo pickers not driven separately: same
    `ReceiverAddressDialog` component, and no consumer prop affects the row rendering.
- Console/network: clean apart from one bare `400` during the cross-ecosystem swap pair
  transition, which the project's QA profile documents as expected quote noise; the
  quote resolved. No failed requests to `zpi.zerion.io`/`api.zerion.io`.
- Animation: no animation or motion code in the diff (the removed CSS blocks carried no
  transitions; `AddressListItem`'s `motion.img` avatar is untouched).
- PR threads: none — GraphQL returns `reviewThreads: []`; the only PR comments are the
  Linear linkback, the build artifact bot, and an earlier `ai-runner` QA post. No human
  feedback to honor.

PR #963, opened 2026-08-21
WLT-2189: interpret the whole Solana batch, drop local action fallbacks

zerts:
## Summary

Two related problems in `SendTransaction.tsx`:

1. **Local action fallbacks.** The page built an `AddressAction` by parsing the dapp's raw transaction client-side — `incomingTxToIncomingAddressAction` for EVM, `parseSolanaTransaction` for Solana — and showed it whenever the backend interpretation was missing. Parsing an untrusted payload that way produces confident-but-wrong summaries, which is exactly the surface we don't want in a signing confirmation.
2. **`signAllTransactions` only ever saw `transactions[0]`.** It interpreted, displayed and detailed the first transaction, so the user approved N transactions after being shown 1.

The backend interpretation is now the single source of the displayed action, and the whole Solana batch is sent to the interpretation endpoint and is inspectable in Details. `firstTx` is gone.

## What changed

- `addressAction` is `interpretQuery.data?.data.action ?? null` on both the EVM and Solana paths; `useLocalAction` and the `parseSolanaTransaction` fallback are deleted from this file.
- **Uninterpreted state** keeps the existing layout and just drops the asset rows. Title falls back to **"Sign Transaction"** / **"Sign Transactions"**, with an **"N transactions"** subtitle for batches of more than one.
- **Blank only while the interpretation is genuinely in flight.** `isInitialLoading` is `false` for a *disabled* query, which is what we want for a network with `supports_simulations: false` (the request never runs, so we go straight to the uninterpreted state rather than hanging forever) — but not while we're deliberately holding the request back waiting for gas prices, hence the explicit `isWaitingForGasrices` check.
- The full batch goes to `wallet/simulate-transactions/v1`. The plural endpoint already accepted an array; only the call site was passing one.
- `TransactionAdvancedView`'s solana branch takes `StringBase64[]` and renders each transaction in its own `Surface` labelled "Transaction 1 of 3"; *Copy Raw Data* copies the whole array. One mechanical call-site update in `TransactionSimulation.tsx`.
- `TransactionWarnings` accepts a null action so the **insufficient-funds warning keeps working** without an interpretation; only the "may fail" branch (which reads `.status`) needs one.
- The defi-sdk client leaves both interpret query keys — it was only ever a testnet-mode proxy, replaced by the explicit `source` key on the Solana side (the EVM key already had it).

`addressAction: null` needed no downstream work: `TransactionContextParams.addressAction` is already nullable and `addressActionToAnalytics` already maps null to `action_type: 'Execute'`.

## Out of scope

`TransactionSimulation.tsx`'s own local-action fallbacks stay. That component serves Zerion's **own** Send/Swap/Bridge forms, where `fallbackAddressAction` is built from our own form state rather than parsed from an untrusted dapp payload — different trust model. `incomingTxToIncomingAddressAction` / `parseSolanaTransaction` are therefore still used and not deleted. Hardware-wallet support for `signAllTransactions` still throws, as before.

Closes WLT-2189

## Test plan

- [ ] **Batched Solana** — trigger `signAllTransactions` from Jupiter (limit order / DCA setup), Meteora, Kamino, or a Tensor/Magic Eden bulk listing. Confirm the interpretation reflects the whole batch, not just the first tx, and that Details lists every transaction with "Transaction _n_ of _N_" labels.
- [ ] **Batch chrome** — an uninterpretable batch shows "Sign Transactions" + "3 transactions" and no asset rows; a single-tx `signTransaction` shows neither the plural title nor the subtitle.
- [ ] **Uninterpreted EVM** — a tx on a chain with `supports_simulations: false` renders *immediately* in the uninterpreted state rather than staying blank; the insufficient-funds warning still fires when the balance is short.
- [ ] **No regression on the happy path** — an interpreted EVM swap/approve still shows full details, the allowance **Edit** affordance, network fee, and custom gas.
- [ ] **Testnet toggle** refetches the Solana interpretation.

Checks: `type-check` ✓ · `lint` ✓ (0 errors) · `prettier` ✓ · `test:unit` 36/37 suites — the one failure (`estimateNetworkFee.test.ts`, hardcoded optimism fee numbers) is pre-existing on `main` and untouched here.

### DIFF
--- a/src/background/Wallet/Wallet.ts
+++ b/src/background/Wallet/Wallet.ts
@@ -1456,10 +1456,16 @@ export class Wallet {
     const keypair = this.getKeypairByAddress(currentAddress);
     const results = SolanaSigning.signAllTransactions(transactions, keypair);
 
-    results.forEach((result, index) => {
+    results.forEach((result) => {
       const contextParamsCopy = { ...transactionContextParams };
-      if (index > 0) {
-        /** TODO: Temporarily assume that addressAction describes only first tx */
+      if (results.length > 1) {
+        /**
+         * The addressAction the UI hands us interprets the batch as a whole
+         * (WLT-2189), so it doesn't belong to any single transaction in it:
+         * attaching it to one would make that transaction's pending history row
+         * and its analytics report the asset movements of all the others. Each
+         * transaction falls back to its own parse instead.
+         */
         contextParamsCopy.addressAction = null;
       }
       emitter.emit(
--- a/src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
+++ b/src/ui/components/address-action/TransactionSimulation/TransactionSimulation.tsx
@@ -177,7 +177,11 @@ export function TransactionSimulation({
               <TransactionAdvancedView
                 network={network}
                 interpretation={interpretation}
-                transaction={transaction}
+                transaction={
+                  transaction.evm
+                    ? { evm: transaction.evm }
+                    : { solana: [transaction.solana] }
+                }
                 addressAction={addressAction}
                 onCopyData={() => toastRef.current?.showToast()}
               />
--- a/src/ui/pages/SendTransaction/SendTransaction.tsx
+++ b/src/ui/pages/SendTransaction/SendTransaction.tsx
@@ -1,13 +1,11 @@
 import React, { useCallback, useMemo, useRef, useState } from 'react';
-import { hashQueryKey, useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation, useQuery } from '@tanstack/react-query';
 import { useNavigate, useSearchParams } from 'react-router-dom';
-import { Client } from 'defi-sdk';
 import type { CustomConfiguration } from '@zeriontech/transactions';
 import {
   getActionApproval,
   type AnyAddressAction,
 } from 'src/modules/ethereum/transactions/addressAction';
-import { incomingTxToIncomingAddressAction } from 'src/modules/ethereum/transactions/addressAction/creators';
 import type {
   IncomingTransaction,
   IncomingTransactionWithChainId,
@@ -72,7 +70,6 @@ import { uiGetBestKnownTransactionCount } from 'src/modules/ethereum/transaction
 import { resolveChainId } from 'src/modules/ethereum/transactions/resolveChainId';
 import { normalizeTransactionChainId } from 'src/modules/ethereum/transactions/normalizeTransactionChainId';
 import { usePreferences } from 'src/ui/features/preferences';
-import { useDefiSdkClient } from 'src/modules/defi-sdk/useDefiSdkClient';
 import {
   adjustedCheckEligibility,
   fetchAndAssignPaymaster,
@@ -86,11 +83,9 @@ import { valueToHex } from 'src/shared/units/valueToHex';
 import { useStaleTime } from 'src/ui/shared/useStaleTime';
 import type { useInterpretTxBasedOnEligibility } from 'src/ui/shared/requests/uiInterpretTransaction';
 import { interpretTxBasedOnEligibility } from 'src/ui/shared/requests/uiInterpretTransaction';
-import { solFromBase64 } from 'src/modules/solana/transactions/create';
 import type { SignTransactionResult } from 'src/shared/types/SignTransactionResult';
 import { whiteBackgroundKind } from 'src/ui/components/Background/Background';
 import type { StringBase64 } from 'src/shared/types/StringBase64';
-import { parseSolanaTransaction } from 'src/modules/solana/transactions/parseSolanaTransaction';
 import {
   hasCriticalWarning,
   InterpretationSecurityCheck,
@@ -243,51 +238,6 @@ function usePreparedTx(transaction: IncomingTransaction, origin: string) {
   };
 }
 
-function useLocalAction({
-  address: from,
-  transactionAction,
-  transaction,
-  networks,
-}: {
-  address: string;
-  transactionAction: TransactionAction;
-  transaction: IncomingTransactionWithChainId;
-  networks: Networks;
-}) {
-  const client = useDefiSdkClient();
-  const { currency } = useCurrency();
-  return useQuery({
-    queryKey: [
-      'incomingTxToIncomingAddressAction',
-      transaction,
-      transactionAction,
-      networks,
-      from,
-      client,
-      currency,
-    ],
-    queryKeyHashFn: (queryKey) => {
-      const key = queryKey.map((x) => (x instanceof Client ? x.url : x));
-      return hashQueryKey(key);
-    },
-    queryFn: () => {
-      return incomingTxToIncomingAddressAction(
-        { transaction: { ...transaction, from }, hash: '', timestamp: 0 },
-        transactionAction,
-        networks,
-        currency,
-        client
-      );
-    },
-    keepPreviousData: true,
-    useErrorBoundary: true,
-    retry: false,
-    refetchOnMount: false,
-    refetchOnReconnect: false,
-    refetchOnWindowFocus: false,
-  });
-}
-
 enum View {
   default = 'default',
   customAllowance = 'customAllowance',
@@ -316,7 +266,8 @@ function TransactionDefaultView({
   chain: Chain;
   origin: string;
   wallet: ExternallyOwnedAccount;
-  addressAction: AnyAddressAction;
+  /** `null` when the transaction could not be interpreted */
+  addressAction: AnyAddressAction | null;
   allowanceQuantityCommon: string | null;
   customAllowanceQuantityBase: string | null;
   interpretation: InterpretResponse | null | undefined;
@@ -367,7 +318,9 @@ function TransactionDefaultView({
             url={origin}
             alt={`Logo for ${origin}`}
           />
-          <UIText kind="headline/h2">{addressAction.type.displayValue}</UIText>
+          <UIText kind="headline/h2">
+            {addressAction?.type.displayValue ?? 'Sign Transaction'}
+          </UIText>
           <UIText kind="small/accent" color="var(--neutral-500)">
             {origin === INTERNAL_ORIGIN ? (
               'Zerion'
@@ -395,34 +348,36 @@ function TransactionDefaultView({
             </UIText>
           </HStack>
         </VStack>
-        <VStack
-          gap={4}
-          style={{
-            ['--surface-background-color' as string]: 'var(--neutral-100)',
-          }}
-        >
-          <AddressActionDetails
-            address={wallet.address}
-            addressAction={addressAction}
-            network={network}
-            allowanceQuantityCommon={allowanceQuantityCommon}
-            customAllowanceQuantityBase={customAllowanceQuantityBase}
-            showApplicationLine={true}
-            singleAssetElementEnd={
-              allowanceQuantityCommon &&
-              addressAction.type.value === 'approve' ? (
-                <UIText
-                  as={TextLink}
-                  kind="small/accent"
-                  style={{ color: 'var(--primary)' }}
-                  to={allowanceViewHref}
-                >
-                  Edit
-                </UIText>
-              ) : null
-            }
-          />
-        </VStack>
+        {addressAction ? (
+          <VStack
+            gap={4}
+            style={{
+              ['--surface-background-color' as string]: 'var(--neutral-100)',
+            }}
+          >
+            <AddressActionDetails
+              address={wallet.address}
+              addressAction={addressAction}
+              network={network}
+              allowanceQuantityCommon={allowanceQuantityCommon}
+              customAllowanceQuantityBase={customAllowanceQuantityBase}
+              showApplicationLine={true}
+              singleAssetElementEnd={
+                allowanceQuantityCommon &&
+                addressAction.type.value === 'approve' ? (
+                  <UIText
+                    as={TextLink}
+                    kind="small/accent"
+                    style={{ color: 'var(--primary)' }}
+                    to={allowanceViewHref}
+                  >
+                    Edit
+                  </UIText>
+                ) : null
+              }
+            />
+          </VStack>
+        ) : null}
         <HStack gap={8} style={{ gridTemplateColumns: '1fr 1fr' }}>
           <InterpretationSecurityCheck
             interpretation={interpretation}
@@ -546,14 +501,6 @@ function SendTransactionContent({
     chain,
   });
 
-  const { data: localAddressAction, ...localAddressActionQuery } =
-    useLocalAction({
-      address: singleAddress,
-      transactionAction,
-      transaction: populatedTransaction,
-      networks,
-    });
-
   const [allowanceQuantityBase, setAllowanceQuantityBase] = useState('');
 
   const configureTransactionToBeSigned = useEvent(
@@ -614,8 +561,6 @@ function SendTransactionContent({
   const paymasterWaiting =
     paymasterPossible && eligibilityQuery.isLoading && !isStale;
 
-  const client = useDefiSdkClient();
-
   const gasPricesReady = Boolean(chainGasPrices);
   const { isStale: isStaleGasPricesValue } = useStaleTime(gasPricesReady, 2000);
   /**
@@ -642,7 +587,6 @@ function SendTransactionContent({
       singleAddress,
       networks,
       transactionAction,
-      client,
       currency,
       source,
       eligibilityQuery.data?.data.eligible,
@@ -650,10 +594,6 @@ function SendTransactionContent({
       origin,
       wallet.address,
     ],
-    queryKeyHashFn: (queryKey) => {
-      const key = queryKey.map((x) => (x instanceof Client ? x.url : x));
-      return hashQueryKey(key);
-    },
     queryFn: async () => {
       const configuredTx = await configureTransactionToSign(
         populatedTransaction,
@@ -684,9 +624,24 @@ function SendTransactionContent({
     interpretQuery.data?.data.warnings
   );
 
-  const interpretAddressAction = interpretQuery.data?.data.action;
+  /**
+   * The interpretation is the _only_ source of the address action: we no longer
+   * derive one locally from the raw transaction, because parsing an untrusted
+   * dapp payload client-side produced confident-but-wrong summaries (WLT-2189).
+   */
+  const addressAction = interpretQuery.data?.data.action ?? null;
 
-  const addressAction = interpretAddressAction || localAddressAction || null;
+  /**
+   * Render nothing until we know what we're showing. `isInitialLoading` is
+   * `false` for a _disabled_ query, which is what we want for a network without
+   * simulation support (the request never runs, so we go straight to the
+   * uninterpreted state) but not while we're deliberately holding the request
+   * back waiting for gas prices — hence the extra check.
+   */
+  const interpretationPending = Boolean(
+    network?.supports_simulations &&
+      (isWaitingForGasrices || interpretQuery.isInitialLoading)
+  );
 
   const view = params.get('view') || View.default;
   const advancedDialogRef = useRef<HTMLDialogElementInterface | null>(null);
@@ -742,16 +697,7 @@ function SendTransactionContent({
     onSuccess: (tx) => handleSentTransaction(tx),
   });
 
-  if (localAddressActionQuery.isSuccess && !localAddressAction) {
-    throw new Error('Unexpected missing localAddressAction');
-  }
-
-  const maybeApproval = interpretQuery.data?.data.action
-    ? getActionApproval(interpretQuery.data.data.action)
-    : null;
-  const maybeLocalApproval = localAddressAction
-    ? getActionApproval(localAddressAction)
-    : null;
+  const maybeApproval = addressAction ? getActionApproval(addressAction) : null;
 
   const fungibleDecimals = maybeApproval?.fungible
     ? getDecimals({
@@ -762,10 +708,7 @@ function SendTransactionContent({
 
   const requestedAllowanceQuantityCommon =
     maybeApproval?.amount?.quantity ??
-    maybeLocalApproval?.amount?.quantity ??
-    (maybeApproval?.unlimited || maybeLocalApproval?.unlimited
-      ? UNLIMITED_APPROVAL_AMOUNT.toFixed()
-      : null);
+    (maybeApproval?.unlimited ? UNLIMITED_APPROVAL_AMOUNT.toFixed() : null);
 
   const requestedAllowanceQuantityBase =
     requestedAllowanceQuantityCommon && fungibleDecimals
@@ -780,7 +723,7 @@ function SendTransactionContent({
       ? baseToCommon(allowanceQuantityBase, fungibleDecimals).toFixed()
       : requestedAllowanceQuantityCommon;
 
-  if (!addressAction) {
+  if (interpretationPending) {
     return null;
   }
 
@@ -982,20 +925,23 @@ function EthSendTransaction() {
 
 function SolDefaultView({
   addressAction,
-  rawTransaction,
+  rawTransactions,
   txInterpretQuery,
   origin,
   wallet,
   networks,
 }: {
   origin: string;
-  addressAction: AnyAddressAction;
-  rawTransaction: StringBase64;
+  /** `null` when the transaction could not be interpreted */
+  addressAction: AnyAddressAction | null;
+  /** The whole batch for `signAllTransactions`, a single-item list otherwise */
+  rawTransactions: StringBase64[];
   txInterpretQuery: ReturnType<typeof useInterpretTxBasedOnEligibility>;
   wallet: ExternallyOwnedAccount;
   networks: Networks;
 }) {
   const originForHref = useMemo(() => prepareForHref(origin), [origin]);
+  const isBatch = rawTransactions.length > 1;
 
   const advancedDialogRef = useRef<HTMLDialogElementInterface | null>(null);
 
@@ -1026,7 +972,15 @@ function SolDefaultView({
             url={origin}
             alt={`Logo for ${origin}`}
           />
-          <UIText kind="headline/h2">{addressAction.type.displayValue}</UIText>
+          <UIText kind="headline/h2">
+            {addressAction?.type.displayValue ??
+              (isBatch ? 'Sign Transactions' : 'Sign Transaction')}
+          </UIText>
+          {isBatch ? (
+            <UIText kind="small/regular" color="var(--neutral-500)">
+              {`${rawTransactions.length} transactions`}
+            </UIText>
+          ) : null}
           <UIText kind="small/accent" color="var(--neutral-500)">
             {origin === INTERNAL_ORIGIN ? (
               'Zerion'
@@ -1054,22 +1008,24 @@ function SolDefaultView({
             </UIText>
           </HStack>
         </VStack>
-        <VStack
-          gap={4}
-          style={{
-            ['--surface-background-color' as string]: 'var(--neutral-100)',
-          }}
-        >
-          <AddressActionDetails
-            address={wallet.address}
-            addressAction={addressAction}
-            network={network}
-            showApplicationLine={false}
-            allowanceQuantityCommon={null}
-            customAllowanceQuantityBase={null}
-            singleAssetElementEnd={null}
-          />
-        </VStack>
+        {addressAction ? (
+          <VStack
+            gap={4}
+            style={{
+              ['--surface-background-color' as string]: 'var(--neutral-100)',
+            }}
+          >
+            <AddressActionDetails
+              address={wallet.address}
+              addressAction={addressAction}
+              network={network}
+              showApplicationLine={false}
+              allowanceQuantityCommon={null}
+              customAllowanceQuantityBase={null}
+              singleAssetElementEnd={null}
+            />
+          </VStack>
+        ) : null}
         <HStack gap={8} style={{ gridTemplateColumns: '1fr 1fr' }}>
           <InterpretationSecurityCheck
             interpretation={txInterpretQuery.data}
@@ -1102,7 +1058,7 @@ function SolDefaultView({
         <RenderArea name="transaction-warning-section" />
       </VStack>
       <div style={{ marginTop: 'auto' }}>
-        {addressAction.fee ? (
+        {addressAction?.fee ? (
           <AddressActionNetworkFee
             fee={addressAction.fee}
             isLoading={txInterpretQuery.isLoading}
@@ -1122,7 +1078,7 @@ function SolDefaultView({
             <TransactionAdvancedView
               network={network}
               interpretation={txInterpretQuery.data}
-              transaction={{ solana: rawTransaction }}
+              transaction={{ solana: rawTransactions }}
               addressAction={addressAction}
               onCopyData={() => toastRef.current?.showToast()}
             />
@@ -1156,6 +1112,13 @@ function assertKnownSolanaMethodParam(
   );
 }
 
+/**
+ * `transactions` always holds the _whole_ payload — the full batch for
+ * `signAllTransactions`, a single-item list for the other methods — so that
+ * interpretation and the details view never have to reach for "the first one".
+ * `transaction` stays on the single-transaction methods only, because those are
+ * the ones whose signing call takes exactly one transaction.
+ */
 function normalizeTxParams(params: URLSearchParams):
   | {
       method: 'signAllTransactions';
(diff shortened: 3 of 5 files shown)


zerts:

**Result:** `pass`

### Screenshots

**Solana signAllTransactions (3 txs) — interpreted, all three transfers + "3 transactions" subtitle**

Solana signAllTransactions (3 txs) — interpreted, all three transfers + "3 transactions" subtitle

**Details dialog lists every transaction in the batch ("Transaction 1/2/3 of 3")**

Details dialog lists every transaction in the batch ("Transaction 1/2/3 of 3")

**Solana batch with no interpretation — "Sign Transactions" + "3 transactions" + Unverified**

Solana batch with no interpretation — "Sign Transactions" + "3 transactions" + Unverified

**Solana signTransaction (single) — "Sign Transaction", no batch subtitle**

Solana signTransaction (single) — "Sign Transaction", no batch subtitle

**EVM send — interpreted action, network fee and nonce intact**

EVM send — interpreted action, network fee and nonce intact

**EVM with interpretation forced to fail — uninterpreted layout, fee + nonce still render**

EVM with interpretation forced to fail — uninterpreted layout, fee + nonce still render

## Summary

Persona "default" (`$DATA/qa/extension/default`), built and loaded `dist/` from the PR head
`d52326aa` as an unpacked extension. Drove the `#/sendTransaction` confirmation route directly
for both ecosystems. It boots, unlocks, and every affected state renders with real backend data.

The central claim of the ticket verified at the network level: with the wallet switched to the
persona's Solana address `2Nsnn…Mq1Xu`, a `signAllTransactions` request for **3** transactions
produced exactly one POST to `https://zpi.zerion.io/wallet/simulate-transactions/v1` whose body
carried **all 3** (`chain: solana`, `domain: https://jup.ag`) — not the first one. The
single-transaction methods still send exactly 1. The interpreted screen then rendered all three
transfers (−0.001 / −0.002 / −0.003 SOL with USD values), which is the user-visible proof that
the whole batch is being summarised rather than `transactions[0]`.

Both halves of the "no interpretation" path were exercised deliberately by fulfilling
`**/simulate-transactions/v1` with a 500: the EVM screen fell back to "Sign Transaction" with no
asset rows while **keeping the network fee and nonce**, and the Solana batch fell back to
"Sign Transactions" + "3 transactions". Those confirm the removal of the local-action fallback
degrades honestly instead of blanking the popup.

## Verification log

- **Profile:** zerion-wallet-extension — persona "default"; run 1 switched the active wallet to
  the Solana account via `#/wallet-select`, run 2 stayed on the default EVM account.
- **Built:** `npm run build:chrome` → `dist/manifest.json` present (no dev server for MV3).
- **Fixtures:** 3 real SOL transfers minted with `@solana/web3.js` from the persona's Solana
  address against a live mainnet blockhash (`H4gd2vMG…XorBQY`), amounts 0.001 / 0.002 / 0.003 SOL
  to three distinct recipients — so the aggregated action is distinguishable from a single-tx one.
- **Flows driven:**
  - `#/sendTransaction?ecosystem=solana&method=signAllTransactions&transactions=[3]` → interpreted
    "Send" + "3 transactions" + three transfer rows + Network Fee → Details → "Transaction n of 3" ×3
  - same route with interpretation forced to 500 → "Sign Transactions" + "3 transactions" + Unverified
  - `#/sendTransaction?ecosystem=solana&method=signTransaction&transaction=…` → "Sign Transaction",
    no subtitle, Details shows the raw payload with **no** "Transaction 1 of 1" label
  - `#/sendTransaction?transaction={evm}` (mainnet, 0.0001 ETH) → interpreted "Send", fee, nonce 247
  - same with interpretation forced to 500 → uninterpreted layout, fee `~30 sec · $0.10`, nonce 247
- **Request assertion:** batch → `transactions.length === 3`; single → `1`. Captured via
  `ctx.route` on `**/simulate-transactions/v1`.
- **Console/network:** clean. The only errors were the 500s I injected myself, plus one `404` from
  `simulate-transactions/v1` declining to interpret a single synthetic SOL transfer — which
  correctly routed to the uninterpreted state rather than erroring. Zero non-noise request failures.
- **Not a finding (pre-existing, unrelated):** my first EVM fixture asked to send 0.01 ETH from a
  wallet holding ~0.00095 ETH; gas estimation threw and `usePreparedTx`'s `resolveGasQuery`
  (`useErrorBoundary: true`) surfaced the raw RPC "insufficient funds" text on a full-page error
  boundary. That path predates this PR and is upstream of anything the diff touches. Re-ran with
  an affordable 0.0001 ETH value, which is what screenshot 05 shows.
- **Out of scope for this gate:** hardware-wallet `signAllTransactions` (still throws by design),
  and design-match of the new subtitle — that is `/loop-review`'s call.

zerts:
## 🔄 Loop review — iteration 1/5

**Verdict:** `needs-changes`

| Severity | Area | Finding |
| --- | --- | --- |
| major | behavior | Confirmation popup stays blank for ~8s when the interpretation request fails (was 0.3s on main) (`src/ui/pages/SendTransaction/SendTransaction.tsx:641`) |
| minor | code | Batch-wide action is still attributed to transactions[0]'s local record / analytics (`src/background/Wallet/Wallet.ts:1462`) |

## Summary

Reviewed `d52326aa` (PR #963, `extension-support-batched-tx-in-solana-wlt-2189`) against
WLT-2189 and the PRD comment on the issue. There are no human review threads on the PR, so
the Linear description plus the author's own PRD comment are the spec.

**Every functional spec point is met and verified live.** A `signAllTransactions` request for
3 transactions produces exactly one `POST wallet/simulate-transactions/v1` carrying all
**3** (`chain: solana`, address `2Nsnnh…`), and the screen renders all three transfers
(−0.001 / −0.002 / −0.003 SOL) — the aggregated batch, not `transactions[0]`. Details lists
"Transaction 1/2/3 of 3", a single-tx request shows neither the plural title nor the
"N transactions" subtitle nor a "Transaction 1 of 1" label, `firstTx` is gone
(`grep`: no hits), the local action fallbacks are gone from this file while
`TransactionSimulation`/`creators.ts` keep theirs as scoped, and the EVM happy path
(interpreted action, fee, nonce, allowance affordance) is intact. `type-check` clean,
`test:unit` and `test:e2e` show only failures that reproduce on `origin/main`.

**What doesn't work is the timing of the uninterpreted state.** Because the render gate now
waits on `interpretQuery.isInitialLoading` and the interpret queries inherit react-query's
default `retry: 3`, an interpretation that fails leaves the *signing confirmation* completely
blank for ~8 seconds before the honest uninterpreted layout appears. I measured the same
scenario on `origin/main`: 315 ms. That is the one thing I'd fix before merging; the rest of
the change is good.

## Findings

### [major/behavior] Confirmation popup stays blank for ~8s when the interpretation request fails

`src/ui/pages/SendTransaction/SendTransaction.tsx:641` (EVM `interpretationPending` → early
`return null` at :726) and `:1280` (Solana `!interpretQuery.isInitialLoading`).

The interpret queries (`:574`, `:1205`) don't set `retry`, so they inherit react-query's
default of 3 retries with 1 s/2 s/4 s backoff — 4 attempts, ≈7.7 s — and `isInitialLoading`
stays `true` for the whole cycle. Removing the local-action fallback removed the thing that
used to fill that window.

Repro (`shots2/`, `shots3/`, driver `drive2.mjs` / `drive3.mjs` in the scratch dir):
force the endpoint to fail with
`ctx.route('**/wallet/simulate-transactions/v1', r => r.fulfill({ status: 500, … }))`, then
open `#/sendTransaction`.

| build | scenario | first non-empty render |
|---|---|---|
| PR head `d52326aa` | EVM `#/sendTransaction`, forced 500 | **8157 ms** (two independent runs: 8157 / 8249 ms) |
| PR head `d52326aa` | Solana `signAllTransactions` ×3, forced 500 | **7159 ms** |
| `origin/main` | EVM `#/sendTransaction`, forced 500 | **315 ms** |

Evidence: `shots2/evm-fail-8000ms.png` catches the screen mid-appearance at t=8 s;
`shots/02-evm-uninterpreted.png` is a fully white 440×900 popup taken 13 s after navigation
(`document.body.innerText === ''`). The Solana path is marginally better only because its
footer renders outside the gate — for 7 s the popup is a Cancel/Confirm pair with an empty
body, i.e. a signable transaction with nothing shown about it.

Aggravating detail, observed once (run 1): after the query settles into `error` it can go
*back* to `isInitialLoading` and blank again. `keepPreviousData` has no data to keep after an
error, so any change to the query key (`chainGasPrices` refresh, the `configuration` object
`TransactionConfiguration` pushes up, eligibility settling) starts a fresh 4-attempt cycle.
Request timeline from that run: attempts at +6.3 s/+7.5 s/+9.6 s/+13.9 s (cycle 1, content
appears at 14.2 s), then +14.3 s/+15.4 s/+17.6 s (cycle 2, screen blank again at 19 s). In the
isolated 46 s run it did not recur, so this is timing-dependent, not guaranteed.

Expected: the PRD's "blank only while the interpretation is genuinely in flight" — i.e. the
first failed attempt should drop straight to the uninterpreted state. `retry: 0` (or `1`) on
the two interpret queries, or treating `isError`/`failureCount > 0` as settled, gets that
without reintroducing any local parsing. Note this is not limited to injected 5xx: the
endpoint 404s on payloads it declines to interpret, and `interpretTransactions`
(`src/ui/shared/requests/interpret.ts:31-44`) throws its own invariants for a tx with no
`to`/`nonce`/`gas` — every one of those paths now costs the user 8 blank seconds. A 200 with
`action: null` is fine and renders immediately, which is the common uninterpretable case.

### [minor/code] Batch-wide action is still attributed to transactions[0]'s local record / analytics

`src/background/Wallet/Wallet.ts:1458-1468` attaches `transactionContextParams.addressAction`
to the first signed transaction of a `signAllTransactions` batch and nulls it for the rest,
with the comment `/** TODO: Temporarily assume that addressAction describes only first tx */`.
That assumption is now inverted: the action the UI hands down describes the *whole* batch, so
the first transaction's local record gets the batch's aggregated asset movements.
`pendingSolanaTxToAddressAction` (`addressAction/creators.ts:285-299`) prefers the stored
action, so a pending History row for tx 1 of a 3-tx batch would show all three transfers while
txs 2–3 fall back to their own per-tx parse. Same over-reporting flows into
`addressActionToAnalytics`. Outside the diff's files and transient (backend history replaces
these rows), so it doesn't block — but the TODO now says the opposite of what the code does.

## Verification log

- **Tests**
  - `type-check` (`tsc --noEmit`): clean.
  - `test:unit`: 361/364 passed. `estimateNetworkFee.test.ts › feeEstimation for optimism`
    fails on drifted hardcoded fee numbers — pre-existing, unrelated to the diff, and called
    out in the PR body. `WalletRecord.test.ts` timed out twice in the full run only because I
    had `tsc` and a parcel build running concurrently; re-run alone with the repo's own
    invocation (`node --experimental-vm-modules node_modules/jest/bin/jest.js …`) it is
    **6/6 passed in 45 s**. Effectively 36/37 suites, 1 pre-existing failure.
  - `test:e2e` (`npx playwright test`, Google Chrome channel, dist built from the PR head):
    **20 passed, 1 failed** — `overview.spec.ts › Matches snapshot`, a stale aria snapshot of
    the Overview nav (Rewards/Perks/Premium vs Perps/Stats). Reproduced on a dist built from
    `origin/main` in `/tmp/loop-review-main`, so pre-existing.
- **Live** — MV3 extension, no dev server: `npm run build:chrome` from the PR head, loaded
  `dist/` unpacked into `chromium --headless=new` via `launchPersistentContext`, persona
  "default" seeded into `chrome.storage.local` and unlocked (443×900 popup viewport).
  Screenshots + drivers + raw JSON results in **`/tmp/loop-review-aXvpVr`**
  (`shots/`, `shots2/`, `shots3/`, `results*.json`, `drive*.mjs`).
  - Solana fixtures: 3 real `SystemProgram.transfer` txs minted for the persona's Solana
    address against a live mainnet blockhash, distinct amounts (0.001/0.002/0.003 SOL) so an
    aggregated batch action is distinguishable from a single-tx one.
  - `signAllTransactions` ×3, interpreted (`shots3/02`, `03`): title "Send", subtitle
    "3 transactions", three transfer rows, No Risks Found, Network Fee $0.001; Details →
    "Transaction 1 of 3 | Transaction 2 of 3 | Transaction 3 of 3", each in its own Surface.
    Intercepted request body: `transactions.length === 3`, `chain: solana`, `kinds: [solana ×3]`.
  - `signAllTransactions` ×3, interpretation forced to 500 (`shots3/04`, `05`): "Sign
    Transactions" + "3 transactions" + Unverified + "unable to simulate…" banner, **no asset
    rows**, existing two-column layout intact; Details still lists all 3.
  - `signTransaction` single (`shots3/06`, `07`): title "Send", **no** subtitle, request body
    `transactions.length === 1`, Details shows the bare payload with **no** "Transaction 1 of 1".
  - EVM interpreted (`shots/01`): "Send", recipient, −0.0001 ETH, Network Fee `~30 sec · $0.07`,
    Nonce 247 — happy path unaffected.
  - EVM uninterpreted (`shots2/evm-fail-12000ms.png`): "Sign Transaction", no asset rows,
    Unverified + banner, **fee and nonce still render**.
  - Wallet switched to the Solana account via `#/wallet-select` before the Solana runs
    (`shots3/00`, `01`).
  - Not driven live, verified by reading the code instead: (1) *Copy Raw Data* copying the full
    JSON array — headless denied `navigator.clipboard.readText`; `transactionFormatted`
    (`TransactionAdvancedView.tsx:202-210`) stringifies the array for N>1 and keeps the bare
    string for N=1. (2) The insufficient-funds warning under a null action — the only route to
    it via `#/sendTransaction` is an unaffordable value, which hard-throws in
    `usePreparedTx`'s `resolveGasQuery` (`useErrorBoundary: true`) to a full-page error
    boundary that predates this PR; statically, `TransactionWarnings` renders
    `InsufficientFundsWarning` unconditionally and only gates the "may fail" branch on
    `addressAction?.status`. (3) The `TransactionSimulation.tsx` call-site change — it feeds
    the *old* Send/Swap/Bridge confirmation views, whose only affordance past the Details
    dialog is a real Confirm that signs and broadcasts, so I did not drive it; the change is a
    mechanical `{ solana: [x] }` wrap that renders identically for a 1-item list (no label) and
    type-checks.
- **Console/network**: clean apart from the 500s I injected. One 404 (dapp favicon for the
  fake origin) and `event-collector.zerion.io … ERR_ABORTED` (analytics beacon on unload) —
  both noise. No `zpi.zerion.io` failures I didn't cause. Zero React errors, no error
  boundaries in any of the reviewed flows.
- **Animation**: no animation code in diff.

zerts:
## ✅ QA smoke gate — iteration 2/5

**Result:** `pass`

### Screenshots

**Popup boots and unlocks — persona portfolio loads**

Popup boots and unlocks — persona portfolio loads

**EVM, interpretation forced to fail — uninterpreted layout, fee and nonce intact**

EVM, interpretation forced to fail — uninterpreted layout, fee and nonce intact

**EVM, interpretation held 5s — default view is up at 1.5s with "Simulating..."**

EVM, interpretation held 5s — default view is up at 1.5s with "Simulating..."

**EVM happy path — Send, recipient, −0.0001 ETH, fee, nonce**

EVM happy path — Send, recipient, −0.0001 ETH, fee, nonce

**Solana signAllTransactions x3, interpretation forced to fail — plural title + "3 transactions"**

Solana signAllTransactions x3, interpretation forced to fail — plural title + "3 transactions"

**Solana signAllTransactions x3 interpreted — all three transfers, not just the first**

Solana signAllTransactions x3 interpreted — all three transfers, not just the first

**Details dialog lists Transaction 1/2/3 of 3 with each raw payload**

Details dialog lists Transaction 1/2/3 of 3 with each raw payload

**Solana signTransaction — no batch subtitle, single transfer row**

Solana signTransaction — no batch subtitle, single transfer row

## Summary

Persona "default" (`ai-runner-data/qa/extension/default`), MV3 extension built from the PR
head `465065a1` and loaded unpacked into headless Chromium. The popup boots, unlocks, and
loads the persona's real portfolio ($712.55 across 13+ chains). Drove `#/sendTransaction`
for both ecosystems across all three interpretation outcomes — resolved, in flight, and
failed — plus the Details dialog. Everything renders; nothing blanks, throws, or sticks in
a skeleton.

The two behaviours this iteration changed both hold up. **The screen is never blank:**
first paint is 40–208 ms in every scenario, including the one the previous review measured
at 8157 ms (EVM with the interpretation failing from a cold cache), and no navigation ever
went back to blank after painting (`reBlankedAtMs: null`, sampled every 150 ms for 12–14 s).
**The batch is interpreted as a whole:** a 3-transaction `signAllTransactions` request sends
one `POST wallet/simulate-transactions/v1` carrying all 3, and the screen shows all three
transfers (−0.001 / −0.002 / −0.003 SOL) under a "3 transactions" subtitle.

## Verification log

- **Profile**: zerion-wallet-extension — persona "default", no flag overrides needed
  (`#/sendTransaction` is route-drivable for both ecosystems). Active wallet switched to
  the persona's Solana address `2Nsnn…Mq1Xu` via `#/wallet-select` for the Solana runs.
- **Built**: `npm run build:chrome` from the worktree at `465065a1` (clean tree,
  `git reset --hard origin/…` before building); `dist/manifest.json` present. No dev
  server — MV3 extension, per profile.
- **Fixtures**: three real `SystemProgram.transfer` transactions minted for the persona's
  Solana address against a live mainnet blockhash, with distinct amounts so an aggregated
  batch action is distinguishable from a single-transaction one at a glance. EVM fixture
  sends 0.0001 ETH (the persona holds ~0.00095 — a larger value hits the pre-existing
  gas-estimation error boundary, see profile).
- **Interpretation control**: `ctx.route('**/wallet/simulate-transactions/v1')` with a
  three-way switch — `live` (pass through), `fail` (inject 500), `slow` (hold 5 s, then
  pass through). The `slow` mode is what makes the in-flight screen screenshottable; it
  otherwise resolves in ~500 ms.

- **Flows and time to first paint** (sampled every 150 ms from `goto`; `reBlanked` = did the
  screen ever go blank again after painting):

  | flow | first paint | reBlanked |
  |---|---|---|
  | EVM, interpretation fails from first attempt, cold cache | 208 ms | null |
  | EVM, interpretation resolves | 47 ms | null |
  | Solana batch ×3, interpretation fails | 40 ms | null |
  | Solana batch ×3, interpretation resolves | 60 ms | null |
  | Solana single, interpretation resolves | 60 ms | null |

  For reference, the previous review measured the first row at **8157 ms** on `d52326aa`
  and **315 ms** on `origin/main`.

- **In-flight state** (interpretation held 5 s, EVM): at 1.5 s the popup shows
  "Sign Transaction" + a shimmering "Simulating..." badge + Details + Network Fee
  `~30 sec · $0.07` + Nonce 247 + Cancel/Confirm — the full default view, no blank, no
  skeleton. At 8.5 s it has filled in to "Send" + recipient + −0.0001 ETH + "No Risks
  Found", fee and nonce unchanged.

- **Batch interpretation, asserted on the wire**: `signAllTransactions` ×3 →
  `transactions.length === 3`, `kinds: [solana, solana, solana]`,
  `address: 2NsnnhZukXq9hqXSTLv46XqtLP6F3hmbUMdasgdMq1Xu`, `domain: https://jup.ag`.
  `signTransaction` → `transactions.length === 1`. EVM → `length === 1`, `kinds: [evm]`.

- **Rendered content**:
  - Batch interpreted: "Send" + "3 transactions" + three rows (−0.001 SOL $0.09148,
    −0.002 SOL $0.183, −0.003 SOL $0.2744) + No Risks Found + Network Fee $0.001.
  - Batch Details: "Transaction 1 of 3 / 2 of 3 / 3 of 3", each raw payload in its own
    surface, Copy Raw Data.
  - Batch uninterpreted: "Sign Transactions" (plural) + "3 transactions" + Unverified +
    the "unable to simulate" banner, no asset rows.
  - Single: "Send", one row, **no** batch subtitle; uninterpreted EVM: "Sign Transaction"
    (singular) with fee and nonce still rendering.

- **Console/network**: clean. The only console errors were the 500s I injected myself
  (4 per navigation — react-query retries) plus one generic 404 (favicon for the fake
  `jup.ag` origin). Zero request failures against `zpi.zerion.io`. No React errors, no
  error boundaries.

- **Note on the run itself, not a finding**: the first end-to-end driver pass had the
  browser context close on it mid-run ("Target page, context or browser has been closed")
  between the Solana batch-uninterpreted screenshot and the next navigation. Re-ran the
  same sequence twice with `page.on('crash')`/`close` listeners attached — both completed
  cleanly with an empty lifecycle log, so it was a harness flake, not an extension crash.
  Every scenario above comes from a completed run.

zerts:
## ✅ Loop review — iteration 2/5

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | behavior | Confirmation now renders before the interpretation settles, contradicting the PRD's "blank until settled" decision and the PR description (`src/ui/pages/SendTransaction/SendTransaction.tsx:628`) |
| minor | visual | Solana batch footer shifts down 65px when the interpretation lands (main never shifted) (`src/ui/pages/SendTransaction/SendTransaction.tsx:999`) |

## Summary

Reviewed `465065a1` (PR #963, `extension-support-batched-tx-in-solana-wlt-2189`) against WLT-2189
and the author's PRD comment on the issue. There are still no human review threads or PR-level
human comments — the Linear description plus the PRD comment are the spec. This head adds one
commit on top of the previously reviewed `d52326aa`: it drops the render gate and stops attaching
the batch-wide address action to `transactions[0]`.

**Both defects from the previous iteration are fixed, and the ticket's own goal verifies cleanly
against a baseline build of `origin/main`.** With the interpretation forced to fail from a cold
document, the EVM confirmation now paints in **827 ms** (was 8157 ms on `d52326aa`; `origin/main`
paints at 1953 ms with a locally-parsed summary), the Solana batch in **969 ms**, and no scenario
ever went back to blank after painting. The in-flight screen is the honest uninterpreted layout
plus a "Simulating..." badge, with network fee and nonce intact. On the batch side, the same
3-transaction `signAllTransactions` request that `origin/main` sends to the interpretation
endpoint as **`n=1`** (one row, one raw payload in Details, only `transactions[0]`'s recipient) is
sent by this head as **`n=3`** and rendered as three transfer rows under a "3 transactions"
subtitle, with "Transaction 1/2/3 of 3" in Details. `Wallet.ts` now nulls the batch action for
every transaction in a batch, so no single pending history row inherits the whole batch's asset
movements — the previous iteration's minor finding, with an accurate comment replacing the
inverted TODO.

`type-check` clean, `eslint` clean on the touched paths, `test:unit` 363/364 with one
pre-existing failure, `test:e2e` 20/21 with one failure reproduced on `origin/main`. What's left
is two minor notes: the code now does the opposite of a documented PRD decision (for good
reasons, but the written record should catch up), and the Solana batch footer moves once when the
interpretation arrives.

## Findings

### [minor/behavior] Confirmation now renders before the interpretation settles, contradicting the PRD's "blank until settled" decision and the PR description

`src/ui/pages/SendTransaction/SendTransaction.tsx:628-636` (EVM) and `:1221-1228` (Solana).
`465065a1` removes `interpretationPending` and the `!interpretQuery.isInitialLoading` gate, so the
view renders immediately with `addressAction === null` and fills in when the interpretation lands.

The PRD's Decisions section says the opposite, explicitly and with a behaviour table:

> **Loading = blank until settled.** Nothing renders while the interpretation is genuinely in
> flight — no flash of the uninterpreted state for a transaction that is about to resolve fine.
>
> | state | render |
> | request pending (or waiting on gas prices, <2s) | blank |

The PR description carries the same claim ("**Blank only while the interpretation is genuinely in
flight**", including the justification for the `isWaitingForGasrices` check that no longer gates
anything — `isWaitingForGasrices` now only feeds the query's `enabled`).

I am **not** asking for the gate back: it is what produced the 8-second blank signing popup, and
the flash it was there to avoid is small and honest (measured below). But the code and the spec
now disagree, so the PRD decision and that PR-body bullet should be updated to describe what
ships, otherwise the next reader reviews against a rule the code deliberately broke.

Measured cold-document behaviour, PR head vs a build of `origin/main`
(`res-h-evm.json` / `res-main-evm.json` / `res-h-sol.json` / `res-main-sol.json`, drivers
`drive-i3.mjs`, screenshots in `shots-h-*` and `shots-main-*`):

| scenario | head: first paint | head: badge sequence | main: first paint |
|---|---|---|---|
| EVM, interpretation resolves | 435 ms | No Risks Found (153 ms) → Simulating… → interpreted @1261 ms | 1846 ms |
| EVM, interpretation held 5 s | 1210 ms | No Risks Found (185 ms) → Simulating… → interpreted @7355 ms | 851 ms |
| EVM, interpretation fails | 827 ms | Simulating… → Unverified @8452 ms | 1953 ms |
| Solana batch ×3, resolves | 1569 ms | Simulating… → interpreted @2029 ms | 378 ms |
| Solana batch ×3, held 5 s | 1048 ms | Simulating… → interpreted @6524 ms | 405 ms |
| Solana batch ×2, fails | 969 ms | Simulating… → Unverified @8120 ms | 383 ms |

`reBlankedAtMs` was `null` in every scenario on both builds (sampled every 150 ms for 12–16 s), so
there is no re-blank after the first paint — the aggravating case from the previous review is gone.

Two side observations that come with rendering early, neither of them regressions (both reproduce
identically on `origin/main`, which is why they are folded in here rather than filed separately):

- The EVM security badge shows a **green "No Risks Found"** for ~150–185 ms before flipping to
  "Simulating...", because the interpret query is disabled while `isWaitingForGasrices` is true and
  `InterpretationSecurityCheck` maps "not loading, not error" to `success`
  (`src/ui/shared/security-check/InterpertationSecurityCheck.tsx:186-192`). Main: 153 ms in the same
  scenario. The 'ok' emit also leaves the page's security gradient green for the rest of the
  simulation (`SecurityStatusBackground` only re-emits for non-loading kinds). The same mapping is
  why a chain with `supports_simulations: false` shows a permanent green badge — pre-existing,
  out of this diff.
- The interpretation still takes ~8 s to give up (react-query's default `retry: 3`, attempts at
  ~0.9/2.1/4.2/8.4 s), and the popup is signable from ~0.8 s, so a user can confirm while the
  security check is still pending. Main is the same (~10 s). `retry: 1` on the two interpret
  queries would cut that in half and would also make a short "blank until settled" gate affordable
  again if the PRD behaviour is still wanted.

### [minor/visual] Solana batch footer shifts down 65px when the interpretation lands

`src/ui/pages/SendTransaction/SendTransaction.tsx:999` — the Solana path renders no
`AddressActionDetails` block and no `AddressActionNetworkFee` while `addressAction` is null, so
the batch screen grows by three asset rows plus the fee line when the interpretation arrives, and
the Cancel/Confirm row moves from `y=532` to `y=597`. In the held-interpretation run that happens
5.5 s after the screen becomes interactive (`shots-h-sol/sol-batch-slow-1500ms.png` vs
`sol-batch-slow-final.png`).

`origin/main` never moved that footer (`y=532` throughout every Solana scenario) because the local
parse gave it one asset row from the first frame; the EVM path on this head is stable to 1 px
(`y=532 → 533`) because fee and nonce hold the space. The shift is downward, so a click aimed at
Confirm lands on empty space rather than on another control — no click-hijack, just a jump. Worth
a reserved-height placeholder for the action block if it bothers you; it does not block.

## Verification log

- **Tests** (from the worktree at `465065a1`, `dist` rebuilt from scratch):
  - `type-check` (`tsc --noEmit`): clean.
  - `npx eslint src/ui/pages/SendTransaction src/background/Wallet/Wallet.ts src/ui/components/address-action/TransactionSimulation`: clean.
  - `test:unit`: **36/37 suites, 363/364 tests**. The single failure,
    `estimateNetworkFee.test.ts › feeEstimation for optimism`, is drifted hardcoded fee numbers in
    a module the diff does not touch — pre-existing and called out in the PR body.
  - `test:e2e` (`playwright test`, Google Chrome channel): **20 passed, 1 failed** —
    `overview.spec.ts › Matches snapshot`, a stale aria snapshot of the Overview nav. Re-ran that
    spec against the `origin/main` build in `/tmp/loop-review-main` this iteration: **fails
    identically**, so pre-existing.
- **Live** — MV3 extension, no dev server: `npm run build:chrome`, `dist/` loaded unpacked into
  `chromium --headless=new` via `launchPersistentContext`, persona "default" seeded into
  `chrome.storage.local` and unlocked, 440×900 popup viewport. Every scenario loads a **fresh
  document** (`about:blank` first) with a **distinct transaction**, because the real confirmation
  popup is always a new window with an empty react-query cache — an earlier pass of this review
  measured hash-navigations instead and got cache-warm numbers, which is why the numbers above
  come from `drive-i3.mjs` only. Artifacts in **`/tmp/loop-review-MIO3Ec`**
  (`drive-i3.mjs`, `res-h-evm.json`, `res-h-sol.json`, `res-main-evm.json`, `res-main-sol.json`,
  `shots-h-evm/`, `shots-h-sol/`, `shots-main-evm/`, `shots-main-sol/`).
  - Fixtures: three fresh `SystemProgram.transfer` transactions minted for the persona's Solana
    address (`2Nsnn…Mq1Xu`) against a live mainnet blockhash, distinct amounts, three distinct
    recipients. EVM: 0.0001 ETH on mainnet, value varied per scenario.
  - Interpretation control: `ctx.route('**/wallet/simulate-transactions/v1')` with `live` /
    `fail` (inject 500) / `slow` (hold 5 s then pass through).
  - **Batch is interpreted as a whole, asserted on the wire**: `signAllTransactions` ×3 →
    one POST, `transactions.length === 3`, `kinds: [solana, solana, solana]`, `chain: solana`,
    `address: 2Nsnnh…Mq1Xu`, `domain: https://jup.ag`. Same request on the `origin/main` build:
    `transactions.length === 1`. `signTransaction` → 1. EVM → 1, `kinds: [evm]`.
  - Batch interpreted (`shots-h-sol/sol-batch-live-final.png`): "Send" + "3 transactions" +
    three rows (−0.001 / −0.002 / −0.003 SOL with USD values) + No Risks Found + Network Fee
    $0.001. On main the same request rendered one row.
  - Batch Details (`sol-batch-live-details.png`): "Transaction 1 of 3 / 2 of 3 / 3 of 3", each raw
    payload in its own Surface, Copy Raw Data. On main: a single unlabelled payload.
  - Batch uninterpreted (`sol-batch-fail-final.png`): "Sign Transactions" + "2 transactions" +
    Unverified + the "unable to simulate" banner, no asset rows, no fee line; Details still lists
    both transactions with labels.
  - Batch in flight (`sol-batch-slow-1500ms.png`): "Sign Transactions" + "3 transactions" +
    "Simulating..." + Details + Cancel/Confirm.
  - Single tx (`sol-single-live-final.png`, `sol-single-live-details.png`): "Send", **no**
    subtitle, one row, and Details shows the bare payload with **no** "Transaction 1 of 1" label
    (`labels: ""`).
  - EVM interpreted (`evm-live-final.png`): "Send", recipient, −0.0001 ETH, Network Fee
    `~30 sec · $0.07`, Nonce 247 — happy path unaffected.
  - EVM in flight (`evm-slow-1500ms.png`): "Sign Transaction" + "Simulating..." + Details +
    fee + nonce + Cancel/Confirm.
  - EVM uninterpreted (`evm-fail-final.png`, `evm-fail-details`): "Sign Transaction" + Unverified
    + banner, **fee and nonce still render**, and the Details dialog renders the raw tx fields with
    a null action without throwing.
  - Wallet switched to the persona's Solana account via `#/wallet-select` before the Solana runs
    (`sol-00-wallet-select.png`, `sol-01-after-switch.png`).
  - Not driven live, code-verified instead: (1) `solana_signAllTransactions` context params
    (`Wallet.ts:1459-1470`) — reaching them requires actually signing and broadcasting, which this
    review does not do; statically, `results.length > 1` nulls `addressAction` for **every**
    transaction in a batch and `pendingSolanaTxToAddressAction`
    (`addressAction/creators.ts:285-293`) then falls back to its own per-tx parse, which is the
    behaviour the comment now describes. (2) *Copy Raw Data* copying the full JSON array —
    `navigator.clipboard.readText` is denied in this headless context;
    `transactionFormatted` (`TransactionAdvancedView.tsx:202-210`) stringifies the array for N>1 and
    keeps the bare string for N=1. (3) A chain with `supports_simulations: false` rendering
    immediately — trivially true now that nothing gates the render. (4) Testnet toggle refetching
    the Solana interpretation — `source` is in the query key (`:1211`). (5) The
    `TransactionSimulation.tsx` call-site wrap, which feeds the old Send/Swap/Bridge views whose
    only affordance past Details is a real Confirm.
- **Console/network**: clean apart from the 500s I injected (4 per failing navigation — the
  react-query retries) and 404s for the fake `jup.ag`/`app.uniswap.org` favicons. Zero request
  failures against `zpi.zerion.io`, no React errors, no error boundaries, no page crashes in any
  run.
- **Observation, not a finding**: for my synthetic batch of three transfers to three different
  recipients, the backend's single aggregated action carries `Recipient 1111111111118Gy…` (the
  all-zeros Solana address) above the three rows. That is the interpretation endpoint's own output
  for this payload, and displaying the aggregated action verbatim is exactly what the PRD asks
  for; worth knowing that a batch with mixed recipients can produce an odd recipient line.
- **Animation**: no animation code in diff.

PR #965, opened 2026-08-22
WLT-1767: enable zero-value sends, promote custom data, add recipient chevron

zerts:
## Summary

Three changes to the extension Send form (`SendForm2`, route `/send-form`).

**1. Zero-value sends are possible again.** Both the button gate and the prepare-send gate collapsed *"no amount entered"* and *"amount is zero"* into the same falsy/`Number(x) === 0` check, so typing `0` left the button reading "Enter an Amount" and no transaction was ever prepared. Zero value is needed for non-payable calls — calldata-based claims, approvals, transfers.

The two states are now distinguished by numeric parse instead of truthiness (`shared/amount.ts`). Form state is URL-backed and only persists truthy strings, so an absent `inputAmount` already means the placeholder is showing; on top of that, anything that parses to a number counts as entered (`0`, `0.`, `0.000`), while partial input that isn't a number yet (a lone `.`) does not. NFT mode still requires at least one unit.

Also fixed as a direct consequence: **a typed zero is never a max-send.** `isMax` compares the amount to the position balance, so on the synthetic zero-balance position built for a token the wallet doesn't hold — the exact calldata-claim case — `0` equalled the balance and would have flipped the request to `max=true`, which the backend answers with an insufficient-balance error rather than a valid zero-value transaction.

**2. Custom data is a standing field.** It moves out of the collapsed Details card (a `Data` row behind a dialog) into its own floating frame below the asset fieldset, with an inline debounced textarea, shown whenever the Developer Tools "Custom Data" preference is on. Helper copy states the semantics Seb asked to clarify on the ticket — the value is appended to the end of the transaction data. Token mode only: `data` is cleared when an NFT is selected and the send hook ignores it in NFT mode, so a hidden field can never attach data invisibly (including via a hand-edited URL). The old row, its dialog, and `SendDataLine` are removed, leaving exactly one place to edit the value.

**3. Recipient trigger chevron.** A chevron sits right after the name in the `To:` row, using the same 20px `chevron-down` / `--neutral-500` / hover-darkening treatment as `ETH ⌄` in the asset selector.

Decisions and the alternatives that were rejected are captured as a PRD comment on the Linear issue.

Closes WLT-1767

## Test plan

Backend behaviour was verified directly against `transaction/get-send/v1` before implementing: `amount=0` (and `0.`, `0.000`, `00`) all return a valid transaction with `value: "0x0"` and no `error`, and a `data` param is appended to the end of the built calldata for both native and ERC-20 sends.

In the extension:

- [ ] Send form, token mode: type `0` → button becomes **Send Now** (not "Enter an Amount"), Details renders with a network fee, and the send signs and broadcasts.
- [ ] Clear the amount → button returns to "Enter an Amount" and the `0` placeholder shows.
- [ ] Type `0.` / `0.000` → enabled; type a lone `.` → still disabled.
- [ ] Pick a token the wallet holds none of, amount `0` → prepares a real zero-value tx instead of showing "Insufficient balance".
- [ ] Type the full balance → still sends as a max-send (unchanged).
- [ ] NFT mode: amount `0` stays disabled.
- [ ] Enable Developer Tools → Custom Data: the Data frame appears under the asset fieldset, typing re-prepares the send, **Clear** empties it (including mid-debounce), and the `Data` row is gone from Details.
- [ ] Select an NFT with data present → the frame disappears and the value is dropped from the URL.
- [ ] Recipient row shows the chevron next to the name, and next to the skeleton before a recipient is picked; tapping the row still opens the recipient dialog and the add-to-address-book button still works.

Checks: `npm install`, `type-check`, `lint` (0 errors; 2 pre-existing warnings in untouched files), `prettier --write` all pass. `test:unit` is at the known baseline — 36/37 suites, 363/364 tests, the one failure being the pre-existing drifted optimism fee assertion in `estimateNetworkFee.test.ts`.


zerts:

**Result:** `pass`

### Screenshots

**Popup unlocked, portfolio loaded (persona "default")**

Popup unlocked, portfolio loaded (persona "default")

**"inputAmount=0 — Send Now enabled with a live network fee (the fix)"**

"inputAmount=0 — Send Now enabled with a live network fee (the fix)"

**No amount entered — grey placeholder, no fee card, Enter an Amount**

No amount entered — grey placeholder, no fee card, Enter an Amount

**Custom Data frame below the asset fieldset, 0xdeadbeef typed**

Custom Data frame below the asset fieldset, 0xdeadbeef typed

**Details expanded — Data row gone, only Network Fee / Nonce / Network**

Details expanded — Data row gone, only Network Fee / Nonce / Network

**Clear empties the field and drops data from the URL**

Clear empties the field and drops data from the URL

**NFT mode — data frame correctly hidden**

NFT mode — data frame correctly hidden

## Summary

Persona "default" (`0x52Fb…78e59`, unlocked with the fixture password) at head `7553158f`,
driven in the popup at 440×900. The extension built, the popup unlocked, and the Send form
rendered real data on every one of the six form states exercised — no error boundary, no
blank screen, no stuck skeleton, and a clean console/network (0 non-noise console errors,
0 request failures against `zpi.zerion.io`).

The headline behaviour works: `#/send-form/?inputChain=ethereum&inputFungibleId=eth&inputAmount=0&to=0x975dd…8e37`
now renders an **enabled "Send Now"** with a live `$0.05` network fee, where the ticket's
own screenshot showed a disabled "Enter an Amount". The no-value state stays distinct —
with `inputAmount` absent the `0` renders as a grey placeholder, no fee card appears, and
the button reads "Enter an Amount". Both the promoted Custom Data frame and the recipient
chevron render as specified.

## Verification log

- **Profile:** zerion-wallet-extension — persona "default", seeded into a persistent
  user-data-dir across two launches; unlocked via the robust recipe (settle 2.5s, assert
  `inputValue`, wait to leave `#/login`).
- **Built:** `npm run build:chrome` → `dist/manifest.json` present (node_modules already
  warm). No dev server — extension loaded unpacked with `--headless=new`.
- **Gate flipped:** Developer Tools → **Custom Data** toggled ON in the UI at
  `#/settings/developer-tools` (persona prefs live in the encrypted wallet record, so they
  can't be seeded). Verified by the downstream effect, not by reading `cb.checked`.

**Amount gate — all six states behaved as designed:**

| Scenario | `inputAmount` | Button | Disabled |
|---|---|---|---|
| A typed zero | `0` | Send Now | **no** |
| B no amount | *(absent)* | Enter an Amount | yes |
| C zero with decimals | `0.000` | Send Now | **no** |
| D lone dot | `.` | Enter an Amount | yes |
| E normal amount (regression) | `0.0001` | Send Now | **no** |
| F zero + custom data | `0` | Send Now | **no** |

**Custom Data frame:**
- Renders as its own floating frame between the asset fieldset and the fee card, with the
  label, Clear action, inline textarea and the "Appended to the end of the transaction
  data" helper copy.
- Typing `0xdeadbeef` pushed `data=0xdeadbeef` into the URL and **re-prepared the send** —
  the network fee moved `$0.05 → $0.06`, i.e. the extra calldata reached the backend rather
  than being cosmetic.
- Details expanded shows only Network Fee / Nonce / Network — the old `Data` row is gone,
  so there's exactly one editor and no duplicate surface.
- **Clear** emptied the textarea (placeholder `0x...` returns) and removed `data` from the
  URL; the mid-debounce remount held, no stale value came back.
- NFT mode (`nftId` in the URL): `textarea[name="data"]` absent — frame correctly hidden.

**Recipient chevron (numeric assertion via `getComputedStyle` / `getBoundingClientRect`):**
`nameGroup` is `display: flex`; the chevron sits **4px** to the right of the name text
("second wallet" right edge 199 → chevron left 203), is **20×20**, and is coloured
`rgb(156, 159, 168)` = `--neutral-500` — matching the `ETH ⌄` treatment in the asset
selector, as specified.

**Console/network:** clean — `errors: []`, `failures: []`, `lifecycle: []`.

**One driver artifact, not a regression:** the NFT-mode screenshot shows an "Insufficient
USDC balance" banner. That's because I drove NFT mode with a fabricated `nftId`
(`0x0000…0000:1`); `inputFungibleId` then defaults to the persona's largest position
(USDC), which is the symbol the backend error is rendered with. The prepare gate for NFT
mode is byte-for-byte equivalent before and after this diff, so this state is unchanged by
the PR — it is an artifact of the fake id, not something this change introduced.

zerts:

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | code | isAmountEntered accepts negatives, so a URL-supplied `-1` passes the gate and mislabels the CTA (`src/ui/pages/SendForm2/shared/amount.ts:15`) |
| minor | tests | No unit test for shared/amount.ts, the helper the whole fix pivots on (`src/ui/pages/SendForm2/shared/amount.ts:13`) |
| minor | visual | Custom-data "Clear" keys off the debounced value, so it is absent for the first ~300ms of typing (`src/ui/pages/SendForm2/CustomDataField/CustomDataField.tsx:41`) |

## Summary

Reviewed head `7553158f` (PR #965) against WLT-1767 and Seb's clarifying comment on
the ticket. All three requested changes are implemented and verified live in the built
extension: a typed `0` now prepares and enables a real zero-value send while an untouched
input stays a placeholder; custom data is a standing floating frame below the asset
fieldset with the "appended to the end" wording Seb asked for; and the recipient trigger
carries a chevron that is pixel-for-pixel the `ETH ⌄` treatment (20×20, `--neutral-500`,
4px after the name, same hover darkening).

The two non-obvious consequences the PR claims to have handled both check out under
measurement, not just by reading. On the synthetic zero-balance position — the actual
calldata-claim case — `amount=0` goes out as `amount=0` with a live $0.08 fee and no
"Insufficient balance", where the old `isMax` comparison would have flipped it to
`max=true`; and typing the full ETH balance still produces `max=true` with no `amount`
param, so the max-send path is intact. Custom data is genuinely single-surfaced: the old
`Data` row is gone from Details, the NFT-mode request carries no `data` param even when
one is hand-planted in the URL, and picking a real NFT through the selector drops `data`
from the URL and hides the frame. The mid-debounce `Clear` remount works — the pending
keystroke never lands.

Type-check and lint are clean; unit tests are at the known baseline (the one failure is
the pre-existing drifted optimism fee assertion in a file this diff does not touch).
Nothing here warrants a fix iteration: the three findings are a hardening nit on
URL-supplied negative amounts, a missing unit test for the new helper, and a small
affordance-timing nit on the Clear button.

## Findings

### [minor/code] isAmountEntered accepts negatives, so a URL-supplied `-1` passes the gate
`isAmountEntered` (`shared/amount.ts:15`) treats anything BigNumber can parse as
"entered", which includes `-1`, `1e-9`, `0x10`, `Infinity` and `" 0"`. The old check was
`Number(x) > 0`, so it rejected all of those.

Repro: load `#/send-form/?inputChain=ethereum&inputFungibleId=eth&inputAmount=-1&to=0x975d…8e37`.
Expected: `Enter an Amount`, no backend traffic (the pre-diff behavior). Actual: the CTA
reads **"Send Now"** on a permanently-disabled button, and four `transaction/get-send/v1`
calls go out with `amount=-1` — three 500s plus a Chromium throttle (screenshot
`shots/07-negative.png`; console errors captured in `results.json`).

Impact is bounded and the state is not reachable from the keyboard — the amount input's
`onChange` filters through `/^\d*\.?\d*$/`, so only a hand-edited or shared URL gets you
there, and the button stays disabled because no transaction is ever prepared. A
`.gte(0)` alongside the `isNaN` check would restore the old guard without touching the
zero case.

### [minor/tests] No unit test for shared/amount.ts
`isAmountEntered` / `isZeroAmount` are the whole fix, and their semantics are exactly the
kind of thing that drifts: `0`, `0.`, `.0`, `0.000`, `00` are "entered"; `.` and `''` are
not. The sibling `TransactionWarning/resolveSendTransactionWarning.test.ts` establishes
the precedent for unit-testing SendForm2 helpers, and this one is cheaper to test than
that one. I verified the table by hand against BigNumber and live in the form, but nothing
in the suite pins it.

### [minor/visual] Custom-data "Clear" keys off the debounced value
`CustomDataField.tsx:41` gates the Clear button on `value` — the committed form state —
not on the textarea draft. So for the first ~300ms after you start typing into an empty
field there is no Clear affordance, and it then pops into the header row. Measured live:
immediately after filling the textarea, `Clear` was not visible (`Q_midDebounceClear.extra
.clearVisible: false`); it appeared once the debounce committed. Cosmetic only — clearing
still works the moment the button exists, including mid-debounce on a subsequent edit.

## Verification log

- **Tests:** `npm install` (clean), `npx tsc --noEmit` → 0 errors, `npx eslint
  src/ui/pages/SendForm2` → 0 problems, `npm run test:unit` → 36/37 suites, 363/363 tests
  passing. The one failing suite is `estimateNetworkFee.test.ts`; run in isolation with a
  `.env` present it fails on hardcoded optimism fee numbers that have drifted
  (`33077529225575` vs `33076771227277`) — a pre-existing baseline failure in a file
  outside the diff. No Playwright e2e run: `e2e/tests` is empty in this tree, so
  `test:e2e` has nothing to execute.
- **Live:** No dev server — this is an MV3 extension. `npm run build:chrome` → `dist/`
  (`manifest.json` present), loaded unpacked into `chromium --headless=new` via
  `launchPersistentContext`, persona "default" (`0x52Fb…78e59`) seeded into
  `chrome.storage.local` across two launches and unlocked with the fixture password.
  Viewports 440×900 (popup) and 1440×900 (tab). Developer Tools → **Custom Data** flipped
  in the UI. 5 driver passes, 30 screenshots in `/tmp/loop-review-FOYU5l/shots`,
  results in `results.json` … `results5.json`.
  - Amount gate, asserted on `input[name="inputAmount"].value` + CTA text/disabled:
    `0` → Send Now enabled + $0.05 fee; absent → grey placeholder, no fee card, "Enter an
    Amount"; `0.` and `0.000` → enabled; lone `.` → disabled; `0.0001` → enabled
    (`02-zero.png`, `03-no-amount.png`, `04-zero-dot.png`, `05-lone-dot.png`).
  - Zero-balance token (LINK, persona holds none, anywhere): `amount=0` → request
    `…&assetId=0x5149…&amount=0`, $0.08 fee, Send Now enabled, **no** insufficient banner
    (`10-zero-balance-token.png`). Same token at `amount=1` still shows "Insufficient LINK
    balance" and stays disabled, so the real insufficient path is untouched.
  - Max-send regression: full ETH balance → `…&assetId=eth&max=true` with no `amount`
    param (`12-max-send.png`).
  - Custom data: frame renders for native **ETH** (Seb's ask) between the asset fieldset
    and the fee card (fieldset bottom 276 → textarea top 336 → fee 494), with the
    "Appended to the end of the transaction data" copy; typing `0xdeadbeef` pushes
    `data=0xdeadbeef` into both the URL and the request; Details expanded shows only
    Network Fee / Nonce / Network (`16-details-expanded.png`); Clear empties the field and
    drops the param; typing then clearing inside the 300ms window commits nothing
    (`15-…`, `17-…`, `18-…`). Slow per-character typing fires one request per keystroke
    boundary including odd-length hex (`0xd`, `0xdea`) — the backend tolerates them, no
    transient error banner, fee and CTA stayed live throughout (`40-typed-slowly.png`).
    Non-hex text is UTF-8-hexlified (`data=0x68656c6c6f20776f726c64`), matching the
    documented convention.
  - NFT mode: frame hidden; a `data=` hand-planted in an NFT-mode URL is absent from the
    outgoing request; picking a real NFT ("Uniswap - 0.3% - USDC/ETH") through the
    selector drops `data` from the URL, and switching back to a token brings the frame
    back empty (`20-…`, `30-…`, `31-…`, `32-…`). `nftAmount=0` on a single-edition NFT is
    normalized to `1` by pre-existing `InputNftPosition` logic, so the NFT zero case is
    only reachable for ERC-1155 — and that gate is byte-identical to pre-diff.
  - Recipient chevron: `nameGroup` is flex; chevron 20×20, `rgb(156,159,168)` =
    `--neutral-500`, left edge 203 against a name right edge of 199 → 4px, identical to
    `AssetSelectorButton`'s treatment. Present in the no-recipient skeleton state too
    (`09-skeleton-chevron.png`). `aria-label="Select recipient"` overlay still opens the
    dialog and the add-to-address-book button is still hit-testable (`pointer-events:
    auto` on the action, `none` on the content layer).
  - Wide viewport: body caps at 393px, so the 1440px tab view is layout-identical; the
    chevron hugs the name for both a short ("zerts") and a long ("second wallet") name
    (`50-wide-zerts.png`).
- **Animation:** the diff's only motion is two `transition: … 0.15s ease` declarations —
  `border-color` on the textarea focus ring (moved verbatim from the deleted
  `.dataInput`) and `color` on the new chevron, the latter an exact copy of
  `AssetSelectorButton.module.css`'s existing rule, hover-gated behind
  `@media (hover: hover)`. No entrance/exit animation, no springs, no transforms, no
  `AnimatePresence`. The custom-data frame mounts and unmounts hard, which is the
  established treatment for banners in this vertical stack. No animation findings.
- **Console/network:** clean on every scenario except the deliberate `inputAmount=-1`
  probe — `errors: []`, `failures: []`, `lifecycle: []` across stages B–E. The negative
  probe produced three `500`s and one Chromium throttle on `transaction/get-send/v1`,
  which is finding 1.
- **PR threads:** GraphQL query returned `reviewThreads: []` — no human review comments,
  resolved or otherwise. The only PR conversation entries are the Linear linkback, the
  CI build artifact, and an `ai-runner` QA comment (skipped per instructions). Seb's two
  ticket comments — clarify that data is appended at the end, and it should appear when
  ETH is selected — are both satisfied.

src/ui/pages/SendForm2/CustomDataField/CustomDataField.tsx
```diff
@@ -0,0 +1,81 @@
...
+              onChange={(event) => handleChange(event.currentTarget.value)}
+              className={styles.input}
+              rows={2}
+              placeholder="0x..."
+              spellCheck={false}
+              autoComplete="off"
+            />
+          )}
+        />
+        <UIText kind="caption/regular" color="var(--neutral-600)">
```
zerts: let's remove this warning

zerts: Removed — the caption is gone entirely, so the field is now just the `Data` label, the `Clear` action, and the textarea.

src/ui/pages/SendForm2/useSendTransaction.ts
```diff
@@ -157,12 +171,18 @@ export function useSendTransaction({
     ? formState.nftAmount ?? '1'
     : resolvedInputAmount;
 
+  // Token mode prepares a send for any entered amount, including `0`; NFT mode
+  // still requires at least one unit.
+  const amountGatePasses = isNftMode
```
zerts, line 176: **Suggestion (duplication):** the NFT-vs-token amount gate now exists in three shapes — `resolveLabel` and `isDisabled` in `SendButton.tsx` each repeat `!formState.nftAmount || Number(formState.nftAmount) === 0` followed by `!isAmountEntered(...)`, and this ternary re-encodes the same branch a third way. `shared/amount.ts` was created for exactly this concern but only the token half got extracted; the NFT half stays `Number(...)`-based in all three places. A single mode-aware helper (e.g. `isSendAmountEntered(formState)` living next to `isAmountEntered`) that all three sites call would keep the gates from drifting apart. One wrinkle to preserve: this site additionally accepts `isMax`, so the helper would take that as input or the `isMax ||` stays local.

src/ui/pages/SendForm2/SendForm2.tsx
```diff
@@ -186,6 +188,16 @@ function SendFormComponent({
   const isNftMode = Boolean(formState.nftId);
   const defaultMode = isNftMode ? 'nfts' : 'tokens';
 
+  // Custom data lives in a standing frame under the asset fieldset whenever the
+  // Developer Tools "Custom Data" preference is on. Token mode only: appending
+  // calldata to an NFT transfer has no use case, and `selectNft` clears the
+  // value so nothing can be attached while the field is hidden.
+  const showCustomData = Boolean(
```
zerts, line 195: **Suggestion (duplication / drift risk):** this eligibility predicate — `configurableTransactionData && isEthereumAddress(address) && !isNftMode` — is written a second time in the `customData` memo in `useSendTransaction.ts` (plus `Boolean(formState.data)`). The hook-side re-check is good defense-in-depth (a leftover URL value must never attach invisibly), but a gate change now requires two synchronized edits. A shared `isCustomDataEligible({ preferences, address, isNftMode })` in `shared/` would keep the defense while removing the drift risk — the hook keeps calling it independently rather than trusting a prop.

src/ui/pages/SendForm2/CustomDataField/CustomDataField.module.css
```diff
@@ -0,0 +1,35 @@
+.frame {
+  background: var(--z-index-1);
+  border: none;
+  border-radius: 24px;
+  box-shadow: 0 4px 24px 0 rgba(0, 0, 0, 0.08);
```
zerts, line 5: **Suggestion (duplication):** `.frame` clones `FormFieldset`'s frame styles verbatim (`var(--z-index-1)`, radius 24, this box-shadow, padding 16). A frame restyle now touches two files in two idioms (inline styles there, CSS module here). Worth extracting a shared frame class or wrapper both use — and if one idiom wins, the CSS-module form here is arguably the better of the two.

src/ui/pages/SendForm2/shared/amount.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Zero-value sends are legitimate (calldata claims, approvals, non-payable
+ * calls), so `0` must not be conflated with an untouched input. Form state is
+ * URL-backed and only persists truthy strings, so an absent value already means
+ * the placeholder is showing; on top of that, anything that parses to a number
+ * counts as entered (`0`, `0.`, `0.000`), while partial input that isn't a
+ * number yet (a lone `.`) does not.
+ */
+export function isAmountEntered(value: string | null | undefined): boolean {
+  if (!value) return false;
+  return !new BigNumber(value).isNaN();
```
zerts: **Two nits, neither blocking:**

1. This largely re-derives `src/shared/isNumeric`, which `SendForm2.tsx` already imports for nonce validation. If BigNumber parsing is deliberate (consistency with the `eq`/`isZero` checks), a one-line comment saying so would stop the next reader from unifying them; otherwise delegating would settle it.

2. Parse looseness: BigNumber accepts `-1`, `0x0`, and `1e3`, so a hand-edited URL param passes this gate and `0.` is forwarded verbatim as the backend `amount`. Typed input is constrained by `FLOAT_INPUT_PATTERN` and main's `Number()` gate had the same slack, so this is inherited rather than a regression — but if you ever tighten it, this is the choke point.

(Also very minor: `isZeroAmount` below is exported with exactly one caller — fine as the pair-complement of `isAmountEntered`, just noting it.)

zerts (commented):
Two-axis review (standards vs spec), summarized:

**Spec** — all three WLT-1767 asks verified at this revision: the entered-vs-zero distinction matches the PRD cases (`0`, `0.`, `0.000`, `00` entered; lone `.` and empty are placeholder), the custom-data frame matches the asset fieldset treatment with the old Data row/dialog fully removed, and the recipient chevron matches the asset-selector treatment exactly. The PRD side-conditions hold too: typed zero never flips `max=true`, `selectNft` clears `data` and the hook ignores it in NFT mode, gating stays behind the Developer Tools preference + EVM senders. No missing or wrong requirements.

**Standards** — no violations of documented conventions (CONTEXT.md, ADR-0001 untouched). The inline comments below are all non-blocking judgement calls, mostly about duplication that could drift: the NFT/token amount gate written in three places, the custom-data eligibility predicate written twice, and the frame CSS cloned from `FormFieldset`.

Nothing blocking — these are pre-merge cleanup candidates.

zerts:

**Result:** `pass`

### Screenshots

**Popup unlocked and portfolio rendered (persona "default")**

Popup unlocked and portfolio rendered (persona "default")

**No value entered — grey placeholder, no fee, "Enter an Amount" disabled**

No value entered — grey placeholder, no fee, "Enter an Amount" disabled

**Typed zero — black 0, live $0.07 fee, "Send Now" enabled (headline fix)**

Typed zero — black 0, live $0.07 fee, "Send Now" enabled (headline fix)

**Partial non-numeric input "." still blocked**

Partial non-numeric input "." still blocked

**Custom Data preference off — no Data frame**

Custom Data preference off — no Data frame

**Custom Data frame standing below the asset fieldset, empty**

Custom Data frame standing below the asset fieldset, empty

**Calldata typed — Clear appears, send re-prepares with data**

Calldata typed — Clear appears, send re-prepares with data

**Clear discards the value and drops it from the URL**

Clear discards the value and drops it from the URL

**Details expanded — Network Fee / Nonce / Network, no Data row**

Details expanded — Network Fee / Nonce / Network, no Data row

**Recipient chevron after the name in the selector trigger**

Recipient chevron after the name in the selector trigger

**NFT mode — Data frame correctly hidden**

NFT mode — Data frame correctly hidden

## Summary

Drove the Zerion extension popup at PR head `b3c7256f` (PR #965, branch
`extension-enable-zero-value-transactions-in-extension-wlt-1767`) with QA persona
"default", active wallet `0x52Fb…78e59`, recipient `0x975dd…8e37` ("second wallet"
from the persona's address book). Built `dist/` from this HEAD, seeded
`chrome.storage.local`, unlocked, and exercised SendForm2 across 15 states. The
extension builds, the popup unlocks, and every affected state renders real backend
data — live network fees, no error boundaries, no blank screens, no stuck skeletons.
Console and network were completely clean on the measured pass (`errors: []`,
`failures: []`). All three ticket requirements are visibly satisfied and the
review-feedback commit (caption removal) did not disturb the field.

## Verification log

- Profile: zerion-wallet-extension — persona "default", password from fixture.
  No feature flag; the custom-data field is gated on the Developer Tools
  **Custom Data** preference, flipped ON in-UI at `#/settings/developer-tools`
  (screenshot 03). Screenshot 02 is the deliberate pref-OFF control.
- Served: N/A — MV3 extension, no dev server. `npm run build:chrome` → `dist/`
  (exit 0, `dist/manifest.json` present), loaded unpacked via
  `launchPersistentContext` + `--headless=new`, viewport 440×900.
- Flows: `#/overview` → `#/settings/developer-tools` (flip pref) →
  `#/send-form` across six amount states → custom-data type/clear →
  Details expand → NFT mode.

### Requirement 1 — no-value vs. typed zero

| `inputAmount` | input `.value` | Network fee | CTA | verdict |
|---|---|---|---|---|
| absent | `""` | none | "Enter an Amount" (disabled) | blocked, as designed |
| `0` | `"0"` | $0.07 | "Send Now" (enabled) | **sendable — the fix** |
| `0.` | `"0."` | $0.07 | "Send Now" (enabled) | sendable |
| `0.000` | `"0.000"` | $0.07 | "Send Now" (enabled) | sendable |
| `.` | `"."` | none | "Enter an Amount" (disabled) | blocked (not numeric yet) |
| `0.0001` | `"0.0001"` | $0.07 | "Send Now" (enabled) | control |

Asserted via `input[name="inputAmount"].value` because the absent and typed-zero
states both render the glyph `0` — the difference is placeholder grey vs. black,
visible in screenshots 04 vs. 05.

Backend evidence from captured `transaction/get-send/v1` calls (9 total, all
`200`, no `errors`):
- `amount=0` was sent with **`max` absent**, confirming the `isMax` guard: on a
  position where the typed amount equals the balance, a zero must not flip to
  `max=true` (the backend answers that with insufficient-balance). `max` and
  `amount` are mutually exclusive in the request builder, so an absent `max` is
  positive proof the guard held.
- `amount=0&data=0xdeadbeef` — custom data reaches the backend on a zero-value
  send.
- **No `get-send` call at all** for the absent and lone-dot states — the amount
  gate blocks preparation rather than sending a bad request.

### Requirement 2 — custom data as a standing frame

- Preference OFF ⇒ no Data frame anywhere (screenshot 02).
- Preference ON ⇒ frame present in every token-mode state, empty and populated
  alike, in its own floating card below the asset fieldset (screenshots 10, 05,
  09 — the field shows regardless of amount, which is the "all the time" ask).
- Typing `0xdeadbeef` reveals `Clear` and re-prepares the send with the data
  param (screenshot 11). `Clear` restores the placeholder, hides itself, and
  drops `data` from the hash (screenshot 12) — the debounce-cancel remount works.
- The old Details row is gone: expanded Details reads exactly
  `["Network Fee", "$0.07", "Nonce", "247", "Network", "Ethereum"]`
  (screenshot 13). Scoped to the Details card, not the page body — the promoted
  field has its own `Data` label that false-positives a body-wide check.
- NFT mode hides the frame (screenshot 15), matching the token-mode-only decision.
- The review-feedback commit removed the helper caption; `grep` confirms neither
  "use with caution" nor "Appended to the end" survives anywhere in `dist/`, and
  the frame still reads cleanly as label + textarea.

### Requirement 3 — recipient chevron

Measured on the filled trigger: `20×20`, `color: rgb(156, 159, 168)`
(`--neutral-500`), `display: block`, sitting `4px` after the name text
("second wallet"), `chevronRightOfName: true`, vertically centred with the name
(`sameRow: true`). Matches the token-name chevron treatment the ticket asked for.

### Known artifact — not a finding

Screenshot 15 carries an "Insufficient USDC balance" banner. That is the
documented consequence of driving NFT mode with a fabricated `nftId`: with no
`inputFungibleId` in the query, `formState` falls back to the persona's largest
position (USDC), so the backend's error for the bogus NFT renders with that
symbol. The NFT amount gate is behaviourally unchanged by this diff. The point of
the capture — that the Data frame is absent in NFT mode — is unaffected.

### Console / network

Clean. `errors: []`, `failures: []`, `lifecycle: []` across all 15 stages, with
only `ERR_ABORTED` (superseded quote streams) and the standard analytics hosts
filtered. `zpi.zerion.io` / `api.zerion.io` were never filtered and never failed.

zerts:

**Verdict:** `approved`

| Severity | Area | Finding |
| --- | --- | --- |
| minor | code | Append-at-the-end semantics now stated nowhere in the UI, and the PR body still claims they are (`src/ui/pages/SendForm2/CustomDataField/CustomDataField.tsx:73`) |
| minor | code | Open thread unanswered — NFT-vs-token amount gate still duplicated in three shapes (`src/ui/pages/SendForm2/useSendTransaction.ts:176`) |
| minor | code | Open thread unanswered — custom-data eligibility predicate written twice (`src/ui/pages/SendForm2/SendForm2.tsx:195`) |
| minor | code | Open thread unanswered — .frame clones FormFieldset's inline frame styles (`src/ui/pages/SendForm2/CustomDataField/CustomDataField.module.css:5`) |
| minor | code | Open thread unanswered — no note on the deliberate BigNumber-over-isNumeric choice (`src/ui/pages/SendForm2/shared/amount.ts:15`) |
| minor | tests | Still no unit test for shared/amount.ts, the helper the whole fix pivots on (`src/ui/pages/SendForm2/shared/amount.ts:13`) |

## Summary

Reviewed head `b3c7256f`, which adds exactly one commit on top of the previously
reviewed `7553158f`: the four-line removal of the custom-data helper caption. Re-verified
the whole feature from scratch rather than trusting the earlier pass.

**The one review thread that existed when the implementation turn ran was genuinely
addressed.** zerts asked "let's remove this warning" on `CustomDataField.tsx:74`; the
caption is gone from the source, the reply on the thread says so accurately, and I
confirmed live that no "Appended to the end…/use with caution" text renders in any of the
fifteen form states I drove (`captionPresent: false` in every one). The frame did not
degrade: it is now `Data` + `Clear` header row, textarea, 16px padding all round, with no
orphaned whitespace where the caption used to sit.

**Nothing regressed.** The full behaviour matrix still holds at this head: typed `0` →
enabled Send Now with a live fee, absent amount → grey placeholder and "Enter an Amount",
`0.` enabled / lone `.` disabled, the zero-balance-LINK claim case sends `amount=0` with
no insufficient banner while the same token at `amount=1` still blocks, full balance still
goes out as `max=true`, typed data reaches the backend as `&data=0xdeadbeef`, NFT mode
hides the frame and strips `data` from the request, Details shows only Network Fee / Nonce
/ Network, and the chevron is still 20×20 `--neutral-500` at 4px after the name.
Type-check and lint clean; unit tests at the known baseline. Console and network clean
across every scenario.

**Four review threads are open and unanswered** — but they were created at 13:59:46Z,
about ten minutes *after* the fix commit was authored (13:49:51Z), so the `/address-review`
pass never saw them. All four are self-labelled non-blocking: three are "Suggestion
(duplication)" refactors (extract a mode-aware amount helper, extract
`isCustomDataEligible`, extract a shared frame class) and one is "Two nits, neither
blocking". I confirmed none of them landed at this head and none has a reply. They are
listed below as `minor` so they travel back to the PR, but they are structural
preferences, not defects, so they do not warrant burning an iteration — the call on
whether to refactor or just reply is the author's.

One carry-over correction: last round I reported the URL-supplied negative amount
(`inputAmount=-1`) as a finding. The reviewer has since logged that same observation on
`amount.ts:15` and accepted it as "inherited rather than a regression", so I am not
re-reporting it. They are right that the label was never gated on it — `main`'s
`Number(x) === 0` check also fell through to "Send Now" for `-1`; only the wasted prepare
request is new, and the button stays disabled either way.

## Findings

### [minor/code] Append-at-the-end semantics now stated nowhere in the UI, and the PR body still claims they are
Removing the caption was explicitly requested and I am not arguing it — the field sits
behind a Developer Tools toggle and the copy was not earning its space. Two loose ends
follow from it, though.

First, Seb's ticket comment ("custom data: clarify it gets appended at the end of tx
data") now has no expression anywhere in the product: the caption is gone, the old
dialog's "Advanced Feature: Use with Caution" line went with `SendDataLine`, the textarea
placeholder is just `0x...`, and `ToggleSettingLine text="Custom Data"` in
`Settings.tsx:458` carries no description. The reporter asked for that clarification and
is likely to notice it silently disappeared — worth a note back to him on WLT-1767 rather
than a silent drop.

Second, the PR body still asserts the opposite: *"Helper copy states the semantics Seb
asked to clarify on the ticket — the value is appended to the end of the transaction
data."* That line is now false and will mislead whoever reads the description at merge
time.

### [minor/code] Open thread unanswered — NFT-vs-token amount gate still duplicated in three shapes
Thread on `useSendTransaction.ts:176` asks for a mode-aware `isSendAmountEntered(formState)`
next to `isAmountEntered`, because the NFT half of the gate stayed `Number(...)`-based in
three places. Unchanged at this head: `SendButton.tsx:32` and `:63` each still carry
`!formState.nftAmount || Number(formState.nftAmount) === 0`, and
`useSendTransaction.ts:176` re-encodes the same branch as a ternary. No code change, no
reply. Behaviourally the three are equivalent today (I verified the NFT gate is
byte-identical to pre-diff), so this is drift risk rather than a live bug.

### [minor/code] Open thread unanswered — custom-data eligibility predicate written twice
Thread on `SendForm2.tsx:195` asks for a shared `isCustomDataEligible({ preferences,
address, isNftMode })` so the render gate and the hook-side re-check cannot drift.
Unchanged: the predicate is still spelled out at `SendForm2.tsx:195-199` and again in the
`customData` memo in `useSendTransaction.ts:133-140`. No code change, no reply. The
duplication is deliberate defense-in-depth and I verified it works — a `data=` planted in
an NFT-mode URL is absent from the outgoing request — so the concern is only that a future
gate change needs two synchronized edits.

### [minor/code] Open thread unanswered — .frame clones FormFieldset's inline frame styles
Thread on `CustomDataField.module.css:5` notes `.frame` reproduces `FormFieldset`'s frame
verbatim in a second idiom. Unchanged: `--z-index-1`, radius 24, `0 4px 24px rgba(0,0,0,.08)`
and padding 16 exist both as inline styles in `FormFieldset.tsx:26` and as a CSS module
here. No code change, no reply. Visually this is currently a feature — the two frames match
pixel-for-pixel in the screenshots — which is exactly why a restyle would need to touch both.

### [minor/code] Open thread unanswered — no note on the deliberate BigNumber-over-isNumeric choice
Point 1 of the thread on `amount.ts:15` asks for a one-line comment explaining why this
uses BigNumber parsing rather than delegating to `src/shared/isNumeric` (which
`SendForm2.tsx` already imports for nonce validation), so the next reader does not "helpfully"
unify them. The docblock at `amount.ts:3-12` still explains only the zero-vs-absent
semantics, not the parser choice. Cheapest item in this list to close.

### [minor/tests] Still no unit test for shared/amount.ts
Carried over from the previous round and still open. `isAmountEntered` / `isZeroAmount` are
the crux of the fix and their table is exactly the kind of thing that drifts: `0`, `0.`,
`.0`, `0.000`, `00` entered; `.` and `''` not. `TransactionWarning/resolveSendTransactionWarning.test.ts`
is the local precedent for unit-testing a SendForm2 helper, and this one is cheaper to
test than that one. I verified the table by hand against BigNumber and live in the form,
but nothing in the suite pins it.

## Verification log

- **Threads:** GraphQL `reviewThreads(first:100)` → 5 threads. 1 resolved+outdated
  (`CustomDataField.tsx`, "let's remove this warning") → **honored in code and in the
  running UI**. 4 unresolved, all created `2026-08-24T13:59:46Z`, i.e. after the fix
  commit's author time (`14:49:51 +0100` = `13:49:51Z`) → not addressed, no reply; each
  re-checked against the head individually (see findings). No `_Posted by ai-runner._`
  comments were treated as spec. Seb's two Linear comments re-read: "should appear if ETH
  is selected" is satisfied (verified live with `inputFungibleId=eth`); "clarify it gets
  appended at the end" was satisfied by the caption and is now unmet by explicit reviewer
  request — finding 1.
- **Tests:** `npx tsc --noEmit` → 0 errors. `npx eslint src/ui/pages/SendForm2` → 0
  problems. `npm run test:unit` → 36/37 suites, 363/364 tests; the single failure is
  `estimateNetworkFee.test.ts` asserting drifted optimism fee constants
  (`33077529225575` vs `33076771227277`) in a file outside the diff — the known baseline.
  No e2e run: `e2e/tests` is empty, so `test:e2e` has nothing to execute.
- **Live:** No dev server — MV3 extension. `npm run build:chrome` → `dist/manifest.json`
  present, loaded unpacked into `chromium --headless=new` via `launchPersistentContext`;
  persona "default" (`0x52Fb…78e59`) seeded into `chrome.storage.local` across two
  launches, unlocked with the fixture password; Developer Tools → Custom Data flipped in
  the UI. Viewport 440×900 (the popup's own size; the form's column caps at 393px, so the
  wide-tab view was confirmed layout-identical last round and is unaffected by a
  text-removal commit). 15 scenarios, 11 screenshots in
  `/tmp/loop-review-Kzr2W2/shots`, raw assertions in `/tmp/loop-review-Kzr2W2/results.json`.
  - Caption: `captionPresent` false in all 15 states. Frame geometry stable at 96px tall
    (label bottom 328 → textarea 336–404, 8px label gap, 16px frame padding), asset
    fieldset bottom 276 → frame top 308, fee card below — still "below the asset
    fieldset" per the ticket (`05-data-frame-empty.png`, `06-data-typed.png`).
  - Amount gate: `0` → Send Now enabled; absent → "Enter an Amount" disabled; `0.` →
    enabled; `.` → disabled; `0.0001` → enabled (`02-zero.png`, `03-no-amount.png`).
  - Claim case: LINK (held nowhere by this persona) at `amount=0` → request
    `…&assetId=0x5149…&amount=0`, live fee, enabled, no insufficient banner
    (`04-zero-balance-token.png`); the same token at `amount=1` → "Insufficient LINK
    balance", disabled.
  - Max-send: full ETH balance → `…&assetId=eth&max=true`, no `amount` param.
  - Custom data: typing → `&data=0xdeadbeef` on the wire; Details expanded reads exactly
    `Network Fee / Nonce / Network` (scoped label read, no Data row —
    `07-details-expanded.png`); Clear empties the field and drops the param; committing
    `0xaaaa` then retyping `0xbbbbbbbb` and clearing inside the 300ms window sent only
    `data=0xaaaa` and settled empty — the pending keystroke was cancelled, not replayed.
  - NFT mode: frame absent, and a `data=` planted in the URL is absent from the outgoing
    request (`10-nft-mode.png`).
  - Chevron: 20×20, `rgb(156,159,168)` = `--neutral-500`, left edge 203 vs name right
    edge 199 → 4px; `aria-label="Select recipient"` overlay and the address-book action
    both still present and hit-testable (`11-chevron.png`).
- **Animation:** the only commit in this round deletes a `UIText` node — no motion code
  touched. The diff's overall motion surface is unchanged from last round: two
  `transition: … 0.15s ease` declarations (textarea focus `border-color`, chevron `color`,
  the latter an exact copy of `AssetSelectorButton.module.css` and hover-gated behind
  `@media (hover: hover)`). No entrance/exit animation, no springs, no transforms. The
  frame mounts and unmounts hard, matching the established treatment for this vertical
  stack. No animation findings.
- **Console/network:** clean — `errors: []`, `failures: []`, `lifecycle: []` across all 15
  scenarios. No negative-amount probe this round, so nothing analogous to last round's
  deliberate 500s.

src/ui/pages/SendForm2/shared/amount.ts
zerts: Fixed the parse-looseness half in 73560395: `isAmountEntered` now matches the amount input's own `FLOAT_INPUT_PATTERN` (anchored) instead of a general BigNumber parse, so `-1`, `0x0`, and `1e3` no longer pass the gate. All PRD cases are unchanged (`0`, `0.`, `0.000`, `00` entered; a lone `.` and empty are the placeholder state) and locked in by a colocated unit test. This also settles nit 1 — the helper no longer re-derives a numeric parse, it delegates to the same pattern the input uses. `0.` is still forwarded verbatim (backend-verified). Leaving `isZeroAmount`'s single export as is.

